PR #125, opened 2013-09-19
Expands "\\n"->"\n" in messages, not just the format

LukeShu:
Dunst expands a "\n" to "\n" in messages. This is inconsistent with notification-daemon and xfce4-notifyd.

I believe that this is unintentional; it is meaning to only expand "\n" in the format string, but because it is done after the message has been inserted, it expands everything.

### DIFF
--- a/notification.c
+++ b/notification.c
@@ -281,7 +281,8 @@ int notification_init(notification * n, int id)
 
         n->urls = notification_extract_markup_urls(&(n->body));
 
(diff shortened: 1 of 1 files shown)


wavexx:
I don't think this PR is necessary anymore after PR #204 / #210.

LukeShu:
This was indeed fixed in #204, but it later regressed in commit 3ef15065.  I'll submit a new version of it.

PR #126, opened 2013-09-19
The INI parser is buggy

LukeShu:
The bug:

Quoted strings and comments don't work like they are clearly intended to. By that, I mean that that portion of the code is inactive. The parsed value is literally just the right side of the assignment passed through `g_strstrip()`. The issue is that a couple of pointers aren't being referenced before assigning to them in `option_parser.c`

The patch:

If you don't like this implementation, there are two others in my repo. This one is (IMO) the most readable, and the most modifiable. There is another one that is faster (premature optimization, yada yada), and one that minimizes the changes to the source, but still behaves in slightly odd (but not "wrong") ways.

The diff for this is on top of the diff for my previous pull request (about `"\\n"`).

I apologize for the size of the patch-set.  One of the commits included is just cleaning up a few string manipulation functions, since I was staring at string manipulation anyway.

### DIFF
--- a/dunstify.c
+++ b/dunstify.c
@@ -199,7 +199,7 @@ void closed(NotifyNotification *n, gpointer foo)
 void add_action(NotifyNotification *n, char *str)
 {
     char *action = str;
-    char *label = strstr(str, ",");
+    char *label = strchr(str, ',');
 
     if (!label || *(label+1) == '\0') {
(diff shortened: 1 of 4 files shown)


wavexx:
The first commit shouldn't be necessary anymore. The rest though might still be relevant.

LukeShu:
The same bug is still in it.  I'm opening a new PR with a rebased version of the last commit.

PR #127, opened 2013-09-19
Feature: Support backslash escapes in format to embed quotes (and tabs, etc)

LukeShu:
The feature:

I would like there to be backslash escapes for values in the INI file. It currently expands `"\n"` -> newline in the format string, but nothing else. There is no way to make a value include a double quote (`"`) character.

The patch:

This makes it recognize `\n`, `\t`, `\"` and `\\`. For anything else, it will print a warning, then include the character after the backslash verbatim (so `\f` would become `f`).

This patchset looks huge; it's not. Only look at the top commit; it is on top of the pull request to fix the parser.

### DIFF
--- a/dunstify.c
+++ b/dunstify.c
@@ -199,7 +199,7 @@ void closed(NotifyNotification *n, gpointer foo)
 void add_action(NotifyNotification *n, char *str)
 {
     char *action = str;
-    char *label = strstr(str, ",");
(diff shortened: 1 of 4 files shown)


wavexx:
Again, is this still necessary after PR #204 / #210 ?
You can now consider incoming text to be either literal text or pango markup.

Issue #152, opened 2014-03-10
Text not centered vertically if icon is too big


knopwob:
![2014-03-10-111830_300x52_scrot](https://f.cloud.github.com/assets/136023/2372618/997c03da-a83d-11e3-9be1-2918fc06ccb0.png)

PR #174, opened 2014-05-29
Vertical centering and padding

tatou-tatou:
I made the text center when the icon is bigger than the text:
centered text
This should solve issue #152.

I also edited the way the frames are drawn and the padding is made, with the current behavior icons are not padded correctly when they are bigger than the text (they appear too low and can even overflow out of the window if there is no padding).
With those edits, the icon will appear centered if it's bigger than the text, as shown in the picture above.

**Edit:**
Here is more explanation for why padding should be edited and how it is related to centering. With the current situation, the padding is very inconsistent when having multiple notifications, the separators are taken into account, but not the frame.

Making the padding internal to the frame (like with CSS borders), and not overlapping with it, would be more coherent in my opinion.

Here are some screenshots showing it with the padding set to zero. On the left is the current situation and on the right is the proposed solution.
padding icons
padding text

### DIFF
--- a/x.c
+++ b/x.c
@@ -457,6 +457,9 @@ static dimension_t x_render_layout(cairo_t *c, colored_layout *cl, dimension_t d
         if (cl->icon && settings.icon_position == icons_left) {
                 cairo_move_to(c, settings.frame_width + cairo_image_surface_get_width(cl->icon) + 2 * settings.h_padding, bg_y + settings.padding + h/2 - h_text/2);
 	}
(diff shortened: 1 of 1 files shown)


Donvittorio:
Would love to see this get merged, +1

gnaqvi:
Patiently waiting for the merge. *rubs hands*

gnaqvi:
Patiently waiting for the merge. @knopwob

PR #210, opened 2014-12-21
Fix/simplify notification_strip_markup

wavexx:
Some extra tweaks to the markup support, since I noticed that the old strip_markup was also broken:
- Strip markup before interpreting entities.
- Handle _any_ tag uniformly (`<span>`is also handled by pango)

### DIFF
--- a/notification.c
+++ b/notification.c
@@ -163,23 +163,12 @@ void notification_free(notification * n)
          */
 char *notification_strip_markup(char *str)
 {
-        char *replace_buf, *start, *end;
-
         if (str == NULL) {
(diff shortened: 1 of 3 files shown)


wavexx:
Closed as fixed?

tsipinakis:
Closed because it has been merged in 812b3f972f29eb3c8568b0325a05d55d738e9eaf

wavexx:
On Sat, Jan 14 2017, Nikos Tsipinakis wrote:
> Closed because it has been merged in 812b3f9

Thanks!

PR #214, opened 2015-01-18
Don't trim leading whitespace, fixes #208.

### LINKED ISSUE #208: Wrong indentation when lines begin with a space
To see how dunst changes the indentation, run these examples with:

```
notify-send "$(cat test.txt)"
```

Make sure not to lose the leading spaces when testing!

Input file 1:

```
 x
 x
 x
```

Notification:

```
x
 x
 x
```

Input file 2:

```
 x
x
 x
```

Notification:

```
x
x
 x
```

Input file 3:

```
 x
x
x
```

Notification:

```
x
x
x
```

imba3r:
Use g_strchomp (https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strchomp) instead of g_strstrip which both removes leading and trailing whitespace.

### DIFF
--- a/notification.c
+++ b/notification.c
@@ -356,7 +356,7 @@ int notification_init(notification * n, int id)
                 n->msg = string_replace_all("%p", "", n->msg);
         }
 
-        n->msg = g_strstrip(n->msg);
+        n->msg = g_strchomp(n->msg);
 
         if (id == 0) {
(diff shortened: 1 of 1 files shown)


hughdavenport:
I can confirm this fixes #208 

Cheers,

Hugh

PR #230, opened 2015-07-01
Add option to always run script (even for suppressed notifications)

springworm:
Adds option always_run_script.
If set to true, scripts defined by a rule will always be executed, even if the format is set to "" to suppress the notification.

Niche, but useful when a user wants an action triggered by an application's notification but doesn't want the actual notification (see issue 212).

Please excuse my being bad at git.

### DIFF
--- a/README.pod
+++ b/README.pod
@@ -115,6 +115,10 @@ print version information.
 
 use alternative config file.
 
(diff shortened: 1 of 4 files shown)


knopwob:
Looks good to me. Thank you!

PR #231, opened 2015-07-20
fix memleaks

kongo2002:
Hi,

I could not find any occurance of those notification fields being freed.

Cheers,
Gregor


dcousens:
@knopwob any change you could merge this?

PR #243, opened 2015-11-03
Fixed #241 and adds a command to replace the last notification

### LINKED ISSUE #241: Add DUNST_COMMAND for shortcut actions
In stead of defining keybindings in dunst, I would like to use my WM to send commands to dunst to `close`, `close_all` and show `history`.

These could be defined like the pause/resume functuionality by sending a notificatione with a summary of eg `DUNST_COMMAND_CLOSE`, `DUNST_COMMAND_CLOSEALL`, `DUNST_COMMAND_HISTORY`.

My C is non-existant but browsing the code here on github maybe at least the last 2 would be easily implemented?

tivervac:
This little piece of code allows to clear the message queue, close the last message, call the last message from history or replace the text of thelast notification by sending a message, the same way the program can be (un)paused.

This allows for these actions to be used in scripts.

### DIFF
--- a/README.pod
+++ b/README.pod
@@ -180,6 +180,11 @@ When paused dunst will not display any notifications but keep all notifications 
 This can for example be wrapped around a screen locker (i3lock, slock) to prevent flickering
 of notifications through the lock and to read all missed notifications after returning to the computer.
 
+Sending a notification with a summary of "DUNST_COMMAND_CLOSEALL" will remove all the notifications.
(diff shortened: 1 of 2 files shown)


surielsanctos:
At the moment DUNST_COMMAND_REPLACE doesnt work with the urgency level. If you send a normal message and a critical one to replace the first one, text is replaced but urgency stays at normal and vice versa.

tsipinakis:
Closing this as there is no way this is getting merged currently. We have plans to add a new dbus interface for commands as part of #284.

PR #260, opened 2016-04-13
Fix crash on OpenBSD 

technoprole:
Adding missing header files and defaulting pointers to NULL fixes crashes on OpenBSD enviroment


enodata:
Appears to be fixed in the current master (tested on OpenBSD 6.1 amd64).

tsipinakis:
Thanks.

PR #270, opened 2016-09-02
fix/read-category-from-rule

bernardn:
Category was missing from the rules parsing.
This made category-based rules applied anytime.

### DIFF
--- a/notification.c
+++ b/notification.c
@@ -335,6 +335,10 @@ int notification_init(notification * n, int id)
 
         n->format = settings.format;
 
+        if (n->category == NULL) {
+                n->category = "";
+        }
+        
         rule_apply_all(n);
 
         n->urls = notification_extract_markup_urls(&(n->body));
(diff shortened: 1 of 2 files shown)


notification.c
```diff
@@ -335,6 +335,10 @@ int notification_init(notification * n, int id)
 
         n->format = settings.format;
 
+        if (n->category == NULL) {
+                n->category = "";
+        }
+        
         rule_apply_all(n);
```
bernardn: If  rule_apply_all is called with NULL notification category AND non-NULL rule category, strlen SIGSEGV.

#0  0x00007f0aca0cf600 in strnlen () from /usr/lib/libc.so.6
#1  0x00007f0aca10fc11 in fnmatch@@GLIBC_2.2.5 () from /usr/lib/libc.so.6
#2  0x000000000040870e in rule_matches_notification (r=r@entry=0x69e7a0, n=n@entry=0x6c0b80) at rules.c:78
#3  0x000000000040874f in rule_apply_all (n=n@entry=0x6c0b80) at rules.c:41
#4  0x000000000040957e in notification_init (n=n@entry=0x6c0b80, id=id@entry=0) at notification.c:338
#5  0x000000000040403b in main (argc=3, argv=0x7ffcf0c57a18) at dunst.c:341

PR #274, opened 2016-11-29
Fix category rule matching

quite:
Bug: the existance of a rule section containing a category match
resulted in *all* notifications matching this particular rule.

Solution: Properly extract the category string from the rule section. If
the rule section did not have a category matcher, make sure we at least
have the empty string before doing rule_apply_all() (fnmatch() borks on
NULL!)


quite:
Oops, here be duplicates...

PR #282, opened 2017-01-20
Fix #281 bad free. Also fix a memory leak.

### LINKED ISSUE #281: Dunst crashes with new_icon and multiple notifications in a row
## Step to Reproduce

1. Use this config file

       [global]
           icon_position = left
           icon_folders = /can/be/any/path
       [thesummary]
           summary = "thesummary"
           new_icon = need_not_be_real_icon

2. Execute `notify-send thesummary thecontent` several (3 or 4) times *before* the notification times out.

## Observed Behavior

1. The icon disappears after the second notification was sent (if real icon is used).
2. Dunst crashes after more notifications were sent.

## Program Version

Dunst 1.1.0 on latest Arch Linux
**EDIT:** Also tested on Dunst v1.1.0-121-gec7905f. Still the same

## Other Details

1. Messages like this are observed on stderr before Dunst crashes:

       Could not load icon: 'Ðá~'
       Could not load icon: 'Àú+'

   Then Dunst crashes with either SIGSEGV or SIGABRT. Sometimes with an assertion error or

       *** Error in `dunst': double free or corruption (fasttop): <some-number> ***

2. Dunst works fine if the icon was specified with `notify-send -i <icon>`.

johnchen902:
With `n->icon = r->new_icon` and later `free(n->icon)`, `r->new_icon` is wrongly freed. Fix that with strdup.
Also fix an obvious memory leak by the way.

Intended to fix #281.

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -2,8 +2,6 @@
 
 #include <glib.h>
 #include <fnmatch.h>
-#include <stdlib.h>
-#include <string.h>
 
 #include "dunst.h"
 #include "rules.h"
@@ -23,8 +21,8 @@ void rule_apply(rule_t * r, notification * n)
                 n->plain_text = r->plain_text;
         if (r->new_icon) {
                 if(n->icon)
-                        free(n->icon);
-                n->icon = strdup(r->new_icon);
+                        g_free(n->icon);
+                n->icon = g_strdup(r->new_icon);
         }
         if (r->fg)
                 n->color_strings[ColFG] = r->fg;
--- a/src/rules.c
+++ b/src/rules.c
@@ -2,6 +2,8 @@
 
 #include <glib.h>
 #include <fnmatch.h>
+#include <stdlib.h>
+#include <string.h>
 
 #include "dunst.h"
 #include "rules.h"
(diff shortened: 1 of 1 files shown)


src/rules.c
```diff
@@ -19,8 +21,11 @@ void rule_apply(rule_t * r, notification * n)
                 n->allow_markup = r->allow_markup;
         if (r->plain_text != -1)
                 n->plain_text = r->plain_text;
-        if (r->new_icon)
-                n->icon = r->new_icon;
+        if (r->new_icon) {
+                if(n->icon)
+                        free(n->icon);
```
tsipinakis: Since we already include glib, let's avoid unecessary includes and use g_free here.

src/rules.c
```diff
@@ -19,8 +21,11 @@ void rule_apply(rule_t * r, notification * n)
...
+                n->icon = strdup(r->new_icon);
```
tsipinakis: Same here, this can be replaced with g_strdup and drop the string.h include.

johnchen902:
Made the requested changes.

BTW why `free` and `g_free` seems to be used interchangeably in this project? Should it be addressed? (not here though; there's too much of this)

tsipinakis:
Thanks!
I agree that using `free` and `g_free` interchangeably is a problem, probably the same goes for all glib functions and their glibc counterparts. It should probably be addressed before the next release.

PR #285, opened 2017-01-27
Fix a double free in on_get_capabilities.

### LINKED ISSUE #215: Assertion error for notifications with volume change in Ubuntu: g_variant_unref: assertion 'value->ref_count > 0' failed
I am seeing the following error when pressing the volume up/down keys:

```
(process:30160): GLib-CRITICAL **: g_variant_unref: assertion 'value->ref_count > 0' failed
```

`dunst -print` does not print anything additionally.

See #159 for what gets send in this case.

The code in question appears to be this: https://github.com/knopwob/dunst/blob/master/dbus.c#L118

Stebalien:
`value` was already being freed by `g_dbus_method_invocation_return_value`.

Fixes #215


tsipinakis:
Thanks!

PR #292, opened 2017-02-07
Fix `#include`s and use only glib for memory management

johnchen902:
1. Fix `#include`s.
2. Use only glib functions for memory management. Don't mix with stdlib
3. Fix a missing `*` in option_parser.c

There are lots of of changes and I wasn't sure if it would be a good pull request, but I was invited by a maintainer.

I know there are conflicts, but I'm busy this week and do not have time to resolve them.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -1,15 +1,16 @@
 /* copyright 2013 Sascha Kruse and contributors (see LICENSE for licensing information) */
-
-#include <stdlib.h>
-#include <stdio.h>
-#include <glib.h>
-#include <string.h>
-#include <gio/gio.h>
-#include "dunst.h"
 #include "dbus.h"
+
+#include <gio/gio.h>
+#include <glib.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "dunst.h"
 #include "notification.h"
-#include "utils.h"
 #include "settings.h"
+#include "utils.h"
 
 GDBusConnection *dbus_conn;
 
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -131,10 +131,7 @@ static void on_notify(GDBusConnection * connection,
         gchar *icon = NULL;
         gchar *summary = NULL;
         gchar *body = NULL;
-        Actions *actions = g_malloc(sizeof(Actions));
-        if(actions == NULL) {
-                die("Unable to allocate memory", EXIT_FAILURE);
-        }
+        Actions *actions = g_malloc0(sizeof(Actions));
         gint timeout = -1;
 
         /* hints */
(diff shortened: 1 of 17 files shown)


tsipinakis:
> There are lots of of changes and I wasn't sure if it would be a good pull request

These are all changes I was considering doing myself, so this saves me a big chunk of time. Thanks!

> but I was invited by a maintainer.

That would be @Eizen. We are planning to do a code style migration soon-ish which will probably cause every little thing to conflict so we are trying to merge what has already been worked on now rather than later.

> I know there are conflicts, but I'm busy this week and do not have time to resolve them.

That's fine, if you can work on this within a reasonable timeframe please do. If not, we can take over from here.

src/dbus.c
```diff
@@ -1,15 +1,14 @@
 /* copyright 2013 Sascha Kruse and contributors (see LICENSE for licensing information) */
-
-#include <stdlib.h>
-#include <stdio.h>
+#include "dbus.h"
```
tsipinakis: To ease readability, we keep the local import under the external ones with an empty line separating them. This happens on a lot of other files, I won't comment on all of them for this issue but they should also be fixed.

src/notification.h
```diff
@@ -2,7 +2,9 @@
 #ifndef DUNST_NOTIFICATION_H
 #define DUNST_NOTIFICATION_H
 
-#include "x.h"
+#include <glib.h>         // for gsize
```
tsipinakis: No need to comment what each include is for, if someone removes one that's needed the compiler will complain anyway.

johnchen902: Seems like I forgot to remove them from include-what-you-use output.

src/dbus.c
johnchen902: I don't feel like "dbus.h" is an import here, because "dbus.c" is the one implementing it. It's more like an export.

Anyway, if it's moved down, how should it be ordered relative to other local includes? Remain the first one or just sorted lexicographically?

tsipinakis: That's a good point, and looking into this further I found the argument that having the header corresponding to the c file at the top also tests that each header can be included without any other includes.

So on second thought, leave it as is but leave an empty line between include layers i.e.
```
own header include
(empty line)
external includes
(empty line)
local includes
```

johnchen902:
Seems like `getline` has no glib counterpart, so there is still a `getline`-`free` pair in option_parser.c

tsipinakis:
> Seems like `getline` has no glib counterpart, so there is still a `getline`-`free` pair in option_parser.c

That's fine, since the `getline` and the `free` call are in the same function it's pretty easy to manage.

This is pretty much ready to be merged but I recommend rebasing this in the spirit of keeping the commit history clean. Specifically
* Squash `ab2891` and `4abe813` into `960814`
* Remove the merge commit and rebase the entire branch on top of master.

tsipinakis:
Thanks!

PR #295, opened 2017-02-15
Allow dunst to be managed by systemd

Stebalien:
This should have no effect on systems without systemd; it just causes dunst to be started through systemd when possible.


tsipinakis:
Thanks!

PR #296, opened 2017-02-16
Makefile: just be verbose

quite:
Show what is happening during build etc, the way most other make-based
projects do.

There is thus no need for the options target. Additionally, this target
caused unnecessary rebuilds of ./dunst each and every time, even on
`make install`.


tsipinakis:
Agreed, there is no need to be fancy here.

PR #297, opened 2017-02-18
Modernize the systemd unit a bit

Stebalien:
* Add `Documentation` field.
* Make it `WantedBy` `default.target` (`multi-user.target` isn't a valid user target).
* Make it a dbus unit (consider it active when it acquires the dbus interface).


tsipinakis:
Thanks! Since I don't use systemd to manage dunst I'm merging only based on what I know from the [unit file spec](https://www.freedesktop.org/software/systemd/man/systemd.unit.html) trusting that you tested it to work as intended.

PR #298, opened 2017-02-20
option_parser: Allow comments on lines with quoted strings

LukeShu:
This is an updated version of #126 (2013).

The current behavior is
 - If the value contains a double-quote:
   - 1. Verify that it must contains at least two quotes.
   - 2. If one of the quotes is the first character, trim it.
   - 3. If one of the quotes is the last character, trim it.
 - Else:
   - 1. Trim a trailing comment from the value.

This has the effect that

    `key = "value" # comment` => `value " #comment`

This is surprising and almost certainly not what the user wants.

However, it allows simple nested quotes like:

    `key = "A string "with quotes""` => `A string "with quotes"`

Adjust the brokenness of the first example at the expense of breaking the
second.  A user seeking that value will now have to type:

    key = "A string \"with quotes\""

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -368,7 +368,7 @@ int notification_init(notification * n, int id)
 
         n->urls = notification_extract_markup_urls(&(n->body));
 
-        n->msg = string_replace_all("\\n", "\n", g_strdup(n->format));
+        n->msg = g_strdup(n->format);
         n->msg = notification_replace_format("%a", n->appname, n->msg,
                 MARKUP_NO);
         n->msg = notification_replace_format("%s", n->summary, n->msg,
--- a/src/option_parser.c
+++ b/src/option_parser.c
@@ -205,11 +205,14 @@ char *clean_value(char *value, int line_num)
                                 unparsed++;
                                 break;
                         default:
-                                // Unrecognized backslash sequence;
-                                // treat the backslash as an ordinary character.
-                                // Consider issuing an error or warning here instead.
                                 unparsed++;
-                                break;
+                                fprintf(stderr,
+                                       "Warning: invalid config file at line %d\n",
+                                       line_num);
+                                fprintf(stderr,
+                                       "Unrecognized backslash sequence '\\%c'\n",
+                                       *unparsed);
+                                return NULL;
                         }
                         break;
                 case '#':
--- a/src/option_parser.c
+++ b/src/option_parser.c
@@ -198,14 +198,12 @@ char *clean_value(char *value, int line_num)
                         in_quote = !in_quote;
                         break;
                 case '\\':
-                        switch (unparsed[1]) {
+                        memmove(unparsed, unparsed + 1, strlen(unparsed));
+                        switch (*unparsed) {
                         case '\\':
                         case '"':
-                                memmove(unparsed, unparsed + 1, strlen(unparsed));
-                                unparsed++;
                                 break;
                         default:
-                                unparsed++;
                                 fprintf(stderr,
                                        "Warning: invalid config file at line %d\n",
                                        line_num);
@@ -214,6 +212,7 @@ char *clean_value(char *value, int line_num)
                                        *unparsed);
                                 return NULL;
                         }
+                        unparsed++;
                         break;
                 case '#':
                 case ';':
--- a/src/option_parser.c
+++ b/src/option_parser.c
@@ -203,6 +203,9 @@ char *clean_value(char *value, int line_num)
                         case '\\':
                         case '"':
                                 break;
+                        case 'n':
+                                *unparsed = '\n';
+                                break;
                         default:
                                 fprintf(stderr,
                                        "Warning: invalid config file at line %d\n",
--- a/src/option_parser.c
+++ b/src/option_parser.c
@@ -29,7 +29,7 @@ static section_t *new_section(char *name);
 static section_t *get_section(char *name);
 static void add_entry(char *section_name, char *key, char *value);
 static char *get_value(char *section, char *key);
-static char *clean_value(char *value);
+static char *clean_value(char *value, int line_num);
 
 static int cmdline_argc;
 static char **cmdline_argv;
@@ -91,7 +91,7 @@ void add_entry(char *section_name, char *key, char *value)
         int len = s->entry_count;
         s->entries = g_realloc(s->entries, sizeof(entry_t) * len);
         s->entries[s->entry_count - 1].key = g_strdup(key);
-        s->entries[s->entry_count - 1].value = clean_value(value);
+        s->entries[s->entry_count - 1].value = g_strdup(value);
 }
 
 char *get_value(char *section, char *key)
@@ -186,20 +186,50 @@ int ini_get_bool(char *section, char *key, int def)
         }
 }
 
-char *clean_value(char *value)
+char *clean_value(char *value, int line_num)
 {
-        char *s;
+        char *unparsed = value;
 
-        if (value[0] == '"')
-                s = g_strdup(value + 1);
-        else
-                s = g_strdup(value);
+        bool in_quote = false;
+        while ((unparsed = strpbrk(unparsed, "\"\\#;")) != NULL) {
+                switch (*unparsed) {
+                case '"':
+                        memmove(unparsed, unparsed + 1, strlen(unparsed));
+                        in_quote = !in_quote;
+                        break;
+                case '\\':
+                        switch (unparsed[1]) {
+                        case '\\':
+                        case '"':
+                                memmove(unparsed, unparsed + 1, strlen(unparsed));
+                                unparsed++;
+                                break;
+                        default:
+                                // Unrecognized backslash sequence;
+                                // treat the backslash as an ordinary character.
+                                // Consider issuing an error or warning here instead.
+                                unparsed++;
+                                break;
+                        }
+                        break;
+                case '#':
+                case ';':
+                        if (in_quote)
+                                unparsed++;
+                        else
+                                *unparsed = '\0';
+                        break;
+                }
+        }
+        if (in_quote) {
+                fprintf(stderr,
+                       "Warning: invalid config file at line %d\n",
+                       line_num);
+                fprintf(stderr, "Missing '\"'\n");
+                return NULL;
+        }
 
-        if (s[strlen(s) - 1] == '"')
-                s[strlen(s) - 1] = '\0';
-
-        return s;
-
+        return g_strstrip(value);
 }
 
 int load_ini_file(FILE * fp)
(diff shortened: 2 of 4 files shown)


tsipinakis:
Before I start reviewing this, can you update the timestamp on both this and #299 to todays date? They are practically entirely different commits at this point.

src/option_parser.c
```diff
@@ -187,20 +187,55 @@ int ini_get_bool(char *section, char *key, int def)
         }
 }
 
-char *clean_value(char *value)
+char *clean_value(char *value, int line_num)
```
tsipinakis: This entire function is way too complicated at a whopping 6 levels of indentation(!!!) for something that can probably be done in a few lines in the `load_ini_file` function.
What's more is that you assume that the first quote is always the start of the string, this means that something like `String \"with quotes\"` fails to parse. A better approach would be to check the first non-whitespace character of the value, if it's a quote then we can go hunting for the closing quote(preferably in a different function), if not we just unescape all quotes and return.

src/option_parser.c
```diff
@@ -187,20 +187,55 @@ int ini_get_bool(char *section, char *key, int def)
...
-        if (value[0] == '"')
-                s = g_strdup(value + 1);
-        else
-                s = g_strdup(value);
-
-        if (s[strlen(s) - 1] == '"')
-                s[strlen(s) - 1] = '\0';
-
-        return s;
+        char *unparsed = value;
```
tsipinakis: `unparsed` is overwritten on the next line, there is no reason for the assignment.

src/option_parser.c
```diff
@@ -187,20 +187,55 @@ int ini_get_bool(char *section, char *key, int def)
...
+        while ((unparsed = strpbrk(unparsed, "\"#;")) != NULL) {
+                char *closing_quote;
+                switch (*unparsed) {
+                case '"':
+                        // remove the opening quote
+                        memmove(unparsed, unparsed + 1, strlen(unparsed));
```
tsipinakis: Why not just update the pointer?

src/option_parser.c
```diff
@@ -249,25 +284,10 @@ int load_ini_file(FILE * fp)
                 }
 
                 *equal = '\0';
-                char *key = g_strstrip(start);
-                char *value = g_strstrip(equal + 1);
+                char *key = g_strdup(g_strstrip(start));
```
tsipinakis: Why call `g_strdup` here? It's better to leave it to `add_entry` to handle memory. (Even though it looks like `value` was `strdup`-ed here earlier, this is as good a time as any to change that)

src/option_parser.c
LukeShu: > This entire function is way too complicated at a whopping 6 levels of indentation(!!!) for something that can probably be done in a few lines in the `load_ini_file` function.

I agree, but if you look at other codebases, it's kind of par-the-course for quoting.

> What's more is that you assume that the first quote is always the start of the string, this means that something like String \"with quotes\" fails to parse

Oh! You are right.  TBH, I probably shouldn't have been pushing when I was so tired.  I'll simplify it a bit and fix that (by making it a bit more similar to the the original #126.

src/option_parser.c
LukeShu: Originally, I'd put the quote-handling code in `load_ini_file`, but it didn't work correctly; as I didn't realize that some of the logic was in `clean_value`; so then I put it all in one place; `clean_value`.  

Now, `clean_value` always did an internal `g_strdup`. `add_entry` called `g_strdup` for the key, and `clean_value` for the value.

However, I found it necessary to move `clean_value` from `add_entry` to `load_ini_file` so that an error from `clean_value` could inhibit `add_entry` from running.  When I did that, I also moved the `g_strdup(key)` for symmetry.

LukeShu:
I've just pushed a new version that I believe addresses your concerns.  It allows backslash sequences anywhere in a string now; before it only allowed them inside of quotes.

src/option_parser.c
```diff
@@ -250,22 +233,46 @@ int load_ini_file(FILE * fp)
...
+                                case '\\':
+                                case '"':
+                                        memmove(unparsed, unparsed + 1, strlen(unparsed));
+                                        unparsed++;
+                                        break;
+                                default:
+                                        // Unrecognized backslash sequence;
+                                        // skip over it without removing the backslash.
+                                        // Consider issuing an error or warning here instead.
+                                        unparsed += 2;
```
johnchen902: If the backslash is at the end of string, this will increment `unparsed` too much and leave next `strpbrk` undefined.

Add `-fsanitize=address` to `CFLAGS`, and use this config to reproduce:
```
[global]
    unused=\
```
Note that there is no trailing new line, i.e.
```
$ xxd /tmp/dunstrc 
00000000: 5b67 6c6f 6261 6c5d 0a20 2020 2075 6e75  [global].    unu
00000010: 7365 643d 5c                             sed=\
```

LukeShu: Oh! You are of course correct.  I was hasty when gutting out the initial version that considered this case to be an error.

LukeShu: I've now updated it to fix this.

LukeShu:
(just rebased on to master)

test/data/test-ini
```diff
@@ -21,7 +21,9 @@
 [string]
 	simple = A simple string
 	quoted = "A quoted string"
-	quoted_with_quotes = "A string "with quotes""
+	quoted_with_quotes = "A string \"with quotes\""
+	quoted_with_comment = "A quoted string" # a comment
+	escaped_quotes = String \"with quotes\"
```
tsipinakis: Tests are lacking a bit, can you expand them to cover the corner cases?
A few examples I can think of right now:
* A non-quoted string with a comment
* String with a comment that starts with a semicolon
* A quoted string that contains comment characters (`;#`)
* String that ends with a backslash
* Unknown escape sequences
* A non-quoted string that has quotes in the middle (e.g. `A "string#"` should yield `A string#`)

LukeShu: > * Unknown escape sequences

That would cement the lax behavior.  So I guess we should discuss it: Do you like the lax behavior of treating an unrecognized escape sequence as verbatim?  There's something to be said for strictness, and having that be an error.

LukeShu:
I've fixed the 2 code organization issues, and added more tests.  I also expanded the commit message a bit.

I did not add tests for:

> * String that ends with a backslash
> * Unknown escape sequences

because I'm not sure that the current unrecognized-backslash-sequence behavior is good; it's just the... null-behavior.  It could do something else there; issue an error or warning, have it treat the next character as verbatim (rather than treating the backslash as ordinary).

On the topic of backslashes, it might also make sense to handle `\n` here, rather than after-the-fact in `notification.c`.  Actually, it almost certainly does; `notification.c` doesn't allow the backslash to be escaped (`\\n` => `0x5C0x0A` rather than `0x5C0x6E`), while we have real backslash handling here.

tsipinakis:
> I hadn't added tests for that before; because I was considering it to undefined behavior (in the C-language sense of the term), and anything the program wants to do with it is valid.

I believe that even for the corner cases or things that should be undefined behaviour it is important to make sure that they are handled consistently across releases to avoid breaking running configs. If a new PR comes along and changes that behaviour, it makes it easy to spot if they are backwards incompatible.

> because I'm not sure that the current unrecognized-backslash-sequence behavior is good; it's just the... null-behavior. It could do something else there; issue an error or warning, have it treat the next character as verbatim

I'm open to suggestions but I think issuing an error here would be the correct course of action since we can't predict the behaviour the user expects.

> On the topic of backslashes, it might also make sense to handle \n here, rather than after-the-fact in notification.c

Agreed on that, can you change it? (In a new commit obviously)

It's important to note that we are talking about breaking changes here, so this PR will either need to be merged into `next` or be put on hold until the next release. The current schedule is to release a minor version once we have all the issues for the next [milestone](https://github.com/dunst-project/dunst/milestone/3) in order to give time to the users to migrate off of the deprecated settings.

LukeShu:
I've pushed added commits to treat unrecognized sequences as errors, and move the `\n` handling.

Should I do anything to rebase this against next?

src/option_parser.c
```diff
@@ -198,14 +198,11 @@ char *clean_value(char *value, int line_num)
...
-                        switch (unparsed[1]) {
-                        case '\\':
-                        case '"':
-                                memmove(unparsed, unparsed + 1, strlen(unparsed));
-                                unparsed++;
-                                break;
+                        memmove(unparsed, unparsed + 1, strlen(unparsed));
+                        switch (*unparsed) {
+                        case '\\': /* verbatim */ break;
+                        case '"': /* verbatim */ break;
```
tsipinakis: To improve readability, there shouldn't be anything else on a line with a switch case label. I think it's best to utilize fall-through on this as it was before.
```
case '\\':
case '"':
        break;
```

src/option_parser.c
```diff
@@ -202,6 +202,7 @@ char *clean_value(char *value, int line_num)
                         switch (*unparsed) {
                         case '\\': /* verbatim */ break;
                         case '"': /* verbatim */ break;
+                        case 'n': *unparsed = '\n'; break;
```
tsipinakis: Same for this, only one command per line and there shouldn't be anything else on the same line as a switch case label.

tsipinakis:
> Should I do anything to rebase this against next?

I've just updated `next` to point to the current `HEAD`. You shouldn't need to do anything else, I'll change the target branch to `next` before merging.

LukeShu:
I've made the stylistic changes.

I also added newline tests to the test suite; I'd meant to do that for the original commit, but apparently forgot.

tsipinakis:
> Added newline tests to the test suite; I'd meant to do that for the original commit, but apparently forgot.

We still don't have any tests that cover escaped backslashes, can you add some?

tsipinakis:
Looks like github added one of my commits to the PR diff after changing the target branch, not sure what's going on there.

LukeShu:
> We still don't have any tests that cover escaped backslashes, can you add some?

Done.

src/notification.c
```diff
@@ -368,7 +368,6 @@ int notification_init(notification * n, int id)
 
         n->urls = notification_extract_markup_urls(&(n->body));
 
-        n->msg = string_replace_all("\\n", "\n", g_strdup(n->format));
         n->msg = notification_replace_format("%a", n->appname, n->msg,
```
tsipinakis: `n->msg` is NULL here leading to a segfault, I assume you didn't notice that the line above it also called `g_strdup` on the format to assign an initial value.

LukeShu: Oh.  Yes.

I've fixed this now, and iverified that the program works.

tsipinakis:
I don't see anything else that can be improved so I'm merging this.
As said earlier, these changes will have to sit in `next` for a while until we plan to have the next major release due to backwards incompatible changes.

Thank you for your work!

PR #299, opened 2017-02-20
Expand "\\n" to "\n" in the format string only; not message text.

LukeShu:
This was originally submitted as #125, but it ended up being incorporated into #204, which was merged instead of my version. That's fine.

However, later, commit 3ef1506 (presumably accidentally) reverted back to the pre-#204 behavior.

So this PR re-applies that change.

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -267,7 +267,6 @@ char *notification_replace_format(const char *needle, const char *replacement,
 
         }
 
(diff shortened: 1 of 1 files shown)


LukeShu:
I've just now bumped the author date to match the committer date, per your request on #298.

tsipinakis:
Thanks!

PR #309, opened 2017-03-08
Tidy and fix markup handling.

LukeShu:
This has 4 commits: The first two tidy and reorganize the markup handling without making any functional changes. The last two commits each fix an issue with the markup handling that is now easy-to-spot in the reorganized code.

There are two user-visible changes here.  In one case, the behavior is straight-up a bug, and fixing it is OK.  In the other case, it improves the error-handling behavior; since an error is encountered anyway (and it says so on stdout), I believe that a change in behavior is fair game.  That is, I believe that this can be merged in to `master`, rather than `next`.

### DIFF
--- a/src/markup.c
+++ b/src/markup.c
@@ -1,0 +1,89 @@
+/* copyright 2013 Sascha Kruse and contributors (see LICENSE for licensing information) */
+
+#define _GNU_SOURCE
+#include "markup.h"
+
+#include <assert.h>
+#include <stdbool.h>
+
+#include "settings.h"
+#include "utils.h"
+
+/*
+ * Quote a text string for rendering with pango
+ */
+static char *markup_quote(char *str)
+{
+        if (str == NULL) {
+                return NULL;
+        }
+
+        str = string_replace_all("&", "&amp;", str);
+        str = string_replace_all("\"", "&quot;", str);
+        str = string_replace_all("'", "&apos;", str);
+        str = string_replace_all("<", "&lt;", str);
+        str = string_replace_all(">", "&gt;", str);
+
+        return str;
+}
+
+/*
+ * Strip any markup from text
+ */
+char *markup_strip(char *str)
+{
+        if (str == NULL) {
+                return NULL;
+        }
+
+        /* strip all tags */
+        string_strip_delimited(str, '<', '>');
+
+        /* unquote the remainder */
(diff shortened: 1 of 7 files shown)


src/markup.c
```diff
@@ -28,6 +25,35 @@ char *markup_quote(char *str)
         return str;
 }
 
+static
+char *markup_unquote(char *str)
+{
+        if (str == NULL) {
```
LukeShu: For the static functions, sure.  But my understanding of the rest of the code base isn't complete enough for me to be confident that we don't depend on passing NULL to markup_strip or markup_transform.

src/markup.c
```diff
@@ -60,24 +82,21 @@ char *markup_transform(char *str, enum markup_mode markup_mode)
                 return NULL;
         }
 
-        if (markup_mode == MARKUP_NO) {
+        switch (markup_mode) {
+        case MARKUP_NULL:
+                assert(false);
```
LukeShu: I agree with changing it to `markup != NULL`, but I like having it as a `case`.  Being able to look at the switch and see "yes, all of the values of the enum are handled" makes the code easier to read.

src/markup.c
tsipinakis: Alright, update the static functions then and I'll investigate later the possibility of doing the same thing in `markup_strip` and `markup_transform`.

src/markup.c
tsipinakis: Agreed, lets just update the `assert` call in place then.

src/markup.c
```diff
@@ -66,8 +66,10 @@ char *markup_strip(char *str)
         /* strip all tags */
         string_strip_delimited(str, '<', '>');
 
-        /* unquote the remainder */
+        /* We want to quote; but unquote first to avoid
+         * double-escaping anything already quoted. */
         str = markup_unquote(str);
+        str = markup_quote(str);
```
LukeShu: I was confused by your statement, then I realized that I misunderstood the code in `r_create_layout_from_notification`.

LukeShu:
I've made the requested changes, except that I haven't added tests yet.

LukeShu:
Ok, I've added some tests now.

LukeShu:
That's what I get for trying to push while someone is talking to me.

tsipinakis:
Merged, thanks for helping organize the mess that is the current codebase.

PR #312, opened 2017-03-30
Document actions and how to interact with them

hobarrera:
There's a lot of open issues regarding this, because the only source of information on this is a comment on an issue which isn't that easy to find.

Add the information to the man page. Let me know if you'd like to see something else mentioned.

### DIFF
--- a/docs/dunst.pod
+++ b/docs/dunst.pod
@@ -197,7 +197,7 @@ Don't timeout notifications popped up from history.
 
 =item B<-show_indicators>
 
-Show indicators for actions and URLs.
+Show indicators for actions and URLs. See ACTIONS below for futher details.
 
 =item B<-sort>
 
@@ -263,6 +263,22 @@ The progress value can be set with a hint, too.
 =item notify-send -h int:value:42 "Working ..."
 
 =back
+
+=head1 ACTIONS
+
+Dunst allows notifiers (i.e.: programs that send the notifications) to specify
+actions. Dunst has support for both displaying indicators for these, and
+interacting with these actions.
+
+If "show_indicators" is true and a notification has an action, an "(A)" will be
+prepended to the notification format. Likewise, an "(U)" is preneded to
+notifications with URLs. It is possible to interact with notifications that
+have actions regardless of this setting, though it may not be obvious which
+notifications HAVE actions.
+
+The "context" keybinding is used to interact with these actions, by showing a
+menu of possible actions. This feature requires "dmenu" or a dmenu drop-in
+replacement present.
 
 =head1 MISCELLANEOUS
 


docs/dunst.pod
```diff
@@ -264,6 +264,15 @@ The progress value can be set with a hint, too.
 
 =back
 
+=head1 ACTIONS
+
+Dunst allows interaction with actions in notifications. If "show_indicators" is
```
tsipinakis: This should probably also mention that actions are not user-configurable and that the program that sent the notification handles them.
Additionally, this sentence is weirdly structured, I think 
> Dunst allows programs to specify actions that the user can choose to perform.

should cover it with minimal confusion.
Maybe call 'programs' notifiers? Not sure.

docs/dunst.pod
```diff
@@ -264,6 +264,15 @@ The progress value can be set with a hint, too.
...
+true, a "(A)" will be appended to the notification title. The "context"
```
tsipinakis: > If "show_indicators" is true, a "(A)" will be appended to the notification title

`(A)` will be *prepended* not *appended* and not to the title, it will be prepended to the format.

This sentence also doesn't answer *when* it will be prepended.
> If one or more actions are available "(A)" will be prepended to the notification text

docs/dunst.pod
```diff
@@ -264,6 +264,15 @@ The progress value can be set with a hint, too.
...
+keybinding can be used to interact with these actions (regardless of the
+indicators being displayed).
+
+This feature requires "dmenu" to be installed.
```
tsipinakis: Should also mention that it doesn't *require* dmenu and it can also work with dmenu drop-in replacements if the `dmenu` setting is set to use them.

hobarrera:
> Sorry for nitpicking this but documentation is currently in shambles 

Yes. 😝 

> this PR is as good a place as any to start improving things.

Yup!

docs/dunst.pod
```diff
@@ -264,6 +264,22 @@ The progress value can be set with a hint, too.
 
 =back
 
+=head1 ACTIONS
+
+Dunst allows notifiers (eg: programs that fire the notifications) to specify
```
tsipinakis: You probably meant `i.e.` here instead of `eg` (`i.e.` can be used as `that is` , `in other words` and `eg` signifies an example) and please add the dots just to please the pendants :)

Also, `fire` may not be the best word to use here, in documentation being literal is usually the best so `send` should probably be used.

hobarrera:
> Oops, approved the changes without checking first.

That's the spirit!

Ready for another review though.

tsipinakis:
Thanks!

hobarrera:
Thank you! :D

PR #315, opened 2017-04-05
added body-hyperlinks to capabilities as it is supported

Tadly:
Not important but as it's supported, why not add it :)

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -110,6 +110,7 @@ static void on_get_capabilities(GDBusConnection * connection,
         builder = g_variant_builder_new(G_VARIANT_TYPE("as"));
         g_variant_builder_add(builder, "s", "actions");
         g_variant_builder_add(builder, "s", "body");
+        g_variant_builder_add(builder, "s", "body-hyperlinks");
 
         if(settings.markup != MARKUP_NO)
                 g_variant_builder_add(builder, "s", "body-markup");


tsipinakis:
Sorry for the delay in responding to this, this week was busier than expected.

While checking the hyperlink capability I discovered a bug that causes a double free and crash to happen when opening a url. I don't think advertising that we have a feature that causes a crash each time it's used is a good idea so I'm putting this one on hold for now until that's fixed.

Tadly:
That's fine.
Actually, I was about to add support for images (basically the exact same thing as for the browser) but than realized it kinda is already there.
Especially, if you use `xdg-open` rather than an actual browser it integrates rather well.

What I'm trying to say is, I'm somewhat unsure about the supported capability now.

Do you think `body-images` should be part of the capabilities as well?

tsipinakis:
> Do you think body-images should be part of the capabilities as well?

No, at least not with the current implementation. Sending an image tag right now generates an error for an unknown tag
```
Error parsing markup: Unknown tag 'img' on line 1 char 58
```
The fact that it *techically* works is simply a side effect from the way we handle markup errors.

A proper image implementation in my opinion would include the following:
* Image alt text should be included in the body with an indication that it is an image(similar to how urls are handled)
* Images should be also indicated as such in the context menu

But looking at the url code and the code responsible for parsing the response from dmenu I feel it needs a complete rewrite. We shouldn't try to parse the response from dmenu to extract the url from it but instead have the options we passed to dmenu cached along with a dictionary with the actual urls and do a lookup.

Basically unless you have the time and energy to help refactor that part of the code I'd suggest you hold off on that for now.

tsipinakis:
Weird, coming back to this after a few weeks of inactivity I can't seem to be able to reproduce the issue mentioned above. I'll merge this and make a note to look into it.

Thanks!

Issue #265, opened 2016-06-20
Application name based Geometry


coxley:
Hi all,

I have a use-case that may not be super common but would be very helpful for me. I have some applications that I would like to use different geometry than the global `geometry` variable (e.g, `0x5-30+20` so these applications aren't width-limited) but others that I want to follow the global value.

Reason for this is some application notifications have a specific format that only looks right with a monospaced font and width is unpredictable. Like lining up columns.

Is this something possible or easy to add?

neg-serg:
I am really want this feature too!
Consider that we have mpd-notification like here:
https://i.imgur.com/MZsbKOR.png

It would be very glad to see another notifications in the different area.

quite:
I agree that we must be able to set geometry dynamically. I'd say what is needed is ability to use `geometry` in category sections.

Lundmore:
I too have a use-case for this feature. It would make dunst much more dynamic!

Flurrywinde:
It seems like this feature isn't implemented yet? Too bad! I too have a use-case.

Today, I decided to experiment, and I discovered one can do something like: `xdotool search --class Dunst windowmove 0 0` to move it.

However, the next time it opens, it'll be in that place again.

However, you can move it back to its default location later (like after calling `sleep`), and it doesn't have to be visible at the time.

I hope this helps some people (and doesn't have side-effects I haven't discovered yet).

s0la:
Any progress on this?

I use dunst quite heavily and I could greatly benefit from such a feature, given I use it with a lot of different custom applications, with a different placement, dimensions and format.. Right now, I basically restart dunst whenever I need a different configuration, but I have one usecase where it acts as a "visual representation" for windows resizing using keyboard, and killing and starting introduces a bit of latency, so the whole process doesn't seem that smooth..

I'd be really interested in this feature as well! :)

bynect:
> Any progress on this?
> 
> I use dunst quite heavily and I could greatly benefit from such a feature, given I use it with a lot of different custom applications, with a different placement, dimensions and format.. Right now, I basically restart dunst whenever I need a different configuration, but I have one usecase where it acts as a "visual representation" for windows resizing using keyboard, and killing and starting introduces a bit of latency, so the whole process doesn't seem that smooth..
> 
> I'd be really interested in this feature as well! :)
> 
> EDIT:
> 
[...]

PR #316, opened 2017-04-11
Add centering to -geometry flag

seletskiy:
It seems from XParseGeometry documentation, that it's valid to have `=` sign as first character of spec, but it's not documented how it's handled. So, I propose to use `=` at the geometry spec as indicator, that notification window should be positioned in the center of monitor (if both X and Y offsets are 0). It makes possible to align notification windows so they are shown at the X-center of active monitor by specifying geometry like `=<width>x<height>+0-40` — display notification window at the X-center of monitor and 40 px from bottom.

### DIFF
--- a/src/x11/x.c
+++ b/src/x11/x.c
@@ -703,6 +703,14 @@ static void x_win_move(int width, int height)
                 y = scr->dim.y + xctx.geometry.y;
         }
 
+        if (settings.centering == CENTERING_HORIZONTAL || settings.centering == CENTERING_BOTH) {
+                x = scr->dim.x + (scr->dim.w - width) / 2 + xctx.geometry.x;
+        }
+
+        if (settings.centering == CENTERING_VERTICAL || settings.centering == CENTERING_BOTH) {
+                y = scr->dim.y + (scr->dim.h - height) / 2 + xctx.geometry.y;
+        }
+
         /* move and resize */
         if (x != xctx.window_dim.x || y != xctx.window_dim.y) {
                 XMoveWindow(xctx.dpy, xctx.win, x, y);
--- a/src/settings.c
+++ b/src/settings.c
@@ -533,6 +533,27 @@ void load_settings(char *cmdline_config_path)
                 "Always run rule-defined scripts, even if the notification is suppressed with format = \"\"."
         );
 
+        {
+                char *c = option_get_string(
+                        "global",
+                        "centering", "-centering", "off",
+                        "Align notifications on screen off/horizontal/vertical/both"
+                );
+
+                if (strcmp(c, "off") == 0)
+                        settings.centering = CENTERING_OFF;
+                else if (strcmp(c, "horizontal") == 0)
+                        settings.centering = CENTERING_HORIZONTAL;
+                else if (strcmp(c, "vertical") == 0)
+                        settings.centering = CENTERING_VERTICAL;
+                else if (strcmp(c, "both") == 0)
+                        settings.centering = CENTERING_BOTH;
+                else
+                        fprintf(stderr,
+                                "Warning: unknown centering option: %s\n", c);
+                g_free(c);
+        }
+
         /* push hardcoded default rules into rules list */
         for (int i = 0; i < LENGTH(default_rules); i++) {
                 rules = g_slist_insert(rules, &(default_rules[i]), -1);
--- a/src/settings.h
+++ b/src/settings.h
@@ -11,6 +11,7 @@ enum icon_position_t { icons_left, icons_right, icons_off };
 enum separator_color { FOREGROUND, AUTO, FRAME, CUSTOM };
 enum follow_mode { FOLLOW_NONE, FOLLOW_MOUSE, FOLLOW_KEYBOARD };
 enum markup_mode { MARKUP_NULL, MARKUP_NO, MARKUP_STRIP, MARKUP_FULL };
+enum centering { CENTERING_OFF, CENTERING_HORIZONTAL, CENTERING_VERTICAL, CENTERING_BOTH };
 
 typedef struct _settings {
         bool print_notifications;
@@ -72,6 +73,7 @@ typedef struct _settings {
         keyboard_shortcut close_all_ks;
         keyboard_shortcut history_ks;
         keyboard_shortcut context_ks;
+        enum centering centering;
 } settings_t;
 
 extern settings_t settings;


src/x11/screen.h
```diff
@@ -14,6 +14,7 @@ typedef struct _dimension_t {
         unsigned int w;
         int mask;
         int negative_width;
+        int center;
```
tsipinakis: This should probably be a `bool`.

src/x11/x.c
```diff
@@ -703,6 +703,14 @@ static void x_win_move(int width, int height)
                 y = scr->dim.y + xctx.geometry.y;
         }
 
+        if (xctx.geometry.center && xctx.geometry.x == 0) {
+                x = (scr->dim.x + (scr->dim.w - width) / 2);
```
tsipinakis: Why is the assignment enclosed in parenthesis? They are not needed and add nothing to readability.

src/x11/x.c
```diff
@@ -703,6 +703,14 @@ static void x_win_move(int width, int height)
...
+        }
+
+        if (xctx.geometry.center && xctx.geometry.y == 0) {
+                y = (scr->dim.y + (scr->dim.y - height) / 2);
```
tsipinakis: You probably meant `scr->dim.h` on the average calculation here, `y` is the screen position and `h` the height.

seletskiy:
@tsipinakis: I'm not particularly happy with use of `=` either. So, maybe, it will be better solution to add two boolean configuration/command-line options like `center_x`, `center_y`? If so, I can rework patch.

src/x11/screen.h
seletskiy: I did it same as with `negative_width`, which also should be `bool`, I think.

src/x11/x.c
seletskiy: Yeah, sorry. It just copy/paste from similar `if` from above.

src/x11/x.c
seletskiy: Yes, of course. Sorry for that.

src/x11/screen.h
tsipinakis: Yes, looks like `negative_width` should also be a bool.

tsipinakis:
> So, maybe, it will be better solution to add two boolean configuration/command-line options like center_x, center_y?

Even better, let's add a single `centering` setting that can be set to `none` (default), `horizontal`, `vertical`, or `both`. I have a suspicion that this might complicate the `geometry` setting a bit *too* much since now the behaviour of the x/y offset depends on *yet another* setting but it's the best solution I could come up with.

seletskiy:
@tsipinakis: Check it out again, please.

seletskiy:
@tsipinakis: fixed.

tsipinakis:
I've delayed reviewing this because something is still bugging me with the way this affects `geometry`, which might end up in some setting deprecations to redesign that part. I am merging this into a topical `centering` branch, I'll make sure this is merged in one way or the other before the next release.

seletskiy:
@tsipinakis: Will you merge this into master?

tsipinakis:
Will merge as part of #336.

Still in a bit of a hiatus due to an apartment move but development should start picking up steam by next week or so when everything has settled down on my end.

rightaway:
Will this work to put to the center of the screen multiple notifications in real time? For example, if I have the first notification and it pops up in the center, then there's a second notification, will the second one just appear below the first one? (So the first one is centered but not the second?) Or will the second one appear and both together will be re-centered?

tsipinakis:
@rightaway As it is now in the `centering` branch it's recentered after a new notification but I plan to have both as a setting, the `relative` centering which will recenter as the window size changes and `absolute` which will simply put the window in the center no matter how many notifications.

glubsy:
I'm also interested in a centered notification, perhaps only upon user request though.

Perhaps the best way to do this would be to make the "geometry" attribute mutable by user request, and add a flag to it to ask to center the notification window. (ie `notify-send "test" -h string:geometry:100x20-15+30centered ` or something like that)

Example: I want a specific application (xautolock) to have its notifications centered in the middle of my screen when it's about to lock my screen, but all others should be in the default location (in the top right for example). 

Would that be possible?

rightaway:
@glubsy Just wondering how do you get xautolock to send you a notification when it's about to lock your screen?

bebehei:
@rightaway [These are the `-notify` and `-notifier` options](https://github.com/bebehei/df/blob/7d3624b6d48703376e91542d2e8f994eb226a463/i3/lockscreen.sh#L55-L58).

glubsy:
@rightaway `xautolock -time 10 -notify 15 -notifier 'notify-send \"LOCKING IN 15 SECONDS\" -i \"dialog-warning\" -a \"my_screen_locker\"' -locker 'i3lock -c 220020' -detectsleep`

rightaway:
@bebehei @glubsy Thank you.

tsipinakis:
> Perhaps the best way to do this would be to make the "geometry" attribute mutable by user request, and add a flag to it to ask to center the notification window. (ie notify-send "test" -h string:geometry:100x20-15+30centered or something like that)

> Basically being able to override the geometry for a specific notification. Would that be possible? (I have a feeling this has been asked before, sorry about that)

It has, #265. The problem with this is that dunst was built around the assumption that everything is in a single window and changing this requires some serious refactoring. The initial stages of that refactor are in the `refactor-drawing` branch

glubsy:
Sounds great! I'll be watching this. Thank you. :)

rightaway:
@tsipinakis Any plans to merge the `centering` branch so that the `relative` feature can be used?

kovetskiy:
@tsipinakis any reaction?

bebehei:
@seletskiy @kovetskiy I actually don't understand @tsipinakis' objections to merge it into master. I'm probably missing some detail here.

But anyways, our complete X11 and drawing stuff is has to be refactored. As @tsipinakis has already mentioned, we'll merge this with #336. Sorry for this. We have to do the refactoring in calm and wisely. Any rash merge act will hinder us on the refactoring.

As a temporary solution, it might be best for you to checkout the gitrepo manually, rebase the centered branch onto master build your stuff from there.

tsipinakis:
Sorry for not responding earlier, I had the previous response marked as read without realising.

> I actually don't understand @tsipinakis' objections to merge it into master

My objection is that there is no point merging it now if it's going to be changed with a refactor later, I'd like to avoid having to add backwards compatibility logic and supporting a hastily merged setting.

kovetskiy:
@bebehei @tsipinakis hastly merged? This change was introduced half year ago — at Apr 11.

What is estimation time of arrival for this feature into master?

bebehei:
@kovetskiy I have to apologize, but our next thing on the TODO list is a full refactor of the whole x11 drawing stuff. Merging it **now** will require us later to add backwards compatibility.

So, let's fix it on the basis of the refactoring and not now to have later awkward backwards compatibility needed. The refactoring probably has an easier way to introduce a centered notification.

I have to apologize, that we hadn't been able to push it to master for half a year. But on the current point it doesn't make sense.

> What is estimation time of arrival for this feature into master?

We're rather debian here. It's done when it's done. I hope before April 10 ;-)

StephenBrown2:
I've been looking for a notification daemon that could support centered notifications (running a 39" 4K TV as my main monitor means corner notifications are hardly ever seen), and so I applied the changes in this PR to master, and the only real change is that the changes in x.c are now in draw.c:

```patch
index e1032c0..c953edc 100644
--- a/src/draw.c
+++ b/src/draw.c
@@ -593,7 +593,9 @@ static void calc_window_pos(int width, int height, int *ret_x, int *ret_y)
         struct screen_info *scr = get_active_screen();
 
         if (ret_x) {
-                if (settings.geometry.negative_x) {
+                if (settings.centering == CENTERING_HORIZONTAL || settings.centering == CENTERING_BOTH) {
+                        *ret_x = scr->x + (scr->w - width) / 2 + settings.geometry.x;
+                } else if (settings.geometry.negative_x) {
                         *ret_x = (scr->x + (scr->w - width)) + settings.geometry.x;
                 } else {
                         *ret_x = scr->x + settings.geometry.x;
@@ -601,7 +603,9 @@ static void calc_window_pos(int width, int height, int *ret_x, int *ret_y)
         }
 
         if (ret_y) {
-                if (settings.geometry.negative_y) {
+                if (settings.centering == CENTERING_VERTICAL || settings.centering == CENTERING_BOTH) {
+                        *ret_y = scr->y + (scr->h - height) / 2 + settings.geometry.y;
+                } else if (settings.geometry.negative_y) {
                         *ret_y = scr->y + (scr->h + settings.geometry.y) - height;
                 } else {
                         *ret_y = scr->y + settings.geometry.y;
diff --git a/src/settings.c b/src/settings.c
index 46187e2..f4050cd 100644
--- a/src/settings.c
+++ b/src/settings.c
@@ -659,6 +659,27 @@ void load_settings(char *cmdline_config_path)
                 "Always run rule-defined scripts, even if the notification is suppressed with format = \"\"."
         );
 
+        {
+                        "centering", "-centering", "off",
+                        "Align notifications on screen off/horizontal/vertical/both"
+                );
+
+                if (strcmp(c, "off") == 0)
+                        settings.centering = CENTERING_OFF;
+                else if (strcmp(c, "horizontal") == 0)
+                        settings.centering = CENTERING_HORIZONTAL;
+                else if (strcmp(c, "vertical") == 0)
+                        settings.centering = CENTERING_VERTICAL;
+                else if (strcmp(c, "both") == 0)
+                        settings.centering = CENTERING_BOTH;
+                else
+                        fprintf(stderr,
+                                "Warning: unknown centering option: %s\n", c);
+                g_free(c);
+        }
+
         /* push hardcoded default rules into rules list */
         for (int i = 0; i < G_N_ELEMENTS(default_rules); i++) {
                 rules = g_slist_insert(rules, &(default_rules[i]), -1);
diff --git a/src/settings.h b/src/settings.h
index 6bcf6b8..b10d9fc 100644
--- a/src/settings.h
+++ b/src/settings.h
@@ -9,6 +9,7 @@
 #include "x11/x.h"
 
 enum alignment { ALIGN_LEFT, ALIGN_CENTER, ALIGN_RIGHT };
+enum centering { CENTERING_OFF, CENTERING_HORIZONTAL, CENTERING_VERTICAL, CENTERING_BOTH };
 enum ellipsize { ELLIPSE_START, ELLIPSE_MIDDLE, ELLIPSE_END };
 enum icon_position { ICON_LEFT, ICON_RIGHT, ICON_OFF };
 enum separator_color { SEP_FOREGROUND, SEP_AUTO, SEP_FRAME, SEP_CUSTOM };
@@ -83,6 +84,7 @@ struct settings {
         struct keyboard_shortcut close_all_ks;
         struct keyboard_shortcut history_ks;
         struct keyboard_shortcut context_ks;
+        enum centering centering;
         bool force_xinerama;
         int corner_radius;
         enum mouse_action mouse_left_click;
```

Issue #314, opened 2017-04-03
Figure out how to handle running dunst as a systemd service


tsipinakis:
In a [comment](https://github.com/dunst-project/dunst/issues/240#issuecomment-291012334) @nmschulte mentioned that the dbus service file fails to launch dunst if the service file is not installed.

This only happens if dbus is started with the `--systemd-activation` flag and it looks like in that case if the service file is not found it ignores the `Exec` line. We should either
* Remove the `SystemdService` line from the dbus service file, essentially not allowing dunst to be managed by systemd by default or
* Consider officially supporting systemd installation and install the unit file by default.

As an additional sidenote that is loosely within the scope of this issue, @mathstuf linked an issue in [the same thread](https://github.com/dunst-project/dunst/issues/240#issuecomment-291129486) that apparently systemd dbus units look in the system bus rather than the session bus as they should.
We should 
* Verify the validity of the report
* Decide if we want to change the systemd unit type to simple instead of dbus

nmschulte:
We could: comment out the `SystemdService` line in the service file, perhaps with a relevant comment above it, and also add a note to the build process (`make install` target, or `install-service` target) about the optional (non-`install`ed`) `dunst.systemd.service` file (with a note that it needs to be renamed to match the assumption in the D-Bus service file).

mathstuf:
Setting `PATH` shouldn't be necessary since systemd requires full paths to the executable in the `Exec*` lines.

nmschulte:
While it's a bit heavy handed (read: the configuration affects all of systemd's user units, not just D-Bus/Dunst's environment), it's an important detail for those that run scripts with Dunst (Dunst supports running scripts keyed by notification attribute matching) and expect their `PATH` from e.g. `~/.profile` (this is called... an login (or interactive?) session? I'm not sure the proper terminology.).

One could deal with `PATH` issues in each script too, but there's no saying that the Dunst configuration (an thus the scripts that care about `PATH`) can't be used for multiple Dunst instances (though... maybe this can't happen; I thought I saw a use-case for this in my testing, but I may have just misunderstood).

mathstuf:
[...]

PR #324, opened 2017-05-10
Tweak systemd.service

hobarrera:
Hardcoding a DISPLAY variable is a bad idea because it means that dunst
will only run on X11, and only for that specific configuration.

FWIW, this actually applies to any desktop app run via systemd:

1. `DISPLAY` should be set by `systemctl --user set-environment` elsewhere.
2. `DISPLAY=:0` is not universally valid.
3. This breaks dunst if attempting to run wayland, and requires manually
  starting it (eg: not via systemd), or editing the file.

See [this comment](https://github.com/dunst-project/dunst/issues/314#issuecomment-300527201) for context.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,7 @@ install-doc:
 install-service: service
 	mkdir -p ${DESTDIR}${PREFIX}/share/dbus-1/services/
 	install -m644 org.knopwob.dunst.service ${DESTDIR}${PREFIX}/share/dbus-1/services
+	install -Dm644 dunst.systemd.service ${DESTDIR}${PREFIX}/lib/systemd/user/dunst.service
 
 install: install-dunst install-doc install-service
 
--- a/contrib/dunst.systemd.service.in
+++ b/contrib/dunst.systemd.service.in
@@ -1,6 +1,7 @@
 [Unit]
 Description=Dunst notification daemon
 Documentation=man:dunst(1)
+PartOf=graphical-session.target
 
 [Service]
 Type=dbus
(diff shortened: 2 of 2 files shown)


hobarrera:
Bump?

tsipinakis:
First of all sorry for the delay, as mentioned earlier I won't have a lot of time to work on dunst for the next few weeks so progress is slow.

I agree with removing the display environment line from the service file but looking at  the `systemd.special(7)` manpage it looks like we should also add `PartOf=graphical-session.target` in the unit section to correctly signal that dunst needs a graphical environment to work properly.

hobarrera:
Huh, I was completely unaware of `graphical-session.target`. I'm actually not using it (oh, my setup is a lie! 😛 ). I'm curious though; when I kill the X server (or wayland), should something somehow stop `graphical-session.target`? It's somehow unclear how to stop a unit file when a process dies (and running X via systemd [seems to be broken](https://wiki.archlinux.org/index.php/Systemd/User#Xorg_as_a_systemd_user_service) right now).

I'll update the PR when I get a chance, but I would really like to investigate it managing such a target is actually feasable in practice.

hobarrera:
PR updated. I've added the `PartOf=graphical-session.target`, but it's unclear how `graphical-session.target` is to be stopped (that's left up to the system integrator).

The service file is now also installed by default.

tsipinakis:
PR is fine, you might want to fix the typo in the commit body though.
> Make dunst part of graphical-session.target, so that it exist immediately
when the session exists.

exists -> exits

hobarrera:
Fixed the typo. Thanks for pointing it out.

tsipinakis:
Thanks!

Issue #234, opened 2015-08-15
Clickable buttons for actions?


Xylemon:
I know dunst has a context menu that can be activated by a key combination, but I honestly miss how other notification daemons had clickable buttons inside the notification itself. So could there be an option for buttons on notifications? I don't care if you do it via GTK or some other method, anything that just shows the actions and buttons for them is enough for me.

demonking:
What do you mean ?

Maybe you mean something like a systray?
E.g some icon, where you click on it and the messaglist appears?

Xylemon:
I would desire something similar to other notification daemons like xfce4-notifyd where there is a button on the notification itself for an action (or actions) that you can click. An image I found off a quick Google search to show what I mean: https://3.bp.blogspot.com/-vfPYETQBckQ/UiOXdAZYQdI/AAAAAAAAKnM/x1MdKs3ygBw/s1600/nuvola3.png

Another example of a program that you could test this out is Pidgin, as with other notification daemons, when someone would message me, I would always have a "Show" button to click on and it would bring up the message window.

The other notification daemons utilized GTK2 for the buttons, but dunst doesn't have to go that far. Just an option to have these buttons would satisfy me.

CyberShadow:
I think being able to click the notification to bring up the action menu (i.e. allow clicks to do exactly the same thing as Ctrl+Shift+. does now) would be a good simple improvement.

miseran:
As a simpler alternative, perhaps a middle click on a notification with an associated action or URL should invoke it. This would only really make sense if there is just a single associated action, or if some action could be determined as the default. But even if it only works in the simple case of one action, it would be very convenient in my opinion.

I would be willing to write a pull request for this.

tsipinakis:
> if some action could be determined as the default

Quoting directly from the notification spec:

> The default action (usually invoked my clicking the notification) should have a key named "default".

I'd gladly accept a PR implementing this if it doesn't break backwards compatibility.

I think the best way to handle this is to have middle click invoke the default action if available and then attempt to open the context menu if not.

I'll also consider implementing the button feature, but that's more on the long term and there are still some issues that need to be solved before any serious work on new feature implementation is done.

sistematico:
Anyway to change middle click to left click?

tsipinakis:
Not from the configuration but if you're fine with patching it yourself this should work:
```
diff --git a/src/x11/x.c b/src/x11/x.c
index 1ca3243..2b2d8bd 100644
[...]

PR #330, opened 2017-06-28
Allow middle click on notification to invoke action

miseran:
A middle click on a notification with a single or default action will
invoke it. If there are multiple actions and no default, the context
menu is opened. If there are no actions, proceed similarly with URLs.

As discussed in #234.

I decided to always prioritize actions over URLs, since that seems more likely to be what the user wants.

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -588,4 +588,32 @@ int notification_get_ttl(notification *n) {
 int notification_get_age(notification *n) {
         return time(NULL) - n->timestamp;
 }
+
+/*
+ * If the notification has exactly one action, or one is marked as default,
+ * invoke it. If there are multiple and no default, open the context menu. If
+ * there are no actions, proceed similarly with urls.
+ */
+void notification_do_action(notification *n) {
+        if (n->actions) {
+                if (n->actions->count == 2) {
+                        action_invoked(n, n->actions->actions[0]);
+                        return;
+                }
+                for (int i = 0; i < n->actions->count; i += 2) {
+                        if (strcmp(n->actions->actions[i], "default") == 0) {
+                                action_invoked(n, n->actions->actions[i]);
+                                return;
+                        }
+                }
+                context_menu();
+
+        } else if (n->urls) {
+                if (strstr(n->urls, "\n") == NULL)
+                        open_browser(n->urls);
+                else
+                        context_menu();
+        }
+}
+
 /* vim: set tabstop=8 shiftwidth=8 expandtab textwidth=0: */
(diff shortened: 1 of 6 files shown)


src/notification.c
```diff
@@ -588,4 +588,32 @@ int notification_get_ttl(notification *n) {
...
+                        if (strcmp(n->actions->actions[i], "default") == 0) {
+                                action_invoked(n, n->actions->actions[i]);
+                                return;
+                        }
+                }
+                context_menu();
+
+        } else if (n->urls) {
+                if (strstr(n->urls, "\n") == NULL)
+                        open_browser(n->urls);
```
tsipinakis: Apparently, urls that were parsed from markup tags (e.g. `<a href="http://www.google.com">Link</a>`), they are in the form of `[#n] url`.

tsipinakis:
And as a (optional) sidenote: Unless there's a reason for it, it's usually considered good etiquette to have your name and/or a valid email address in your commit. See the [git documentation](https://git-scm.com/book/tr/v2/Customizing-Git-Git-Configuration)

miseran:
- The button release passing to the window below also happened with left and right clicks, as I found out with xev. Hence, I changed all mouse clicks to be on release.
- I don't have a solution to the problem with multiple middle clicks either. It doesn't happen with rofi, which closes if you click anywhere else.
- I moved the code that removes the `[#1]` prefix to the open_browser function, so it applies to middle clicks as well. In fact, it appears `<a href>` links were actually broken and would crash dunst, which seems to be fixed with my commit. Apparently it was a bad free, but I don't understand this very well.

I should say that I'm not very experienced with C, but I hope my code is acceptable anyway.

tsipinakis:
> In fact, it appears `<a href>` links were actually broken and would crash dunst, which seems to be fixed with my commit. Apparently it was a bad free, but I don't understand this very well.

I had this issue in mind but didn't get around to fixing it yet, guess this resolves that too.

Looking at how the code was previously, looks like it's because the `in` pointer is re-assigned to the result of the `strchr` call and it crashes since it's an error to call `free` with a pointer that doesn't point to the beginning of an allocated block.

> I moved the code that removes the `[#1]` prefix to the open_browser function, so it applies to middle clicks as well.
> I should say that I'm not very experienced with C, but I hope my code is acceptable anyway.

I *really* don't like this solution but it's not a problem with your code, it's a problem with the implementation of urls and actions in dunst generally, and looking at it, this is probably the best that can be done right now. I'm slowly working towards refactoring it into a simpler and hopefully more expandable system but it takes time as usual.

I'm merging this as-is then. Thank you for your work.

Issue #263, opened 2016-05-14
Please consider tagging a release


Stebalien:
There have been quite a few bug fixes since the last release.

Stebalien:
:crying_cat_face:  :pray:

dubhater:
Surely it's not that difficult to type 'git tag -a v2' or whatever?

Stebalien:
To be fair, they're probably waiting to clear the issue backlog.

Stebalien:
Probably the [https://github.com/dunst-project/dunst/milestone/3](2.0 milestone).

jplatte:
Yeah, it's a bit annoying to have to use the -git package just to not have big icons completely mess up the notifications..

hobarrera:
Can we at least have a 1.2?

The missing features an bugs in the last stable is ridiculous, even if the current master has bugs, 1.1 also has so many issues that it's a pain. Be honest: do you ever except to **completely** clean up all GH issues? (no offence intended here, thanks for the cool work! 😗  )

tsipinakis:
> Be honest: do you ever except to completely clean up all GH issues? (no offence intended here, thanks for the cool work! 😗 )

No, the [1.2](https://github.com/dunst-project/dunst/milestone/3)(previously named 2.0) milestone contains the issues that I consider blockers for a new release and I also have a few work-in-progress local branches that I want merged before the next version(doc rework, and a few config file changes/deprecations).

Additionally, I am currently in a month-long hiatus that'll last till the middle of June so I don't currently have the time to do the work needed to prepare a release.

tsipinakis:
Last issue in the 1.2 milestone was closed. Current todo list for the release is the following:

- [ ] Test the `xinerama-flag` branch
- [ ] Merge `xinerama-flag` to master
- [ ] Do a final check to verify full backwards compatibility with 1.1
- [ ] ... release?

bebehei:
Having alltogether ~800 commits on master, while [240 of them have never been released](../compare/v1.1.0...master), I doubt that there is a solution to ensure full backwards compatibility.

What about just releasing 1.2? People who use 1.1 currently will report issues anyway, because so much stuff has changed. And all others, who have depended on a fix compile it either from source or use something like [dunst-git package](https://aur.archlinux.org/packages/dunst-git).

tsipinakis:
> Having alltogether ~800 commits on master, while 240 of them have never been released, I doubt that there is a solution to ensure full backwards compatibility.

[...]

PR #332, opened 2017-07-07
limit length of single fields in notification

### LINKED ISSUE #248: No character limit with Mumble
For some reason, certain user messages or own messages can be as long as possible in dunst when using it with Mumble. The result is your entire system will freeze till the notification goes away. A great example of this is when someone sends an image in the chat, it displays all the characters that make up the encrypted image (which is thousands of characters).

bebehei:
This would be my 2 cents to fix #248 and remove the last blocker to have a new release.

This PR adds the option `max_fieldlength` to have a maximum of specific chars per field. So overlong messages will get truncated to the specified length.

Also if it's not wanted to have a limit, a minus-value can be chosen to skip the cutting.

I'm not sure yet, what a reasonable default would be. I've used `100`, but I'd like to have other's opinions here. Alternatively `-1` (=> no limit) would also be a good default.

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -329,6 +329,16 @@ int notification_init(notification * n, int id)
 
         n->msg = g_strchomp(n->msg);
 
+        /* truncate overlong messages */
+        if (strlen(n->msg) > DUNST_NOTIF_MAX_CHARS) {
+                char* buffer = g_malloc(DUNST_NOTIF_MAX_CHARS);
+                strncpy(buffer, n->msg, DUNST_NOTIF_MAX_CHARS);
+                buffer[DUNST_NOTIF_MAX_CHARS-1] = '\0';
+
+                g_free(n->msg);
+                n->msg = buffer;
+        }
+
         if (n->icon != NULL && strlen(n->icon) <= 0) {
                 g_free(n->icon);
                 n->icon = NULL;
--- a/src/notification.h
+++ b/src/notification.h
@@ -12,6 +12,8 @@
 #define NORM 1
 #define CRIT 2
 
+#define DUNST_NOTIF_MAX_CHARS 5000
+
 typedef struct _raw_image {
         int width;
         int height;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Text and icons are now centred vertically
 - Notifications aren't considered duplicate if urgency or icons differ
 - The frame width and color settings were moved to the global section as frame\_width and frame\_color respectively.
+- The maximum displayed field length is limited to 5000 characters
 
 ### Deprecated
 - `allow_markup` will be removed in later versions. It is being replaced by `markup`


tsipinakis:
I'm not sure a user visible setting is the best approach here. I doubt that truncating based on the number of characters in a notification has any real use case other than preventing the DoS that happens in #248.

I think a macro set to a high enough value that it's unlikely for anyone to bump into(Say 5000?) is the best and we can adjust that as needed.

bebehei:
I thought long about this. But yes, it's better to have this inside a macro and not expose it to the user.

IMHO a good notification is always less than 100 chars. If it's a long text, nevertheless you click on the notification to open the messenger and read it there. The introduced option would have made the truncation possible, which also prevents DoSing dunst.

But it just creates too much noise in `dunstrc`.

src/utils.c
```diff
@@ -7,6 +7,19 @@
 #include <stdlib.h>
 #include <string.h>
 
+char* string_cut_maxlen(char* string, int maxlen){
+        char *buffer;
+
+        if (maxlen < 0 || strlen(string) < maxlen) {
+                buffer = string;
+        } else {
+                size_t buffer_maxlen = maxlen;
```
tsipinakis: Why create a new variable instead of making the type of `maxlen` be `size_t` in the function definition? And with this change, there should be no need to check if `maxlen < 0` since `size_t` is unsigned.

src/utils.c
```diff
@@ -7,6 +7,19 @@
...
+                buffer = calloc(buffer_maxlen, sizeof(char));
```
tsipinakis: We're using glib so instead of `calloc` `g_malloc0` should be used but since we already know we are going to fill the entire buffer there is no reason to zero it out so `g_malloc` would probably be the correct function to use here.

src/utils.c
```diff
@@ -7,6 +7,19 @@
...
+                strncpy(buffer, string, buffer_maxlen);
```
tsipinakis: Quoting from `strcpy(3)`
> If there is no terminating null byte in the first n bytes of src, strncpy() produces an unterminated string in dest.

In other words at this point `buffer` is unterminated. The last character should be replaced with a null byte.

src/utils.h
```diff
@@ -9,6 +9,9 @@ char *string_replace_char(char needle, char replacement, char *haystack);
 char *string_replace_all(const char *needle, const char *replacement,
                          char *haystack);
 
+/* Cut the given string at buffer_maxlen */
```
tsipinakis: The parameter is named `maxlen` not `buffer_maxlen`

src/utils.c
bebehei: Thanks for pointing this out. My C-Skills are quite limited and I have not known the differences of all available allocation-functions yet.

src/utils.c
bebehei: Full ACK.

src/utils.c
bebehei: Good catch. Thanks.

src/utils.h
bebehei: Thanks.

bebehei:
Thanks for the feedback, @tsipinakis. I processed all requested changes.

src/notification.h
```diff
@@ -12,6 +12,8 @@
 #define NORM 1
 #define CRIT 2
 
+#define MAX_CHARS 5000
```
bebehei: @tsipinakis Is there any naming convention for macros? IMHO `MAX_CHARS` is too generic.

tsipinakis: Not really, but I do agree that it's too generic maybe `NOTIFICATION_MAX_LENGTH`?

src/notification.c
```diff
@@ -305,11 +305,11 @@ int notification_init(notification * n, int id)
         n->urls = notification_extract_markup_urls(&(n->body));
 
         n->msg = string_replace_all("\\n", "\n", g_strdup(n->format));
-        n->msg = notification_replace_format("%a", n->appname, n->msg,
+        n->msg = notification_replace_format("%a", string_cut_maxlen(n->appname, MAX_CHARS), n->msg,
```
tsipinakis: Oops, I missed this earlier but this creates a memory leak since the memory you allocate in `string_cut_maxlen` is never freed. Additionally I'm not sure if having a limit per-field is the correct option. In my opinion I'd change this to a single `string_cut_maxlen` call at the end of the format replacements.

Also, it should probably be placed after the icon processing, since we have flags to add the icon path to the format and a huge icon path would cause the same problem.

src/utils.h
```diff
@@ -2,13 +2,18 @@
 #ifndef DUNST_UTILS_H
 #define DUNST_UTILS_H
 
+#include <stdlib.h>
```
tsipinakis: `size_t` is defined in `stddef.h`. While `stddlib` includes `stddef` it's usually best practice to use the more specific header file.

src/utils.h
```diff
@@ -2,13 +2,18 @@
...
+
 /* replace all occurences of the character needle with the character replacement in haystack */
 char *string_replace_char(char needle, char replacement, char *haystack);
 
 /* replace all occurrences of needle with replacement in haystack */
 char *string_replace_all(const char *needle, const char *replacement,
                          char *haystack);
 
+/* Cut the given string at maxlen */
+char* string_cut_maxlen(char* string, size_t maxlen);
```
tsipinakis: Small nitpick, but I think `string_truncate` would be a better name for this function.

bebehei:
@tsipinakis Thank you for your feedback.

I dropped the stuff in `utils.c/h`, it limits the whole message size now.

Do we care about broken HTML if the output got truncated?

tsipinakis:
> I dropped the stuff in utils.c/h, it limits the whole message size now.

That's much better indeed, good call.

> Do we care about broken HTML if the output got truncated?

I don't think we should, any notification that reaches that limit will probably clip the monitor edges anyway. This is just a stopgap to stop dunst from freezing in case something messes up and sends much more data than it should.

bebehei:
Well, then will be there a 1.2 and close #263 ?

tsipinakis:
> Well, then will be there a 1.2 and close #263 ?

I posted an update in #263 with the status of the release.

Issue #327, opened 2017-06-18
GDK3 instead of GDK2?


hobarrera:
Is is possible to rely on / build using glib and gdk 3, rather than 2? How much work would upgrading compatibility imply?

hobarrera:
Oh, actually, changing `gdk-2.0` to `gdk-3.0` seems to work okay. I'm not sure what's up with `gio` though.

tsipinakis:
Latest stable version of glib and gio is 2.52.2, there is no version 3 as far as I know.

I'll look into switching to gdk 3 thought I'm not sure how beneficial it would be at this point.

tsipinakis:
Closed via 111fcaf5838a61faa2449c6edcdca163da5f287e

PR #333, opened 2017-07-13
depend on gtk3 instead of gtk2

bebehei:
in #263 there was mentioned to depend on gtk3 instead of gtk2. Mainly this is only incrementing the number, as it seems to work so far. Further testing **is** needed.

- [x] verify the behavior of the used functions did not change (or adapt our code)

Edit: seen, that there's already an issue open: #327

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ addons:
     - libpango1.0-dev
     - libcairo2-dev
     - libnotify-dev
-    - libgtk2.0-dev
+    - libgtk-3-dev
 dist: trusty
 sudo: false
 language: c
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,6 @@ test-clean:
 	rm -f test/test test/*.o
 
 dunstify: dunstify.o
-	${CC} ${CFLAGS} -o $@ dunstify.o $(shell pkg-config --libs --cflags glib-2.0 libnotify gdk-2.0)
+	${CC} ${CFLAGS} -o $@ dunstify.o $(shell pkg-config --libs --cflags glib-2.0 libnotify gdk-3.0)
 
 .PHONY: all clean dist install uninstall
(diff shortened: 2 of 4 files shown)


tsipinakis:
Oops, misclicked and requested a review from @Eizen. You can ignore that for now until the PR is marked as ready to be merged.

bebehei:
@tsipinakis This was a small misconception. Quite shortly before releasing 1.2, you changed the default value of max_icon_size (see: 79fec8a557adaf586d45e2661ed81b0bdee5e7cf) to unlimited. This change is also contained in here.

So while running my self-compiled dunst, e.g. Spotify-Icons had been very large and I blamed the gtk2->gtk3 transition for it.

But the transition works fine.

tsipinakis:
Awesome, since there are no modifications needed to support gtk3 maybe we should support both for now to avoid breaking the build configs, what do you think?

bebehei:
> Awesome, since there are no modifications needed to support gtk3 maybe we should support both for now to avoid breaking the build configs, what do you think?

- Which build configs exactly?
- How can we support both simultaneously? What?

tsipinakis:
> Which build configs exactly

I'm mostly talking about the users that compile manually (though I guess dependency problems are a given with that method).

> How can we support both simultaneously?

I was thinking about a simple if-else-if using pkg-config, try using `gdk-3.0`, if not available use `gdk-2.0`

I'm still undecided though whether this is a good idea or not.

tsipinakis:
Looking back, the same arguments that I gave against this in #317 also apply here, so this might not be a good idea after all.

bebehei:
> I'm still undecided though whether this is a good idea or not.

IMHO this would add to much complexity. GTK2 is deprecated, GTK3 is the evolution. If you have to run it with gtk2, you have to patch it.

tsipinakis:
I was in the middle on this decision but after looking a bit more into it I now think you're right that this adds complexity without much benefit.

Merged, thank you!

hobarrera:
Thanks!

PR #335, opened 2017-07-14
Add support for minimum notification window size.

Shaptic:
I like having my notification windows be expandable, but sometimes they're too small if there isn't enough text. This lets you specify a "minimum width" for the windows. It's only enabled for `-shrink` or dynamic width mode.

Parameter details:
  - `-nw`/`-notification-width` -- command line
  - `min_notification_width` -- config file

Naming suggestions are welcome.

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -30,6 +30,7 @@ int word_wrap = false;
 int ignore_newline = false;
 int line_height = 0;            /* if line height < font height, it will be raised to font height */
 int notification_height = 0;    /* if notification height < font height and padding, it will be raised */
+int min_notification_width = 0; /* if window width < min notification width, it will be raised */
 
 int separator_height = 2;       /* height of the separator line between two notifications */
 int padding = 0;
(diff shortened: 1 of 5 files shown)


tsipinakis:
Sorry for the extremely late response, I have been debating whether I want to merge this or not. I already half-heartedly included `notification_height` in the current release.

The problem I'm seeing is that a lot of the existing configuration settings have been built on top of each-other to satisfy a specific use case as it comes along with very little consideration as to the design. As a result, we now have a web of settings that depend on the values of other settings e.g. `shrink` depends on the value of width and now this is another setting to add on top of that.

I think before anything else is done on this front we're going to have to rethink the design of those settings.

tsipinakis:
See issue #336 for more discussion on this.

Shaptic:
Wow, I completely missed your response. I definitely agree with your decision to put more thought into the settings design; as I was writing the code myself, I could tell that things had been gradually patched on to support the various dimension-changing settings.

I'll close the PR and await the author's refactor. :) :+1:

PR #342, opened 2017-07-23
improve readability in dbus.c

bebehei:
Is there a coding styleguide for this project?

IMO readability goes over linelength here.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -152,125 +152,66 @@ static void on_notify(GDBusConnection * connection,
 
                         switch (idx) {
                         case 0:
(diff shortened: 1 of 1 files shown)


tsipinakis:
> Is there a coding styleguide for this project?

Not officially, unofficially we're loosely trying to follow the linux kernel code style but there are a few places where we deviate from it (typedefs, spaces instead of tabs etc). There has been some talk of migrating to tabs but it's probably best to leave it as is since it's going to cause conflicts for every fork and break `git blame`.

And I do agree that in this case, breaking down the lines actually hurts readability more than anything else.

Merged, thank you.

PR #344, opened 2017-07-23
RGB*A* support (fixes #289)

### LINKED ISSUE #289: Transparancy enabled colors
Dunst can be set to transparent in the config but this sets the whole window transparent. Many applications are able to use rgba-color-values to display colors accordingly. This has the advantage of being able to display the text non-transparent and the background transparently. That improves readability (vs. everything is transparent).

bebehei:
uses now rgba values instead of simple rgb. RGB is still the fallback, if no transparency value is given.

removes the transparency option (as it is redundant now).

fixes #289

### DIFF
--- a/config.h
+++ b/config.h
@@ -15,7 +15,6 @@ settings_t defaults = {
 .timeouts = { 10*G_USEC_PER_SEC, 10*G_USEC_PER_SEC, 0 }, /* low, normal, critical */
 .icons = { "dialog-information", "dialog-information", "dialog-warning" }, /* low, normal, critical */
 
-.transparency = 0,           /* transparency */
 .geom = "0x0",               /* geometry */
 .title = "Dunst",            /* the title of dunst notification windows */
 .class = "Dunst",            /* the class of dunst notification windows */
--- a/docs/dunst.pod
+++ b/docs/dunst.pod
@@ -160,13 +160,6 @@ This is used mainly in order to have the shrinking benefit of dynamic width (see
 geometry) while also having an upper bound on how long a notification can get
 before wrapping.
 
-=item B<transparency> (default: 0)
-
-A 0-100 range on how transparent the notification window should be, with 0
-being fully opaque and 100 invisible.
-
-This setting will only work if a compositor is running.
-
 =item B<notification_height> (default: 0)
 
 The minimum height of the notification window in pixels. If the text and
--- a/dunstrc
+++ b/dunstrc
@@ -38,11 +38,6 @@
     # width is 0.
     shrink = no
 
-    # The transparency of the window.  Range: [0; 100].
-    # This option will only work if a compositing window manager is
-    # present (e.g. xcompmgr, compiz, etc.).
-    transparency = 0
-
     # The height of the entire notification.  If the height is smaller
     # than the font height and padding combined, it will be raised
     # to the font height and padding.
--- a/src/settings.c
+++ b/src/settings.c
@@ -340,12 +340,6 @@ void load_settings(char *cmdline_config_path)
                 "horizontal padding"
         );
 
-        settings.transparency = option_get_int(
-                "global",
-                "transparency", "-transparency", defaults.transparency,
-                "Transparency. range 0-100"
-        );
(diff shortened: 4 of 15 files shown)


mbellgb:
Hi, thanks for working on this fix, it's something I was literally wondering about for my setup five minutes ago 😄 

I checked out your code to give it a test, and I'm not sure if it's quite working for me, the transparency doesn't seem to be working.

Config before:
```
[urgency_normal]
  background = "#d90e79"
  foreground = "#f1f1f1"
```

2017-07-23_145248_scrot

Config after:
```
[urgency_normal]
  background = "#d90e79aa"
  foreground = "#f1f1f1"
```

2017-07-23_145418_scrot

It seems like the transparency might not be getting parsed properly? Unless I'm not writing my config properly of course!

mbellgb:
Just to add that foreground transparency works fine, so I don't think it's the parsing that's the problem. Seems like it's actually the rendering that's going wrong somewhere...

bebehei:
Thanks for the feedback. I forgot to modify the drawing of the frame. The different color results of merging the frame color and the notification background.

Eizen:
The problem is the way the frame is implemented, one rectangle filled on top of another. One approach would be to use stroke instead of filling the rectangle behind but i'm not sure how much control one has over the borders. On another hand making the rectangle behind transparent probably doesn't play well if there's not compositor so it would be better not to draw anything there.

bebehei:
Thanks @Eizen . That would have been my first thought, too. But after using dunst now for multiple hours with the patch, I've found some other issues, too.

I have to fix them during the review process. This may take some time.

Eizen:
@bebehei If you need some help you can drop by `#dunst` irc channel on Freenode as i was going to work on this anyway. Otherwise take your time, no rush here.

bebehei:
After using this on my machine for about a week, I found multiple heavy issues in the chain.

I use compton with all settings on default.

### shadows with compton are ugly

Compton draws a shadow around every window. So the separator won't get shaded, which shows up quite ugly. (Separator set to transparent `#00000000`)

shadows-separator

To solve this, we have to draw every notification in a separate window.

### Different notifications heights kill the separator

separator
Or well, the separator will get drawn, but it its transparency will use the previous notification and not the actual lower window.

### much more

There are some more issues about this. Some of them I can't reproduce right now.

Is here an Cairo/Xlib pro, who can support me?

tsipinakis:
> To solve this, we have to draw every notification in a separate window.

This is a huge change since dunst was built around the assumption that everything is a single window. That being said, I can see how using multiple windows can be an advantage customization wise (it'll also allow customizing the geometry per-notification which is a very commonly requested feature).

However this will probably have to wait until the render code is cleaned up a bit. I just pushed the `refactor-drawing` branch which is my WIP attempt to decouple pango/cairo from X11.

bebehei:
Let's close this. I've found today a good topic, [how to draw transparency](https://www.cairographics.org/FAQ/#clear_a_surface).

It does the wanted effect, but it's not a simple fix, because the drawing structure is a bit cumbersome. Let's properly do this in a drawing refactoring.

TRSx80:
@bebehei 

Well, good try anyway. :smile: :+1:

jpldcarvalho:
So... I was able to fix the RGBA issue, but when we have multiple notifications, it will stack everything and becomes really ugly

bebehei:
@jpldcarvalho This may be another hint: https://github.com/dunst-project/dunst/issues/573#issuecomment-450522461

jpldcarvalho:
@bebehei I think we need to keep rendering from x to x time in the timeout loop. That way it will enable the changes in the background while the notification is enabled.
I'm not that familiar with the code, do you know where can I mess with the timeout loop?

bebehei:
What is the 'timeout loop' for you?

We've got the event loop, which is located inside `dunst.c`, which gets called periodically over and over by the interval, which is taken from `queues_get_next_datachange()`.

jpldcarvalho:
It was that what I was looking for. Thank you very much !

PR #346, opened 2017-08-01
change order of option_get_* to be displayed during --help

bebehei:
This does basically, what's discussed in #170


tsipinakis:
Thanks!

PR #348, opened 2017-08-03
print usage only on stdout

bebehei:
Print out the full usage on STDERR. Currently only the string `usage` gets printed there, but not the other stuff. This is stupid.

### DIFF
--- a/src/dunst.c
+++ b/src/dunst.c
@@ -390,7 +390,7 @@ int dunst_main(int argc, char *argv[])
 
 void usage(int exit_status)
 {
-        fputs("usage:\n", stderr);
+        puts("usage:\n");
         char *us = cmdline_create_usage();
         puts(us);
         exit(exit_status);


bebehei:
Example:
```
[bebe:~] % dunst --help >/dev/null
usage:
[bebe:~] % dunst --help 2>/dev/null
-v/-version                              - Print version
--version                                - Print version
-conf/-config (string)                   - Path to configuration file
-force_xinerama                          - Force the use of the Xinerama extension
[...]
```

tsipinakis:
Is there a reason we're printing usage information to stderr? I think stdout would be a more appropriate place for it.

bebehei:
> Is there a reason we're printing usage information to stderr? I think stdout would be a more appropriate place for it.

Well, actually no. `usage` will only get called, if `--help` is used. So it's even better to have it on stdout. I updated it.

tsipinakis:
Thanks!

PR #351, opened 2017-08-04
replace variables and ~ in script paths

### LINKED ISSUE #350: $HOME or ~ as path
Please allow them as paths for running scripts

bebehei:
wordexp replaces all variables and `~` in the string. TBH I don't quite like the thing, that it replaces all variables, but there is no other way.

Fixes #350

### DIFF
--- a/src/option_parser.c
+++ b/src/option_parser.c
@@ -109,6 +109,11 @@ char *get_value(char *section, char *key)
         return NULL;
 }
 
+char *ini_get_path(char *section, char *key, const char *def)
+{
+        return string_to_path(ini_get_string(section, key, def));
+}
+
 char *ini_get_string(char *section, char *key, const char *def)
 {
         char *value = get_value(section, key);
@@ -355,6 +360,17 @@ char *cmdline_get_string(char *key, const char *def, char *description)
                 return g_strdup(def);
 }
 
+char *cmdline_get_path(char *key, const char *def, char *description)
+{
+        cmdline_usage_append(key, "string", description);
+        char *str = cmdline_get_value(key);
+
+        if (str)
+                return string_to_path(g_strdup(str));
+        else
+                return string_to_path(g_strdup(def));
+}
+
 int cmdline_get_int(char *key, int def, char *description)
 {
         cmdline_usage_append(key, "double", description);
@@ -389,6 +405,24 @@ int cmdline_get_bool(char *key, int def, char *description)
 bool cmdline_is_set(char *key)
 {
         return cmdline_get_value(key) != NULL;
+}
+
+char *option_get_path(char *ini_section, char *ini_key, char *cmdline_key,
+                        const char *def, char *description)
+{
+        char *val = NULL;
+
+        if (cmdline_key) {
+                val = cmdline_get_path(cmdline_key, NULL, description);
+        }
+
+
+        if (val) {
+                return val;
+        } else {
+                return ini_get_path(ini_section, ini_key, def);
+        }
+
 }
 
 char *option_get_string(char *ini_section, char *ini_key, char *cmdline_key,
--- a/test/option_parser.c
+++ b/test/option_parser.c
@@ -10,6 +10,7 @@ TEST test_next_section(void)
         char *section = NULL;
         ASSERT_STR_EQ("bool", (section = next_section(section)));
         ASSERT_STR_EQ("string", (section = next_section(section)));
+        ASSERT_STR_EQ("path", (section = next_section(section)));
         ASSERT_STR_EQ("int", (section = next_section(section)));
         ASSERT_STR_EQ("double", (section = next_section(section)));
         PASS();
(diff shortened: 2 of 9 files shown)


bebehei:
Well, it might be good here, to add tests, too.

tsipinakis:
Yeah, tests are a good idea. `option_parser.c` already has tests for most of the functions so it should be trivial to extend them for the path one.

bebehei:
> Well, it might be good here, to add tests, too.

Here they are.

src/option_parser.c
```diff
@@ -109,6 +110,19 @@ char *get_value(char *section, char *key)
         return NULL;
 }
 
+char *ini_get_path(char *section, char *key, const char *def)
+{
+        char *path = ini_get_string(section, key, def);
+
+        wordexp_t replaced_path;
+        if (path && 0 == wordexp(path, &replaced_path, WRDE_NOCMD)){
+                path = g_strdup(replaced_path.we_wordv[0]);
```
bebehei: @tsipinakis I'm missing here a `free(path);`, right?

test/option_parser.c
```diff
@@ -59,6 +61,39 @@ TEST test_ini_get_string(void)
...
+        char *section = "path";
+        char *ptr;
+        char *path;
+        wordexp_t replaced_path;
+
+        ASSERT_STR_EQ("/home/test/.path", (ptr = ini_get_path(section, "expand_none", "")));
+        free(ptr);
+
+        path = ini_get_string(section, "expand_home", "");
+        wordexp(path, &replaced_path, WRDE_NOCMD);
```
tsipinakis: Since we know the contents of the expansion we should use the simplest method (read: `getenv`) to get the data we need to compile the test case. Using the exact same code the function you're testing is using is error prone and can miss quite a few bugs.

 (hint: You can use the `string_replace` function from `utils.c` to do the actual replacement)

test/option_parser.c
```diff
@@ -59,6 +61,39 @@ TEST test_ini_get_string(void)
...
+        ASSERT_STR_EQ(replaced_path.we_wordv[0], (ptr = ini_get_path(section, "expand_home", "")));
+        free(ptr);
+        free(path);
+        wordfree(&replaced_path);
+
+        path = ini_get_string(section, "expand_tilde", "");
+        wordexp(path, &replaced_path, WRDE_NOCMD);
+        ASSERT_STR_EQ(replaced_path.we_wordv[0], (ptr = ini_get_path(section, "expand_tilde", "")));
```
tsipinakis: Same with this test, we should use the simplest method to setup the test case.

src/option_parser.c
tsipinakis: Yeah you are missing a `free`, `path` from `init_get_string` is never freed.

src/option_parser.c
```diff
@@ -109,6 +110,19 @@ char *get_value(char *section, char *key)
         return NULL;
 }
 
+char *ini_get_path(char *section, char *key, const char *def)
+{
+        char *path = ini_get_string(section, key, def);
+
+        wordexp_t replaced_path;
+        if (path && 0 == wordexp(path, &replaced_path, WRDE_NOCMD)){
```
tsipinakis: Some more error handling here would be nice, especially for `WRDE_BADVAL` and `WRDE_CMDSUM`.

src/option_parser.c
tsipinakis: You're also only getting the first word here, we_wordv is a string array (`char **`), not a `char *`. It's a common mistake so please also add a regression test for this.

bebehei:
*Disclaimer: I'm currently not at my laptop for the next few days and can't test it. I read the wordexp manpage over and over again until I fully understood the Problems.*

I have to conclude, that this is the shittiest approach possible.

Here is why:

> You're also only getting the first word here, we_wordv is a string array

Yes this is correct. So if there are multiple words, we have to strcat all `we_wordv` elements together with `$IFS`.

> [Field splitting is done using the environment variable $IFS. If it is not set, the field separators are space, tab and newline.](https://linux.die.net/man/3/wordexp)

Well, if someone sets `IFS=/', the returned path is broken.

Even worse: IFS is unset, the path contains spaces. So how should we deterministically chain the words again together?

Wordexp also does more stuff, I personally dislike. It also replaces other variables. IMO in our case only `$HOME` (maybe additionally `$USER`) should get expanded and nothing more.

> (hint: You can use the string_replace function from utils.c to do the actual replacement)

What about using this in the actual `ini_get_string`? I think about replacing the first two characters if they match `~/` with `getenv('HOME')`.

bebehei:
>> Field splitting is done using the environment variable $IFS. If it is not set, the field separators are space, tab and newline.
>
> Well, if someone sets `IFS=/', the returned path is broken.
>
> Even worse: IFS is unset, the path contains spaces. So how should we deterministically chain the words again together?

Well, it turns out to be true in all cases. `wordexp` makes classical word. This is uesless.

> What about using this in the actual `ini_get_string`? I think about replacing the first two characters if they match `~/` with `getenv('HOME')`.

So, I went for that. Here are the commits. Sorry for the spam on travis. In my previous commits, I failed using `strcat`. :facepalm:

tsipinakis:
> So, I went for that. Here are the commits. Sorry for the spam on travis. In my previous commits, I failed using strcat.

No worries, you're not a C programmer until you've wasted a few hours debugging weird memory issues :p 

It might be beneficial to also apply this to the `dmenu` and `browser` settings, they are also paths after all.

bebehei:
> It might be beneficial to also apply this to the dmenu and browser settings, they are also paths after all.

Thanks for the hint.

bebehei:
> It might be beneficial to also apply this to the dmenu and browser settings, they are also paths after all.

I moved the logics now to a function in `utils.h`, added cmdline support and dmenu and the browser use these settings now, too.

The tests from `test_ini_get_path` about correct expansion have moved now to a separate test in `utils.h`, while in `test_(ini|cmdline|option)_get_path` now only test, if expansion works. and the handling of cmdline precedence.

tsipinakis:
LGTM 👍 

Thank you for implementing this.

PR #352, opened 2017-08-06
handle removing of elements in GList correctly

bebehei:
When closing a notification in check_timeouts, `iter` will get removed of the list and therefore `iter->next` is not available at the end of the loop anymore. This requires calling check_timeouts again.

Introducing the `nextiter` pointer, which will save `iter->next` in advance, solves this problem and does not require traversing the already traversed part of the list again.

### DIFF
--- a/src/dunst.c
+++ b/src/dunst.c
@@ -57,8 +57,8 @@ void check_timeouts(void)
         if (displayed->length == 0)
                 return;
 
-        for (GList * iter = g_queue_peek_head_link(displayed); iter;
-             iter = iter->next) {
+        GList *iter = g_queue_peek_head_link(displayed);
+        while (iter) {
                 notification *n = iter->data;
 
                 /* don't timeout when user is idle */
@@ -72,12 +72,12 @@ void check_timeouts(void)
                         continue;
                 }
 
+                /* shift iter before we possibly delete the current notification */
+                iter = iter->next;
+
                 /* remove old message */
                 if (difftime(time(NULL), n->start) > n->timeout) {
-                        /* close_notification may conflict with iter, so restart */
                         notification_close(n, 1);
-                        check_timeouts();
-                        return;
                 }
         }
 }


src/dunst.c
```diff
@@ -57,8 +57,17 @@ void check_timeouts(void)
...
                 return;
 
-        for (GList * iter = g_queue_peek_head_link(displayed); iter;
-             iter = iter->next) {
+        /*
+         * If the iter element gets removed from the queue, iter->next won't
+         * point to the next element anymore. (It's NULL.)
+         * nextiter is therefore used as a buffer to have access to iter->next
+         * also after removing iter from the list
+         * */
```
tsipinakis: There shouldn't be an asterix before a comment close. e.g.
```c
/*
 * Comment
 */
```

tsipinakis:
Nice, micro-optimizations are always welcome.

Though this seems to me that it overcomplicates the loop quite a lot without much gain, why not use a simple while loop and re-assign `iter` before removing the notification, am I missing something?
```c
GList *iter = g_queue_peek_head_link(displayed);
while(iter) {
notification *n = iter->data;
// We have the notification which is technically all we need, but we should hold off re-assigning iter to avoid confusion later on if we decide to use it for some reason
// do stuff

// Re-assign iter since we are not using it anymore and it can potentially be removed in the next command
iter = iter->next

//remove notification
}
```

bebehei:
Yes, that's true. It just was the first time, I actually "abused" a for-loop to use multiple variables. I also learned about the comma operator (WTF? such weirdly named thing exists!?)

But yeah, it's much easier using a while loop.

src/dunst.c
```diff
@@ -72,12 +72,12 @@ void check_timeouts(void)
                         continue;
                 }
 
+                /* shift iter before we possibly delete the current notification */
+                iter = iter->next ? iter->next : NULL;
```
tsipinakis: Small nitpick, but why use a conditional here? If `iter->next` is `NULL`, `iter` will be assigned to `NULL` anyway.

bebehei: Shit this is a typo. The first `iter->next` should be iter.

tsipinakis: Even if it's meant to be `iter` it's the same thing, if `iter` is NULL it wouldn't have got past the while loop check so it wouldn't reach this line.

bebehei: If `iter` is null, `iter->next` is not available.

tsipinakis: My point is that `iter` *cannot* be null at this point because
a) the while loop checks if `iter` is NULL and since this is the only line where we modify `iter` there is no way for it to be null without it being a bug and 
b) We already access `iter->data` at the top of the loop, if somehow `iter` is null dunst would have crashed long before reaching this point.

In my opinion silently handling an impossible condition is a bad practice since since it can obscure bugs that might be introduced.

bebehei: 🙈 yes. full ACK.

tsipinakis:
Thank you!

dikiaap:
I have a problem about critical notifications.

I just checked every commit before this commit, all is well.
And now I use this commit untill new commit, I have a problem in critical notifications that can not be closed.

To solve my problem, I should change the `timeout = 0` to `timeout = 10` or some more than 0
[My config](https://github.com/dikiaap/dotfiles/blob/master/.dunst/dunstrc#L53)

tsipinakis:
Weird, thanks for spotting this. Can you file an issue about it since it's usually easier to track than a comment in a PR?

tsipinakis:
Nvm I figured it out. Apparently I missed the fact that the `continue` calls create an infinite loop since iter isn't being incremented. Rookies mistake, my bad.

dikiaap:
Whoops. Okay.

bebehei:
Sorry for this spam

PR #353, opened 2017-08-06
parse raw icons from the (deprecated) image_data attribute

sfs:
The desktop notification spec version 1.1 used "image_data" instead of "image-data" as the key for raw icons. This is still used in some applications, e.g., in the [Clementine music player](https://github.com/clementine-player/Clementine/blob/42aafd247a3a96f7a6c07d31b3a0881691a8daf2/src/widgets/osd_x11.cpp#L111).

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -196,6 +196,8 @@ static void on_notify(GDBusConnection * connection,
 
                                         dict_value = g_variant_lookup_value(content, "image-data", G_VARIANT_TYPE("(iiibiiay)"));
                                         if (!dict_value)
+                                                dict_value = g_variant_lookup_value(content, "image_data", G_VARIANT_TYPE("(iiibiiay)"));
+                                        if (!dict_value)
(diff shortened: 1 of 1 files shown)


bebehei:
What about fixing it there, too?

sfs:
I did submit a [PR](https://github.com/clementine-player/Clementine/pull/5808) to Clementine, but I'm using a non-standard branch. Not sure how useful this will be.

tsipinakis:
> I did submit a PR to Clementine, but I'm using a non-standard branch. Not sure how useful this will be.

Looks like your PR got rejected since the maintainer doesn't want to merge fixes into that branch. I'd recommend at least filing an issue about it to server as a reminder.

---

The notification spec doesn't say anything about where `image_data` should be in the priority list, but assuming it's equivalent to `image-data` I'm merging this as is, but might reconsider if it starts causing problems.

Thank you!

PR #354, opened 2017-08-08
Add libgen.h include and guard against basename(3) modifying its arg

enodata:
Implicitly declared basename() makes dunst dump core on FreeBSD 11.1 (in src/notification.c:209)

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -13,6 +13,7 @@
 #include <sys/wait.h>
 #include <time.h>
 #include <unistd.h>
+#include <libgen.h>
 
 #include "dbus.h"
 #include "dunst.h"
@@ -313,10 +314,12 @@ int notification_init(notification * n, int id)
                 n->markup);
 
         if (n->icon) {
-                n->msg = notification_replace_format("%I", basename(n->icon),
+                char *tmp = g_strdup(n->icon);
+                n->msg = notification_replace_format("%I", basename(tmp),
                         n->msg, MARKUP_NO);
                 n->msg = notification_replace_format("%i", n->icon,
                         n->msg, MARKUP_NO);
+                g_free(tmp);
         }
 
         if (n->progress) {


tsipinakis:
From `basename(3)`:
> There are two different versions of basename() - the POSIX version described above, and the GNU version, which one gets after
> ```c
>    #define _GNU_SOURCE         /* See feature_test_macros(7) */
>    #include <string.h>
> ```
> The GNU version never modifies its argument, and returns the empty string when path has a trailing slash, and in particular also when it is "/".  There is no GNU version of dirname().
> With glibc, one gets the POSIX version of basename() when <libgen.h> is included, and the GNU version otherwise.

It looks like we're using the GNU version for one reason or another. Any ideas why that doesn't work on FreeBSD?

enodata:
FreeBSD doesn't use glibc so the only available basename(3) is POSIX version.

tsipinakis:
Sorry for the delay, I've been busy these past few days.
I've opened #359 to check for similar issues and ultimately phase out `_GNU_SOURCE`. As for this PR, the current usage of `basename` doesn't account for the fact that the argument may be modified in the POSIX version. Can you also update it so that it passes a duplicate and then frees it?

enodata:
Hi Nikos,

Thanks for pointing this out. basename(3) in current production FreeBSD versions (11.x and 10.x) doesn't modify its argument (and I believe OpenBSD's version doesn't either), but basename(3) in FreeBSD 12-CURRENT (dev version) does. I updated PR to make a path copy to safeguard against argument modifications.

tsipinakis:
Thank you!

PS: Can you confirm that dunst works fine in BSD now? I've been meaning to close #260 for a while but I'm not sure if the issue mentioned is fixed.

enodata:
I can confirm that dunst now works on FreeBSD. I'll test on OpenBSD later today and post results in #260.

PR #364, opened 2017-08-24
Fix handling of non V1.5 XRandr.

SteveJones:
Due to the way Xlib handles errors XRRGetMonitors will cause dunst to exit if
the server doesn't support version 1.5 of Randr. The check for null is
effectively dead code. Change the code to fetch the Randr version and fallback
if the version is less than 1.5 when getting monitors.

### DIFF
--- a/src/x11/screen.c
+++ b/src/x11/screen.c
@@ -23,6 +23,11 @@ int screens_len;
 
 bool dunst_follow_errored = false;
 
+int randr_event_base = 0;
+
+static int randr_major_version = 0;
+static int randr_minor_version = 0;
+
 void randr_init();
 void randr_update();
 void xinerama_update();
@@ -80,8 +85,6 @@ void alloc_screen_ar(int n)
         screens_len = n;
 }
 
-int randr_event_base = 0;
-
 void randr_init()
 {
         int randr_error_base = 0;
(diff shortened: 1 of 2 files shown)


src/x11/screen.c
```diff
@@ -82,27 +82,33 @@ void alloc_screen_ar(int n)
 
 int randr_event_base = 0;
 
+static int randr_major_version = 1;
```
tsipinakis: Why default these to 1 and 5? Just leave them uninitialized (since they're globals they're guaranteed to be 0).

src/x11/screen.c
```diff
@@ -82,27 +82,33 @@ void alloc_screen_ar(int n)
...
+        XRRQueryVersion(xctx.dpy, &randr_major_version, &randr_minor_version);
         XRRSelectInput(xctx.dpy, RootWindow(xctx.dpy, DefaultScreen(xctx.dpy)), RRScreenChangeNotifyMask);
 }
 
 void randr_update()
 {
-        int n;
-        XRRMonitorInfo *m = XRRGetMonitors(xctx.dpy, RootWindow(xctx.dpy, DefaultScreen(xctx.dpy)), true, &n);
-
-        if (m == NULL || n == -1) {
```
tsipinakis: In [`xrandr.c`](https://cgit.freedesktop.org/xorg/app/xrandr/tree/xrandr.c#n2641) I see that there's an error check for `n == -1` so this check should probably stay and have a separate check for version.

src/x11/screen.c
```diff
@@ -82,27 +82,33 @@ void alloc_screen_ar(int n)
...
+        if (randr_minor_version < 5) {
```
tsipinakis: You're only checking for `minor < 5` here which is going to break at the next major version. A more complete check would be `randr_major_version > 1 || (randr_major_version == 1 && randr_minor_version < 5)`

Additionally, there should probably be a more descriptive error message mentioning the outdated version.

tsipinakis:
@SteveJones Ping, are you still active?

SteveJones:
Hi yeah, sorry this slipped my mind. I'll try and get this fixed up in the next day or so.

src/x11/screen.c
SteveJones: This was me reading the protocol docs which say that the client has to send the version it expects, but the xrandr code sends this itself, it doesn't use the initial value of the pointers you pass in. I'll fix this.

src/x11/screen.c
SteveJones: Hmm, looking at https://cgit.freedesktop.org/xorg/lib/libXrandr/tree/src/XrrMonitor.c it's checking a weird edge case where the server returns 0 monitors, but I guess that's possible. I'll put the check back.

src/x11/screen.c
SteveJones: Good point.

src/x11/screen.c
tsipinakis: > This was me reading the protocol docs which say that the client has to send the version it expects

Where does it say that? It's not mentioned anywhere in the xrandr man page?

SteveJones: It's in the protocol specification, but the library API doesn't match that.

tsipinakis:
Looks like you misspelled the major version variable.

Have you tested this in a system with an outdated version (and probably on one with no xrandr installed at all)? Just to confirm that it works as intended.

SteveJones:
That'll teach me to add in a cosmetic change after testing it builds (it won't).

I've tested it on outdated, XRandR, it's why I wrote it.

src/x11/screen.c
```diff
@@ -98,7 +98,8 @@ void randr_init()
 
 void randr_update()
 {
-        if (randr_minor_version < 5) {
+	if (randr_major_version > 1
```
tsipinakis: Noticed this just as I was about to click merge:

We're checking for the error condition so it should be `randr_major_version < 1`

SteveJones:
Right, fixed that and squashed. Tests just passed. It builds and seems happy. Should be good to go.

tsipinakis:
> It builds and seems happy. Should be good to go.

Apparently not :p Just realised you used tabs, we're using 8 spaces for indentation. should be a quick fix.

tsipinakis:
> If we're so nitpicky now

That's not a nitpick in my opinion, it's adhering to the projects code style else it just becomes a mess.

The commit message format is generally up to the contributor and I don't interfere with that, just give hints about the correct standards (and I didn't actually notice the period this time).

tsipinakis:
Ping @SteveJones

bebehei:
I misused my privileges the first time now pushed the stylefixes to this PR.

PR #365, opened 2017-08-28
Prevent replacement of format strings in message

### LINKED ISSUE #322: Format inside notification contents are being expanded as in format
Using `%I` inside the message body expands `%I` which is not supposed to happen:

> 4:23:42 PM - nikos: Also, you are also using a second bug here which is that %I etc values get expanded in the body of the message. This hasn't been fixed yet but might get to it soon.

bebehei:
To replace all occuring format strings inside the msg, replace_all had
been used previously. This leads to buggy behavior, if a format string
occurs in the replaced text, as the format string will get replaced
again under certain conditions.

Introducing a pointer, which skips the already replaced parts, will
prevent doubly replacing format strings from content.

Fixes #322

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -387,6 +387,13 @@ int notification_init(notification *n, int id)
                                                 n->progress ? pg : "",
                                                 MARKUP_NO);
                                 break;
+                        case '%':
+                                n->msg = notification_replace_single_field(
+                                                n->msg,
+                                                &substr,
+                                                "%",
+                                                MARKUP_NO);
+                                break;
                         case '\0':
                                 fprintf(stderr, "WARNING: format_string has trailing %% character."
                                                 "To escape it use %%%%.");
--- a/src/notification.c
+++ b/src/notification.c
@@ -200,17 +200,33 @@ void notification_free(notification *n)
 }
 
 /*
- * Replace all occurrences of "needle" with a quoted "replacement",
- * according to the markup settings.
+ * Replace the two chars where **needle points
+ * with a quoted "replacement", according to the markup settings.
+ *
+ * The needle is a double pointer and gets updated upon return
+ * to point to the first char, which occurs after replacement.
+ *
  */
-char *notification_replace_format(const char *needle, const char *replacement,
-                                  char *haystack, enum markup_mode markup_mode) {
-        char *tmp;
+char *notification_replace_single_field(char *haystack, char **needle,
+                const char *replacement, enum markup_mode markup_mode) {
+
+        assert(*needle[0] == '%');
+        // needle has to point into haystack (but not on the last char)
+        assert(*needle >= haystack);
+        assert(*needle - haystack < strlen(haystack) - 1);
+
         char *ret;
 
-        tmp = markup_transform(g_strdup(replacement), markup_mode);
-        ret = string_replace_all(needle, tmp, haystack);
-        g_free(tmp);
+        int pos = *needle - haystack;
+
+        char *input = markup_transform(g_strdup(replacement), markup_mode);
+        ret = string_replace_at(haystack, pos, 2, input);
+
+        // point the needle to the next char
+        // which was originally in haystack
+        *needle = ret + (*needle - haystack) + strlen(input);
+
+        g_free(input);
 
         return ret;
 }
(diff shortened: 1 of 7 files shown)


src/notification.c
```diff
@@ -199,17 +199,29 @@ void notification_free(notification * n)
...
+ * The needle is a double pointer and gets updated upon return
+ * to point to the first char, which occurs after replacement.
+ *
  */
-char *notification_replace_format(const char *needle, const char *replacement,
-                                  char *haystack, enum markup_mode markup_mode) {
-        char* tmp;
+char *notification_replace_single_field(char *haystack, char **needle,
+                const char *replacement, enum markup_mode markup_mode) {
+
```
tsipinakis: Since this is assuming that `**needle` is the beginning of a format flag, I think an assertion is in order to catch any bugs.

src/notification.c
```diff
@@ -306,31 +318,80 @@ int notification_init(notification * n, int id)
...
+                                break;
+                        case '%':
+                                n->msg = notification_replace_single_field(
+                                                n->msg,
+                                                &substr,
+                                                "%",
+                                                MARKUP_NO);
+                                break;
+                        default:
+                                break;
```
tsipinakis: There should probably be a warning here, let's not silently ignore unknown items.

test/notification.c
```diff
@@ -70,21 +70,25 @@ TEST test_notification_is_duplicate(void *notifications)
...
-        ASSERT_STR_EQ("Markup and &amp; &lt;i&gt;is&lt;/i&gt; escaped", (str = notification_replace_format("%a", "and & <i>is</i>", str, MARKUP_NO)));
+        substr = strchr(str, '%');
+        ASSERT_STR_EQ("Markup and &amp; &lt;i&gt;is&lt;/i&gt; escaped", (str = notification_replace_single_field(str, &substr, "and & <i>is</i>", MARKUP_NO)));
+        ASSERT_EQ(38, substr - str);
 
         strcpy(str, "Markup %a");
-        ASSERT_STR_EQ("Markup is removed and &amp; escaped", (str = notification_replace_format("%a", "<i>is removed</i> and & escaped", str, MARKUP_STRIP)));
+        substr = strchr(str, '%');
+        ASSERT_STR_EQ("Markup is removed and &amp; escaped", (str = notification_replace_single_field(str, &substr, "<i>is removed</i> and & escaped", MARKUP_STRIP)));
+        ASSERT_EQ(strlen("Markup is removed and &amp; escaped"), substr - str);
```
tsipinakis: Why use `strlen` here but hardcode all the others?

src/notification.c
```diff
@@ -199,17 +199,29 @@ void notification_free(notification * n)
...
+ * with a quoted "replacement", according to the markup settings.
+ *
+ * The needle is a double pointer and gets updated upon return
+ * to point to the first char, which occurs after replacement.
+ *
  */
-char *notification_replace_format(const char *needle, const char *replacement,
-                                  char *haystack, enum markup_mode markup_mode) {
-        char* tmp;
+char *notification_replace_single_field(char *haystack, char **needle,
```
tsipinakis: If we going to use a double pointer for the needle, I think it'd be better to also use it for the haystack and avoid the awkward assignments in `notification_init`.

bebehei:
Found a bug: format string `<some stuff>%A` creates endless loop

src/notification.c
bebehei: This is fixed now the line below.

src/notification.c
bebehei: Is this more "C-ish" than the current signature? I changed it, now for evaluation.

src/notification.c
bebehei: Also I added another assertion, that `needle` points into `haystack`.

tsipinakis:
> Found a bug

I'd love to have a regression test for this but that means it'd also print the warning with it. I've opened #368 to look into somehow controlling how logging works.

src/notification.c
```diff
@@ -199,19 +199,33 @@ void notification_free(notification * n)
...
 
-        tmp = markup_transform(g_strdup(replacement), markup_mode);
-        ret = string_replace_all(needle, tmp, haystack);
-        g_free(tmp);
+        assert(*needle[0] == '%');
+        // needle has to point into haystack
+        assert(*needle >= *haystack);
+        assert(*needle - *haystack <= strlen(*haystack));
+
+        char* input;
```
tsipinakis: While currently it's already in a similar way in `master` I think it's worth fixing these nitpicks since we're refactoring this anyway:

1) There is no reason to have this declaration here and not a few lines bellow where `input` is first used
2) While it's pretty inconsistent across the project, we usually put the asterisk next to the variable name and not the type e.g. `char *input`.
The rationale for this is simply that it gets confusing when using commas, for example in `char* a,b` `a` is a pointer while `b` is not.

src/notification.c
```diff
@@ -199,19 +199,33 @@ void notification_free(notification * n)
...
+void notification_replace_single_field(char **haystack, char **needle,
+                const char *replacement, enum markup_mode markup_mode) {
 
-        tmp = markup_transform(g_strdup(replacement), markup_mode);
-        ret = string_replace_all(needle, tmp, haystack);
-        g_free(tmp);
+        assert(*needle[0] == '%');
+        // needle has to point into haystack
+        assert(*needle >= *haystack);
+        assert(*needle - *haystack <= strlen(*haystack));
```
tsipinakis: This assertion should probably test for less than not less-than-or-equal: We don't want a trailing % at the end of the string, if anything it might trip over `string_replace_at` into undefined behaviour.

src/notification.c
```diff
@@ -306,31 +320,85 @@ int notification_init(notification * n, int id)
...
+                        case '%':
+                                notification_replace_single_field(
+                                        &n->msg,
+                                        &substr,
+                                        "%",
+                                        MARKUP_NO);
+                                break;
+                        default:
+                                fprintf(stderr, "WARNING: format_string %%%c"
+                                                " is unknown\n", substr[1]);
```
tsipinakis: There should be a different message if `substr[1]` is NULL, aka we're at the end of the string.

e.g. `WARNING: Trailing % mark in format`

src/notification.c
bebehei: > While it's pretty inconsistent across the project, we usually put the asterisk next to the variable name and not the type e.g. char *input.
The rationale for this is simply that it gets confusing when using commas, for example in char* a,b a is a pointer while b is not.

Thanks for answering this question. I knew this was the C-specific equivalent of "tabs vs spaces", so I kept avoiding it, because I thought I could not understand all arguments. But thanks it's very clear now.

> There is no reason to have this declaration here and not a few lines bellow where input is first used

So, while we're at these "religious" questions:

```
char *string = function_returning_string();
```

```
[begin of the block]
char *string;
[maybe some code]
string = function_returning_string();
```

I've seen both styles here in the project, but I haven't found a scheme. My observations had been, that a C-programmer rather uses the 2nd style.

What's better or when use what?

bebehei:
> I'd love to have a regression test

I'd love that, too. But I consider this part of the function as "not testable yet". The function is too big and needs to be refactored to have a senseful unittest.

src/notification.c
bebehei: Treated.

src/notification.c
bebehei: Solved.

tsipinakis:
Can you squash 1aaf109 into ac5542a and resolve the conflicts?

bebehei:
Rebased and LGTM.

tsipinakis:
LGTM 👍

PR #366, opened 2017-08-30
Fix memory leak in get_pixbuf_from_path

bebehei:
If the svg-file is not found, the string containing the svg ending is never freed again.


tsipinakis:
Nice catch, thanks!

PR #369, opened 2017-09-06
Refactor star scheme of pointers

bebehei:
As @tsipinakis mentioned in #365 ([here](https://github.com/dunst-project/dunst/pull/365#discussion_r137354820)), this unifies all pointer declarations.

To achieve a consistent scheme, all stars of the pointers should be on the side of the variable and not on the side of the type.

wrong: `char* a`
wrong: `char * a`
good:  `char *a`

This commit is generated by the following sed command with manual fixes of all false positives.

```
find src \( -name '*.c' -or -name '*.h' \) -print0 | xargs -0 -n1 sed -i 's/\([a-zA-Z]\+[-_a-zA-Z0-9]*\)\s*\*\s\+\([a-zA-Z]\+[-_a-zA-Z0-9]*\)/\1 *\2/g'
```

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -59,30 +59,30 @@ static const char *introspection_xml =
     "   </interface>"
     "</node>";
 
-static void on_get_capabilities(GDBusConnection * connection,
-                              const gchar * sender,
-                              const GVariant * parameters,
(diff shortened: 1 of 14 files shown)


tsipinakis:
Awesome work as usual, thank you!

PR #371, opened 2017-09-11
Improve usage of config.h

bebehei:
I had to look over it again and the complete `default_rules` formatting looked very ugly.


tsipinakis:
I'm not sure about this, sure it's a good stylistic change but I really don't like the the current system for hardcoding rules and would love to change it.

I'll leave this on hold till next week or so when I return to see what can be done long-term.

bebehei:
> I really don't like the the current system for hardcoding rules and would love to change it.

Yeah, I'm thinking the same about this. AFAICS, the default rules get overwritten by rules in the config file. We should also prefix the default rules with `default_` (or sth. else).

bebehei:
@tsipinakis I've thought about this again.

Albeit, the last commit is about removing `STATIC_CONFIG`, I'm not definite about `STATIC_CONFIG`. I guess we all agree, that we should remove it at least in next major.

So my thoughts about `config.def.h`:

On the developer sight, using `config.def.h` and `config.h` is absolutely clumsy. When switching branches, you always have to remove `config.h` first, before you'll be able to compile.

So, as a developer, I want to remove it.

But that's just the side of a developer. I also want to lay out, why there is a better technique for the original use case of `config.h`:

Why do we want to have a separate `config.def.h`? Previously, when there had been no config file support you had to hardcode settings and recompile it. So you wanted to have a file, where you save your settings independent of the code. Additionally, you could use `config.h` to ignore it in git, so if you only change your settings, the gitrepo is not polluted.

But do you really want that behavior? First: Any addition in `config.def.h` breaks your config and you have to manually port the changes from `config.def.h` to `config.h` (otherwise dunst doesn't compile at all). Second: If you `git pull` and then make a manual diff of `config.h` and `config.def.h`. How do you know, that the change is done by yourself? You have to trace back the git history.

There are two techniques, which work much better for this usecase:

- Remove config.h and use only `config.def.h`: If you change the defaults manually, git will know, what you have changed and what upstream has changed. (Even better: automatic merge control)
- Force-Add an empty `config.h`, which includes `config.def.h` and remove the lines in the makefile about `config.h` handling. Pro: the stuff in `config.h` is still not tracked, you don't have to merge anything and only add there the lines, what you really want to change. Con: additional file. tracked file, which is also in gitignore (IMO a superbad hack in common development).

---

I hope I have explained well, that both approaches are suited better for all stakeholders here.

tsipinakis:
I've been thinking about how we should handle this, here's my train of thought:
* The config basically breaks on every release as new settings are added or removed so if anyones using it they should be ready to handled breakages (Granted this is a very big change - but i'd look past that). So I'm for including the stylistic changes.

* While I haven't heard or seen anyone use `STATIC_CONFIG` I'd be hesitant to remove it outright without deprecating first, so I'd recommend to add the following:
  - Noticeable comment in `config.mk` that it's deprecated
  - Runtime warning every time dunst starts

I've also filed #414.

Fun fact: While digging through the history I discovered that `config.h` was removed(5fa67d5f89f5719d061094ac2aa5f186adf995e3) and reintroduced(311e2ff4379ee317d4c2abbbf862dccbb85f263c) once before without any further comment as to why.

bebehei:
> Fun fact: While digging through the history I discovered that config.h was removed(5fa67d5) and reintroduced(311e2ff) once before without any further comment as to why.

Yes, I've seen that, too.

Long story short: Shall I change the `STATIC_CONFIG` commit to not remove it but deprecate it?

knopwob:
I can shed some light on why it was reintroduced (and shame on me for not explaning it in the commit message).
I got a few complains for removing it, since some people prefered the dwm/suckless style of having a config.h instead of config files. That's why introduced STATIC_CONFIG to make those people happy.

But IMHO since the userbase has grown since than. The users who are prefering a static configuration are such a minority that the maintainability is more important. So I have no objections against removing it.

tsipinakis:
Thanks for providing some insight.

> But IMHO since the userbase has grown since than the users who are prefering a static configuration are such a minority that the maintainability is more important. So I have no objections against removing it.

This made me second guess this decision a bit, perhaps it's a good idea to add a link back to the issue tracker prodding any users to report back with their use-cases if they're using the feature.

---

As for this PR,

The compile time warning doesn't seem to be working correctly I can't seem to pinpoint exactly what it's doing, I've had it both fire when static config was disabled and not fire when it was enabled.

Very slight nitpick/complaint: Removing `make distclean` broke my muscle memory, oh well -.-

bebehei:
> Very slight nitpick/complaint: Removing make distclean broke my muscle memory, oh well -.-

Mine, too. But you don't need distclean, if it doesn't do anything different than clean. And second we remove the actual difference of `distclean` vs `clean`. So there is no point of keeping `distclean`.

> The compile time warning doesn't seem to be working correctly I can't seem to pinpoint exactly what it's doing, I've had it both fire when static config was disabled and not fire when it was enabled.

🤦‍♂️ Sorry, this was absolutely my fault. This is now fixed. I wonder, how I tested it yesterday!?

> This made me second guess this decision a bit, perhaps it's a good idea to add a link back to the issue tracker prodding any users to report back with their use-cases if they're using the feature.

Well, we don't remove the `config.h` behavior. We pointed out, that it's actually shit using dunst with config.h while we break this with `config.def.h`. It smarter tracking it in git and having a three way diff.

The real dropping of STATIC_CONFIG is done in #412

---

>  I got a few complains for removing it, since some people prefered the dwm/suckless style of having a config.h instead of config files. That's why introduced STATIC_CONFIG to make those people happy.

Albeit, everyone can think on one's own, whether it's a good pattern hardcoding configuration on compilation or not, I had previously much respect about the suckless project. When I read the [philosophy page of them](https://suckless.org/philosophy), I had the immersion, that it spoke out of my mind.

So, I must have missed something! There must be something good about `config.h` and `config.def.h`! Let's look, how they handle the `config.h` stuff and disprove my position!

It turns out, [they even care less](https://git.suckless.org/dmenu/tree/Makefile#n21). What is the point of having a separate `config.h`, when you overwrite it **all the time**? WTF!? This is such an antipattern! (Even dunst has this fixed in #179)

tsipinakis:
Not much else to say here other than I agree with everything said.

> It turns out, they even care less. What is the point of having a separate config.h, when you overwrite it all the time? WTF!? This is such an antipattern! (Even dunst has this fixed in #179)

That's very surprising - I'd love to hear from someone who actually uses this system about why and how that works.

Narrat:
Somehow I'm getting an error compiling master (3691b7e)

```sh
src/settings.c:14:10: fatal error: config.h: No such file or directory
 #include "config.h"
          ^~~~~~~~~~
compilation terminated.
```
Trying to take a look at it

PR #372, opened 2017-09-12
Sort linker input file list

bmwiedemann:
so that dunst binaries build in a reproducible way
in spite of indeterministic filesystem readdir order

See https://reproducible-builds.org/ for why this is good.

Note: the same change could be applied to the test/test part
but that does not end up in the (openSUSE) package so does not matter as much.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ include config.mk
 CFLAGS += -I.
 LDFLAGS += -L.
 
-SRC := $(shell find src/ -name '*.c')
+SRC := $(sort $(shell find src/ -name '*.c'))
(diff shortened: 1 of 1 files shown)


bebehei:
@bmwiedemann This should be everything to make dunst fully reproducible? Really? I can't believe this!

bmwiedemann:
Actually it is (at least for openSUSE where we have a normalized build environment with constant user, build-path, timezone). And maybe because we have helpful patches in other places like gcc and rpm that override macros like __DATE__

bebehei:
Wow!

PR #374, opened 2017-09-17
Add option to control how to ellipsize truncated lines

fhost:
Add the option `ellipsize` to be able to select how the truncated lines should be ellipsized, accepting `start`, `middle` and `end`.

`middle` is the default value to not change the current behaviour.
`start` is maybe useless as it removes the "summary", but who knows.
`end` is useful to not ellipsize the (long) summary (e.g. IRC notifications where the summary (and/or the start of the body) contains "IRC", the user, and the server/channel).

### DIFF
--- a/docs/dunst.pod
+++ b/docs/dunst.pod
@@ -351,6 +351,11 @@ If it's set to false, long lines will be truncated an ellipsised.
 If it's set to true, long lines will be broken into multiple lines expanding
 the notification window height as necessary for them to fit.
 
+=item B<ellipsize> (values: [start/middle/end], default: middle)
+
+If word_wrap is set to false, specifies where truncated lines should be
+ellipsized.
+
 =item B<ignore_newline> (values: [true/false], default: false)
 
 If set to true, replace newline characters in notifications with whitespace.
--- a/src/settings.c
+++ b/src/settings.c
@@ -169,6 +169,29 @@ void load_settings(char *cmdline_config_path)
                 "Truncating long lines or do word wrap"
         );
 
+        {
+                char *c = option_get_string(
+                        "global",
+                        "ellipsize", "-ellipsize", "middle",
+                        "Ellipsize truncated lines on the start/middle/end"
+                );
+
+                if (strlen(c) > 0) {
+                        if (strcmp(c, "start") == 0)
+                                settings.ellipsize = start;
+                        else if (strcmp(c, "middle") == 0)
+                                settings.ellipsize = middle;
+                        else if (strcmp(c, "end") == 0)
+                                settings.ellipsize = end;
+                        else {
+                                fprintf(stderr,
+                                        "Warning: unknown value for ellipsize\n");
+                                settings.ellipsize = middle;
+                        }
+                        g_free(c);
+                }
+        }
+
         settings.ignore_newline = option_get_bool(
                 "global",
                 "ignore_newline", "-ignore_newline", ignore_newline,
--- a/src/x11/x.c
+++ b/src/x11/x.c
@@ -6,6 +6,7 @@
 #include <X11/Xatom.h>
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
+#include <assert.h>
 #include <cairo.h>
 #include <cairo-xlib.h>
 #include <gdk/gdk.h>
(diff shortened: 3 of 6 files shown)


docs/dunst.pod
```diff
@@ -351,6 +351,10 @@ If it's set to false, long lines will be truncated an ellipsised.
 If it's set to true, long lines will be broken into multiple lines expanding
 the notification window height as necessary for them to fit.
 
+=item B<ellipsize> (values: [start/middle/end], default: middle)
+
+Specifies where truncated lines should be ellipsized.
```
bebehei: Could you please add also the information, that it's only used, if word_wrap is set to no? Also it would be great, if you add it to the dunstrc file, too.

src/x11/x.c
```diff
@@ -404,7 +404,20 @@ static colored_layout *r_init_shared(cairo_t *c, notification *n)
...
 
         if (!settings.word_wrap) {
-                pango_layout_set_ellipsize(cl->l, PANGO_ELLIPSIZE_MIDDLE);
+                PangoEllipsizeMode ellipsize;
+                switch (settings.ellipsize) {
+                        case start:
+                                ellipsize = PANGO_ELLIPSIZE_START;
+                                break;
+                        case middle:
+                        default:
```
tsipinakis: Default should be an error condition, if ellipsize is anything other than `start`, `middle` or `end` it's most definitely a bug.

src/settings.c
```diff
@@ -169,6 +169,27 @@ void load_settings(char *cmdline_config_path)
                 "Truncating long lines or do word wrap"
         );
 
+        {
+                char *c = option_get_string(
+                        "global",
+                        "ellipsize", "-ellipsize", "",
```
tsipinakis: Since we know middle is the default it's a good idea to set this here else the fact that middle is used by default is only because it's the first entry in the enum.

src/settings.c
```diff
@@ -169,6 +169,27 @@ void load_settings(char *cmdline_config_path)
...
+                if (strlen(c) > 0) {
+                        if (strcmp(c, "start") == 0)
+                                settings.ellipsize = start;
+                        else if (strcmp(c, "middle") == 0)
+                                settings.ellipsize = middle;
+                        else if (strcmp(c, "end") == 0)
+                                settings.ellipsize = end;
+                        else
+                                fprintf(stderr,
+                                        "Warning: unknown truncating\n");
```
tsipinakis: `unknown truncating` Doesn't say much, a better message would be `unknown value for ellipsize`

tsipinakis:
Oh, I also missed that you forgot to add it to the example dunstrc file.

src/x11/x.c
fhost: Ok. In fact I copy/paste how the alignment is handle (with `left`, `center` and `right` values). So I don't know how you want me to handle this case.
Should I throw an error saying something went wrong and exiting?

src/settings.c
fhost: Same here, I based my code on the alignment, that why I did not put `"middle"` as the default value.
I'll put back the enum in a logical order too (as `middle` doesn't have to be the first one anymore).
Does that mean an update should also be done on the alignment to keep some consitency (that also include the handle of the `default` case)?

fhost:
Also, should I add a line in the changelog?

src/settings.c
tsipinakis: > Does that mean an update should also be done on the alignment to keep some consitency (that also include the handle of the default case)?

Yes but it's out of the scope of this PR (let's keep things organized). Unfortunately there are a lot of such inconsistencies in the code, I keep putting off doing a general minor cleanup with such changes.

tsipinakis: > So I don't know how you want me to handle this case.
Should I throw an error saying something went wrong and exiting?

An `assert` statement like should be all that's needed. It's not expected for the user to ever see that error so there's no need to be fancy with log messages and since there is specific error check here, `assert(false)` should suffice.

src/x11/x.c
tsipinakis: > So I don't know how you want me to handle this case.
Should I throw an error saying something went wrong and exiting?

An `assert` statement like should be all that's needed. It's not expected for the user to ever see that error so there's no need to be fancy with log messages and since there is specific error check here, `assert(false)` should suffice.

tsipinakis:
> Also, should I add a line in the changelog?

Sure, the changelog when there are no current plans for a release is a bit neglected so if someone adds something I'll take it, if not it'll be added when I do the final pre-release review anyway.

src/settings.c
fhost: I needed to set the ellipsize value in case the user value is invalid, to fallback on the default behaviour (as `middle` is not the first anymore).

tsipinakis:
LGTM 👍

PR #387, opened 2017-09-25
Tests for string append

bebehei:
This is somewhat unrelated to previous PRs, but during test runs it bugged me, that this single skipped test was there.

Additionally, I made the `string_append` function also respect empty strings and concatenate the separator only, if both values are set.

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -78,6 +78,8 @@ char *string_append(char *a, const char *b, const char *sep)
 {
         if (!a)
                 return g_strdup(b);
+        if (!b)
+                return a;
 
         char *new;
         if (!sep)
--- a/src/utils.c
+++ b/src/utils.c
@@ -76,9 +76,11 @@ char *string_replace_all(const char *needle, const char *replacement,
 
 char *string_append(char *a, const char *b, const char *sep)
 {
-        if (!a)
+        if (!a || *a == '\0') {
+                g_free(a);
                 return g_strdup(b);
-        if (!b)
+        }
+        if (!b || *b == '\0')
                 return a;
 
         char *new;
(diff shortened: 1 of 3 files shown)


tsipinakis:
Hmm looking at the behaviour of `string_append` is there a reason we're not using `g_strjoin`?

bebehei:
g_strjoin does not work if a is NULL, as the list is ended and b is avoided

src/utils.c
```diff
@@ -76,9 +76,9 @@ char *string_replace_all(const char *needle, const char *replacement,
 
 char *string_append(char *a, const char *b, const char *sep)
 {
-        if (!a)
+        if (!a || *a == '\0')
```
tsipinakis: `a` isn't freed in this case if it's `\0` which could cause an (albeit minor) memory leak.

I have no clue *why* this function needs to free only `a` but that's how it works it's best to preserve that until we can figure out why.

bebehei: > a isn't freed in this case if it's \0 which could cause an (albeit minor) memory leak.

a very good catch!

bebehei: > I have no clue why this function needs to free only a but that's how it works it's best to preserve that until we can figure out why.

It's cool to have such a function to concatenate some stuff to something existing without having to check for anything:

```
n->msg = string_append(n->msg, addition, "\n");
```

In one line, you appended it if it's a non-empty and not-null string.

bebehei: > a isn't freed in this case if it's \0 which could cause an (albeit minor) memory leak.

Now after freeing `a`, I found a missing `g_strdup` for the first parameter in a unittest. Thanks for that good catch.

bebehei:
IMO this code should get merged now.

All invocations of `g_strconcat` are not this easily replaceable with `string_append`.

tsipinakis:
LGTM

PR #388, opened 2017-09-25
Use config.h for default ellipsize value

fhost:
Get the default value for `ellipsize` from `config.h`, instead of setting it in two different places (in case it is not set, and in case its value is invalid).


tsipinakis:
Thanks!

PR #389, opened 2017-09-26
Add option to handle file:// differently than other url:s

ogg1980:
Add file_handler to your dunstrc.

Example:

file_handler thunar

to open url:s starting with file:// in thunar.


bebehei:
It would be very cool, if you indent the stuff correctly (8 spaces per indentation) and optionally put it in one single commit.

PR #390, opened 2017-09-26
Add option to handle file:// differently than other url:s

ogg1980:
Add file_handler to your dunstrc.

Example:

file_handler thunar

to open url:s starting with file:// in thunar.

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -53,6 +53,7 @@ int monitor = 0;
 char *dmenu = "/usr/bin/dmenu";
 
 char *browser = "/usr/bin/firefox";
+char *file_handler = "/usr/bin/thunar";
 
 int max_icon_size = 0;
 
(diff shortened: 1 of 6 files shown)


tsipinakis:
Why the re-submit? You don't need to send a new PR to update the contents, you can just rebase the branch.

tsipinakis:
This is still missing:
* Documentation in the manpage
* Documentation/default value in `dunstrc`

ogg1980:
I'm still kind of a git rebase noob, but I think I got the hang of it now.

Should the man page include some more elaborate examples? If you read it now, it may seem kind of pointless to handle file:// with a different script.

I'll be using it with something like [this](https://gist.github.com/larsemil/2f20262c6d8e52272647df39a7ee09f8).

tsipinakis:
> Should the man page include some more elaborate examples? If you read it now, it may seem kind of pointless to handle file:// with a different script.

Now that I look this over again, it *does* seem kind-of pointless, isn't it more intuitive to have the browser - or script - that open urls configured to open different programs for each protocol? - `http(s)` with a browser `ftp` with a ftp client and so on.

ogg1980:
Yes, of course, you are right. You can bury this pull request =D.

PR #391, opened 2017-09-26
Replace notifications flickerless

### LINKED ISSUE #320: Flicker with --replace
In a bash shell, run:

```bash
r=0; while true ; do r=$(dunstify --replace=$r --printid "Hi $r - $RANDOM") ; sleep 1 ; done
```

Very often there is a visible flicker when the notification is replaced.

Would it be possible to eliminate it?

bebehei:
Previously, notifications had been replaced by removing the notification
out of the displayed/queue lists, redrawing the window and then adding
the new notification into the queue. This produced a flickering in
dunst.

By avoiding the overhead of closing and opening the window (simply
replacing the datapointer of the list), the flickering disappears.

Fixes #320

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -428,7 +428,6 @@ int notification_init(notification *n, int id)
         if (id == 0) {
                 n->id = ++next_notification_id;
         } else {
-                notification_close_by_id(id, -1);
                 n->id = id;
         }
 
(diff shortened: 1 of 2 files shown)


src/notification.c
```diff
@@ -516,7 +515,8 @@ int notification_init(notification *n, int id)
                 }
                 printf("skipping notification: %s %s\n", n->body, n->summary);
         } else {
-                g_queue_insert_sorted(queue, n, notification_cmp_data, NULL);
+                if (id == 0 || !notification_replace_by_id(n))
```
tsipinakis: I don't see how `id` can be 0 at this point, since it's assigned in line 431.

bebehei: keep in mind: `id` is the parameter here and **not the id of the notification**. if `id == 0`, it indicates, that we should not replace a notification.

tsipinakis: Oh, woops missed that.

tsipinakis:
LGTM 👍

PR #402, opened 2017-10-06
Remove unused struct fields

bebehei:


tsipinakis:
Good catches, just for the record `status` in `cairo_ctx_t` is also unused but it has be removed in `refactor-drawing` since removing it now in master will cause conflicts that I'm not in the mood to resolve.

PR #403, opened 2017-10-06
Work around non-gnu basename implementations

bebehei:
Redo #354, as it got pushed out in #365

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -326,6 +326,7 @@ int notification_init(notification *n, int id)
                   substr = strchr(substr, '%')){
 
                 char pg[16];
+                char *icon_tmp;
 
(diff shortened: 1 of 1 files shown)


tsipinakis:
JFTR I also found an unrelated bug while reviewing this: `n->icon` is initialized after parsing the format so `%I` and `%i` don't work with the default icons. See c55308b8a7c1f79a8dda062eecf41d3a80a6ee7e

Other than that looks good to me.

bebehei:
> JFTR I also found an unrelated bug while reviewing this: n->icon is initialized after parsing the format so %I and %i don't work with the default icons. See c55308b

Yes, this has to be done in a complete refactor of `notificiation_init`.

PR #407, opened 2017-10-08
Remove _GNU_SOURCE macro

### LINKED ISSUE #359: Remove usage of non-portable GNU functions
By defining `_GNU_SOURCE` in many header files we're enabling the usage of many GNU-specific functions that might not be portable to other OSs as demonstrated in #354. We should:

* [x] Figure out what GNU-specific functions we depend on
* [x] Find a suitable replacement or implement one ourselves if one isn't available
* [x] Remove the `_GNU_SOURCE` macro definition

bebehei:
After some research, the _GNU_SOURCE does not affect any function call
anymore. To avoid future failures on other systems not using glibc,
_GNU_SOURCE gets removed.

---
To answer my own question:

> So, how can we require POSIX.1-2008?

I had the wrong mindset about this. There is no dynamic check for different versions of  functions. Also these do not change over time. So you simply don't check for this in your code. Either the function is defined (and then POSIX.1-2008 compatible) or the function is not defined (then it's not POSIX.1-2008) and the code does not compile. So dunst never runs with a wrong libc.

Fixes #359

### DIFF
--- a/src/dunst.c
+++ b/src/dunst.c
@@ -1,6 +1,5 @@
 /* copyright 2012 - 2013 Sascha Kruse and contributors (see LICENSE for licensing information) */
 
-#define _GNU_SOURCE
 #define XLIB_ILLEGAL_ACCESS
 
 #include "dunst.h"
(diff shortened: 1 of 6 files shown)


tsipinakis:
Thanks for looking into this!

PR #409, opened 2017-10-09
Add issue template

bebehei:
Add an issue template and give automatic hints to the users to provide better details.

Any comments?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,24 @@
+<!--
+
+If you want to report bugs, it's the best to get something reproducible!
+
+These program calls might help:
+
+While the notification gets sent:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,24 @@
+<!-- *Delete this comment before submitting*
```
tsipinakis: Why does this need to be deleted, I thought HTML comments don't show up in issues anyway?

bebehei: Yes, it actually makes no sense. My intention was, that the users don't write their message content into the comment by accident.

tsipinakis: Closing the second comment at the end should help to avoid that :)

tsipinakis:
Not sure why I've delayed this - LGTM

PR #416, opened 2017-10-18
Free notification before closing

### LINKED ISSUE #415: dunstify doesn't keep notifications
When specifying the timeout with `-t`, the source code of dunstify tells me that a value of 0 keeps the notification forever, > 0 for the amount of milliseconds and -1 for the server default.

However, when specifying 0, the notification immediately disappears.

bebehei:
The notification object should actually get freed by notify_uninit. But
before freeing the notification, notify_uninit also sends the
CloseNotification signal directly, if the timeout is 0.

But the timeout of 0 should actually indicate a notification, which
never expires.

Fixes #415

### DIFF
--- a/dunstify.c
+++ b/dunstify.c
@@ -324,6 +324,8 @@ int main(int argc, char *argv[])
        g_print("%d\n", get_id(n));
     }
 
+    g_object_unref(G_OBJECT (n));
+
     die(0);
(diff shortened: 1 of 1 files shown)


tsipinakis:
LGTM 👍 

Thanks

PR #417, opened 2017-10-22
Buildstack

bebehei:
To improve code quality, treat warnings in travis now as fatal. This forces better code quality in PRs. Additionally dunstify is built there.

Introduces two build types (debug/release). This will become quite handy when we introduce logging and have a macro for debug messages, which gets activeted by setting a flag.

The makefile got restructured to have clearer structure and splitting of `config.mk` and `Makefile`.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -2,11 +2,6 @@
 PREFIX ?= /usr/local
 MANPREFIX = ${PREFIX}/share/man
 
-VERSION := "1.2.0-non-git"
-ifneq ($(wildcard ./.git/.),)
-VERSION := $(shell git describe --tags)
-endif
-
 # Warning: This is deprecated behavior
 # uncomment to disable parsing of dunstrc
 # or use "CFLAGS=-DSTATIC_CONFIG make" to build
--- a/config.mk
+++ b/config.mk
@@ -11,6 +11,10 @@ CPPFLAGS += -D_DEFAULT_SOURCE -DVERSION=\"${VERSION}\"
 CFLAGS   += -g --std=gnu99 -pedantic -Wall -Wno-overlength-strings -Os ${STATIC} ${CPPFLAGS}
 LDFLAGS  += -lm -L${X11LIB}
 
+CPPFLAGS_DEBUG := -DDEBUG_BUILD
+CFLAGS_DEBUG   := -O0
+LDFLAGS_DEBUG  :=
+
 pkg_config_packs := dbus-1 \
                     gio-2.0 \
                     gdk-3.0 \
--- a/config.mk
+++ b/config.mk
@@ -2,23 +2,24 @@
 PREFIX ?= /usr/local
 MANPREFIX = ${PREFIX}/share/man
 
-# Warning: This is deprecated behavior
 # uncomment to disable parsing of dunstrc
 # or use "CFLAGS=-DSTATIC_CONFIG make" to build
-#STATIC= -DSTATIC_CONFIG
-
-PKG_CONFIG:=$(shell which pkg-config)
-ifeq (${PKG_CONFIG}, ${EMPTY})
-$(error "Failed to find pkg-config, please make sure it is installed")
-endif
+#STATIC= -DSTATIC_CONFIG # Warning: This is deprecated behavior
 
 # flags
 CPPFLAGS += -D_DEFAULT_SOURCE -DVERSION=\"${VERSION}\"
 CFLAGS   += -g --std=gnu99 -pedantic -Wall -Wno-overlength-strings -Os ${STATIC} ${CPPFLAGS}
+LDFLAGS  += -lm -L${X11LIB}
 
-pkg_config_packs := dbus-1 x11 xscrnsaver \
-                    "glib-2.0 >= 2.36" gio-2.0 \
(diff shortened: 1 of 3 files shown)


config.mk
```diff
@@ -8,9 +8,15 @@ MANPREFIX = ${PREFIX}/share/man
...
 # flags
 CPPFLAGS += -D_DEFAULT_SOURCE -DVERSION=\"${VERSION}\"
-CFLAGS   += -g --std=gnu99 -pedantic -Wall -Wno-overlength-strings -Os ${STATIC} ${CPPFLAGS}
+CFLAGS   += --std=gnu99 -pedantic -Wall -Wno-overlength-strings ${STATIC} ${CPPFLAGS}
 LDFLAGS  += -lm -L${X11LIB}
 
+ifeq (Release, ${BUILD_TYPE})
+CFLAGS += -Os
+else
+CFLAGS += -O0 -g
```
tsipinakis: I'm not sure about building with debug by default, I'd prefer it if `debug` was actually a separate target since with the way you implemented it now every downstream user will have to add `BUILD_TYPE=Release` before building.

config.mk
```diff
@@ -8,9 +8,15 @@ MANPREFIX = ${PREFIX}/share/man
 
 # flags
 CPPFLAGS += -D_DEFAULT_SOURCE -DVERSION=\"${VERSION}\"
-CFLAGS   += -g --std=gnu99 -pedantic -Wall -Wno-overlength-strings -Os ${STATIC} ${CPPFLAGS}
+CFLAGS   += --std=gnu99 -pedantic -Wall -Wno-overlength-strings ${STATIC} ${CPPFLAGS}
 LDFLAGS  += -lm -L${X11LIB}
 
+ifeq (Release, ${BUILD_TYPE})
+CFLAGS += -Os
```
tsipinakis: Since now we've been including the debug symbols in the release build (`-g`) and noone has complained so far. Any specific reason for removing them? - This is not a required change just looking for some justification.

config.mk
bebehei: Ok, let's explain my thoughts, why I chose this:

```
debug: CFLAGS += -O0 -g
release: CFLAGS += -Os
```
- `make debug` does nothing
- `make debug release all` does shit
- `make debug all` is the right

```
debug: CFLAGS += -O0 -g
debug: all
```

- `make debug` perfect
- You can't build separate targets as `debug`

---

Using a variable is the best thing, as you can only build a single type (and a variable can hold only a single value).

---

> Since now we've been including the debug symbols in the release build (-g) and noone has complained so far. Any specific reason for removing them?

Well, debug symbols are nothing to go into release.

---

every downstream user will have to add BUILD_TYPE=Release before building.
Also if noone has complained, I complain:

```
96K     dunst
320K    dunst.dbg
```

> every downstream user will have to add BUILD_TYPE=Release before building.

Adding a variable to make something work better (it still works without), I consider this a minor change.

I see something else much more worth: Someone opens an issue, saying dunst segfaults regularly. Then you just have to advise the person to build it from the main repo and trigger the condition.

bebehei:
As you haven't complained about that introduced warning, I'll remove the last commit from this PR (which was solely there to visualize the introduced -Werror in travis).

config.mk
tsipinakis: > Ok, let's explain my thoughts, why I chose this: [...]

I tried doing something like

```
debug: CFLAGS += -O0

ifeq (debug, $(MAKECMDGOALS))
debug: all
endif
```

seems to work but there's still a warning about `debug` doing nothing in the output if `make debug dunstify` for example is used. What do you think?

But I do see the reasoning behind choosing a variable.

---

> Well, debug symbols are nothing to go into release.
every downstream user will have to add BUILD_TYPE=Release before building.
Also if noone has complained, I complain:

My argument for including the debug symbols in is that this is something to be managed by the distribution, not upstream.

Since I maintain packages in Debian I'll give an example from there: Upstream projects are expected (or the maintainer has to make sure) that they include the full debug symbols and then debhelper strips them from the binary and places them in a separate package as `$(PACKAGENAME)-dbgsym`.

I'd expect that those bypassing the distributions and installing manually either are experienced enough to know to disable debug symbols by themselves, or, like you, will need them sometime.

bebehei: > there's still a warning about debug doing nothing in the output if make debug dunstify

Before introducing new warnings in new code, I think we should go on and find a better solution first.

> My argument for including the debug symbols [... saving the full quote]

Your argumentation makes perfectly sense. I'll change this.

bebehei: > I tried doing something like
> ```
> debug: CFLAGS += -O0
> 
> ifeq (debug, $(MAKECMDGOALS))
> debug: all
> endif
> ```
> seems to work

`dmake debug dunstify` doesn't work. the additional `CFLAGS` are only passed to debug target and its dependencies.

There is also no possibility to define a void target:
```
debug: CFLAGS += -O0

ifeq (debug, $(MAKECMDGOALS))
debug: all
else
debug:
    @true
endif
```

This wouldn't work either.

---

Conclusion: We introduce a debug target, which is only useable as standalone.

bebehei: > Conclusion: We introduce a debug target, which is only useable as standalone.

I pushed this now.

PR #418, opened 2017-10-22
Microopt

bebehei:
Minimal changes found over the last month. The only common thing of those commits, is having no common topic.

### DIFF
--- a/src/option_parser.c
+++ b/src/option_parser.c
@@ -248,8 +248,7 @@ int load_ini_file(FILE *fp)
 
                         *end = '\0';
 
-                        if (current_section)
-                                g_free(current_section);
+                        g_free(current_section);
                         current_section = (g_strdup(start + 1));
                         new_section(current_section);
                         continue;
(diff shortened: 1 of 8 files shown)


src/option_parser.c
```diff
@@ -294,7 +294,7 @@ int load_ini_file(FILE *fp)
 
                 add_entry(current_section, key, value);
         }
-        free(line);
+        g_free(line);
```
tsipinakis: Not sure about this one, according to the [glib docs](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html):
> It's important to match g_malloc() (and wrappers such as g_new()) with g_free(), g_slice_alloc() (and wrappers such as g_slice_new()) with g_slice_free(), plain malloc() with free(), and (if you're using C++) new with delete and new[] with delete[]. Otherwise bad things can happen, since these allocators may use different memory pools (and new/delete call constructors and destructors).

bebehei: Good point. I removed the commit.

tsipinakis:
Thanks as usual!

Issue #358, opened 2017-08-13
Rounded Corners Implementation


jonathashenriques:
Would be nice if `Dunst` offers an option to use rounded corners. Is that possible ?

dmitryplatonov:
Yes, that feature would be very useful. Maybe more general way of styling?

bebehei:
Implemented in #420

tsipinakis:
Implemented in 15c252afb051a81de1c86a14c5d839cb34aba5de.

PR #420, opened 2017-10-27
Added: support for round corners

dj95:
As requested in issue [#358](https://github.com/dunst-project/dunst/issues/358) here's the support for round corners via the x11 shape extension.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -28,6 +28,7 @@ pkg_config_packs := dbus-1 \
                     pangocairo \
                     x11 \
                     xinerama \
+                    xext \
                     "xrandr >= 1.5" \
                     xscrnsaver
 
--- a/dunstrc
+++ b/dunstrc
@@ -207,6 +207,13 @@
     # debug: all less than unimportant stuff
     verbosity = mesg
 
+    # Define the corner radius of the notification window
+    # in pixel size. If the radius is 0, you have no rounded
+    # corners.
+    # The radius will be automatically lowered if it exceeds half of the
+    # notification height to avoid clipping text and/or icons.
+    corner_radius = 0
+
     ### Legacy
 
     # Use the Xinerama extension instead of RandR for multi-monitor support.
--- a/src/draw.c
+++ b/src/draw.c
@@ -172,6 +172,8 @@ static struct dimensions calculate_dimensions(GSList *layouts)
         dim.h += 2 * settings.frame_width;
         dim.h += (g_slist_length(layouts) - 1) * settings.separator_height;
 
+        dim.corner_radius = settings.corner_radius;
+
         int text_width = 0, total_width = 0;
         for (GSList *iter = layouts; iter; iter = iter->next) {
                 colored_layout *cl = iter->data;
@@ -217,6 +219,8 @@ static struct dimensions calculate_dimensions(GSList *layouts)
                         dim.h += h;
                         text_width = MAX(w, text_width);
                 }
+
+                dim.corner_radius = MIN(dim.corner_radius, h/2);
         }
 
         if (dim.w <= 0) {
@@ -389,12 +393,103 @@ static int layout_get_height(colored_layout *cl)
         return MAX(h, h_icon);
 }
 
+static void draw_rounded_rect(cairo_t *c, int x, int y, int width, int height, int corner_radius, bool first, bool last)
+{
+        const float degrees = M_PI / 180.0;
+
+        if (first && last) {
+                cairo_new_sub_path(c);
+                cairo_arc(c,
+                          x + width - corner_radius,
+                          y + corner_radius,
+                          corner_radius,
+                          -90 * degrees,
+                          0 * degrees);
+                cairo_arc(c,
+                          x + width - corner_radius,
+                          y + height - corner_radius,
+                          corner_radius,
+                          0 * degrees,
+                          90 * degrees);
+                cairo_arc(c,
+                          x + corner_radius,
+                          y + height - corner_radius,
+                          corner_radius,
+                          90 * degrees,
+                          180 * degrees);
+                cairo_arc(c,
+                          x + corner_radius,
+                          y + corner_radius,
+                          corner_radius,
+                          180 * degrees,
+                          270 * degrees);
+                cairo_close_path(c);
+        } else if (first) {
+                cairo_new_sub_path(c);
+                cairo_arc(c,
+                          x + width - corner_radius,
+                          y + corner_radius,
+                          corner_radius,
+                          -90 * degrees,
+                          0 * degrees);
+                cairo_arc(c,
+                          x + width,
+                          y + height,
+                          0,
+                          0,
+                          0);
+                cairo_arc(c,
+                          x,
+                          y + height,
+                          0,
+                          0,
+                          0);
+                cairo_arc(c,
+                          x + corner_radius,
+                          y + corner_radius,
+                          corner_radius,
+                          180 * degrees,
+                          270 * degrees);
+                cairo_close_path(c);
+        } else if (last) {
+                cairo_new_sub_path(c);
+                cairo_arc(c,
+                          x + width,
+                          y,
+                          0,
+                          0,
+                          0);
+                cairo_arc(c,
+                          x + width - corner_radius,
+                          y + height - corner_radius,
+                          corner_radius,
+                          0 * degrees,
+                          90 * degrees);
+                cairo_arc(c,
+                          x + corner_radius,
+                          y + height - corner_radius,
+                          corner_radius,
+                          90 * degrees,
+                          180 * degrees);
+                cairo_arc(c,
+                          x,
+                          y,
+                          0,
+                          180 * degrees,
+                          270 * degrees);
+                cairo_close_path(c);
+        } else {
+                cairo_rectangle(c, x, y, width, height);
+        }
+}
+
 static cairo_surface_t *render_background(cairo_surface_t *srf,
                                           colored_layout *cl,
                                           colored_layout *cl_next,
                                           int y,
                                           int width,
                                           int height,
+                                          int corner_radius,
                                           bool first,
                                           bool last,
                                           int *ret_width)
@@ -403,12 +498,15 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
 
         cairo_t *c = cairo_create(srf);
 
-        if (first) height += settings.frame_width;
-        if (last) height += settings.frame_width;
-        else height += settings.separator_height;
+        if (first)
+                height += settings.frame_width;
+        if (last)
+                height += settings.frame_width;
+        else
+                height += settings.separator_height;
 
         cairo_set_source_rgb(c, cl->frame.r, cl->frame.g, cl->frame.b);
-        cairo_rectangle(c, x, y, width, height);
+        draw_rounded_rect(c, x, y, width, height, corner_radius, first, last);
         cairo_fill(c);
 
         /* adding frame */
@@ -426,7 +524,7 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
                 height -= settings.separator_height;
 
         cairo_set_source_rgb(c, cl->bg.r, cl->bg.g, cl->bg.b);
-        cairo_rectangle(c, x, y, width, height);
+        draw_rounded_rect(c, x, y, width, height, corner_radius, first, last);
         cairo_fill(c);
 
         if (   settings.sep_color != SEP_FRAME
@@ -504,7 +602,7 @@ static struct dimensions layout_render(cairo_surface_t *srf,
         int bg_width = 0;
         int bg_height = MAX(settings.notification_height, (2 * settings.padding) + cl_h);
 
-        cairo_surface_t *content = render_background(srf, cl, cl_next, dim.y, dim.w, bg_height, first, last, &bg_width);
+        cairo_surface_t *content = render_background(srf, cl, cl_next, dim.y, dim.w, bg_height, dim.corner_radius, first, last, &bg_width);
         cairo_t *c = cairo_create(content);
 
         render_content(c, cl, bg_width);
--- a/src/draw.c
+++ b/src/draw.c
@@ -393,94 +393,58 @@ static int layout_get_height(colored_layout *cl)
         return MAX(h, h_icon);
 }
 
+/**
+ * Create a path on the given cairo context to draw the background of a notification.
+ * The top corners will get rounded by `corner_radius`, if `first` is set.
+ * Respectably the same for `last` with the bottom corners.
+ */
 static void draw_rounded_rect(cairo_t *c, int x, int y, int width, int height, int corner_radius, bool first, bool last)
 {
         const float degrees = M_PI / 180.0;
 
-        if (first && last) {
-                cairo_new_sub_path(c);
-                cairo_arc(c,
-                          x + width - corner_radius,
-                          y + corner_radius,
-                          corner_radius,
-                          -90 * degrees,
-                          0 * degrees);
+        cairo_new_sub_path(c);
+
+        if (last) {
+                // bottom right
                 cairo_arc(c,
                           x + width - corner_radius,
                           y + height - corner_radius,
                           corner_radius,
-                          0 * degrees,
-                          90 * degrees);
+                          degrees * 0,
+                          degrees * 90);
+                // bottom left
                 cairo_arc(c,
                           x + corner_radius,
                           y + height - corner_radius,
                           corner_radius,
-                          90 * degrees,
-                          180 * degrees);
-                cairo_arc(c,
-                          x + corner_radius,
-                          y + corner_radius,
-                          corner_radius,
-                          180 * degrees,
-                          270 * degrees);
-                cairo_close_path(c);
-        } else if (first) {
-                cairo_new_sub_path(c);
-                cairo_arc(c,
-                          x + width - corner_radius,
-                          y + corner_radius,
-                          corner_radius,
-                          -90 * degrees,
-                          0 * degrees);
-                cairo_arc(c,
-                          x + width,
-                          y + height,
-                          0,
-                          0,
-                          0);
-                cairo_arc(c,
-                          x,
-                          y + height,
-                          0,
-                          0,
-                          0);
-                cairo_arc(c,
-                          x + corner_radius,
-                          y + corner_radius,
-                          corner_radius,
-                          180 * degrees,
-                          270 * degrees);
-                cairo_close_path(c);
-        } else if (last) {
-                cairo_new_sub_path(c);
-                cairo_arc(c,
-                          x + width,
-                          y,
-                          0,
-                          0,
-                          0);
-                cairo_arc(c,
-                          x + width - corner_radius,
-                          y + height - corner_radius,
-                          corner_radius,
-                          0 * degrees,
-                          90 * degrees);
-                cairo_arc(c,
-                          x + corner_radius,
-                          y + height - corner_radius,
-                          corner_radius,
-                          90 * degrees,
-                          180 * degrees);
-                cairo_arc(c,
-                          x,
-                          y,
-                          0,
-                          180 * degrees,
-                          270 * degrees);
-                cairo_close_path(c);
+                          degrees * 90,
+                          degrees * 180);
         } else {
-                cairo_rectangle(c, x, y, width, height);
+                cairo_line_to(c, x + width, y + height);
+                cairo_line_to(c, x,         y + height);
         }
+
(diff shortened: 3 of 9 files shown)


src/x11/x.h
```diff
@@ -4,6 +4,7 @@
 
 #include <X11/X.h>
 #include <X11/Xlib.h>
+#include <X11/extensions/shape.h>
```
bebehei: h > c. Please sort it.

config.mk
```diff
@@ -23,7 +23,8 @@ pkg_config_packs := dbus-1 \
                     x11 \
                     xinerama \
                     xrandr \
-                    xscrnsaver
+                    xscrnsaver \
+					xext
```
bebehei: Keep track of your indentation please. For alignment you have to use spaces.

dunstrc
```diff
@@ -198,6 +198,9 @@
     # automatically after a crash.
     startup_notification = false
 
+    # window corners
```
bebehei: Please add this stuff to the manpage, too and elaborate it a bit more and give a unit. For a passive user, this could be anything (em, inch, pixels).

config.mk
bebehei: Also e < s. Please sort the libs correctly.

bebehei:
I tweaked now the `frame` and `corner` options a bit to make dunst very ugly. I don't claim, that there is a user, who is so stupid, to configure dunst this way, but this explains the stuff, which should get fixed first:

screenshot from 2017-10-27 13-06-21

- For some reasons, compton (or in general the compositing manager) does not get informed, that the window is round.
- Also when having `frame > 0`, the corners of the frame get rounded, but not the actual notification window. So `frame` and `corner` cannot get combined.

dunstrc
```diff
@@ -198,6 +198,9 @@
...
+    corner = 0
```
tsipinakis: `rounded_corner_radius` would be a better name for this setting in my opinion.

src/x11/x.c
```diff
@@ -1014,7 +1070,44 @@ void x_setup(void)
...
+
+        int rad = settings.corner_radius;
+        int dia = 2 * rad;
+
+        XSetForeground(xctx.dpy, shape_gc, 0);
+        XFillRectangle(xctx.dpy, mask, shape_gc, 0, 0, width, height);
+
+        XSetForeground(xctx.dpy, shape_gc, 1);
+
+        XFillArc(xctx.dpy, mask, shape_gc, 0, 0, dia, dia, 0, 23040);
```
tsipinakis: * `2304` is an ugly way to say `360*64`, it'll be optimized away by the compiler anyway and it's much better for readability to write it out.

* This is the first time I'm looking into the X drawing API so correct me if I'm wrong but is there a reason you're drawing the entire 360 degrees instead of just the ones needed? Is there a downside to it?

src/x11/x.c
```diff
@@ -1014,7 +1070,44 @@ void x_setup(void)
...
+        XFillArc(xctx.dpy, mask, shape_gc, width-dia-1, 0, dia, dia, 0, 23040);
```
tsipinakis: Style nitpick: Please put a space after each `-` to make it obvious that it's an expression.

src/x11/x.c
```diff
@@ -1014,7 +1070,44 @@ void x_setup(void)
...
+        int width = win_attr.width + win_attr.border_width;
+        int height = win_attr.height + win_attr.border_width;
+
+        Pixmap mask = XCreatePixmap(xctx.dpy, xctx.win, width, height, 1);
+        XGCValues xgcv;
+
+        GC shape_gc = XCreateGC(xctx.dpy, mask, 0, &xgcv);
+
+        int rad = settings.corner_radius;
+        int dia = 2 * rad;
```
tsipinakis: The diameter of the corner should be capped to the window width/height so that it doesn't cause rendering weirdness in case it's bigger than the window size.

src/x11/x.c
dj95: To the second point: There's no specific reason to draw the entire 360 degrees. I just did it for simplicity. I'm going to test the behaviour with just filling the need arc.

bebehei:
screenshot from 2017-10-29 13-56-27

Better: Now, there is frame in the corner. But the frame of the corner and the frame of the side do not match.

dj95:
screenshot_2017-10-29_14-52-45

Fixed the behaviour that frame and corner frame doesnt match in the following commit. This only happens, when the corner radius is bigger than the height of the notification. I restricted it to the maximum height as @tsipinakis stated out in his review.

bebehei:
@dj95 Your screenshot looks correct. 

I'm using dc10dce

screenshot from 2017-10-29 15-36-43

There's a frame missing in the corners.

screenshot from 2017-10-29 15-43-01

Ah, there's the missing frame. But, well 😕 

Have you pushed the commit, you made the screenshot with?

dj95:
@bebehei Yes, I did. Can you post your configuration for the frame and corner radius so I can investigate? I played with the config values and for me it seem to work fine. :thinking:

bebehei:
@dj95 Here it is:[dunstrc.txt](https://github.com/dunst-project/dunst/files/1425224/dunstrc.txt)

bebehei:
I have to admit, I've set horrible values. But IMHO dunst should be able to handle (or sanitize) them.

Interestingly, I got now 3 notifications from thunderbird. There, the corner frame was correct. But after a short time, the lower one began to flap.

dj95:
I know. But to ensure the software quality dunst should be able to sanitize those values :)

src/x11/x.c
```diff
@@ -833,6 +890,8 @@ gboolean x_mainloop_fd_dispatch(GSource *source, GSourceFunc callback,
                 switch (ev.type) {
                 case Expose:
                         if (ev.xexpose.count == 0 && xctx.visible) {
+                                x_win_round_corners();
+
                                 x_win_draw();
```
bebehei: There are many more calls of x_win_draw(). One for example is in [`run()` of dunst.c](https://github.com/dunst-project/dunst/blob/2d7dc6fbc592bf7a764286d8d8d42fda7db7549b/src/dunst.c#L223).

For my understanding the rounded corners have to be called every time the window gets drawn. Why don't you put it into `x_win_draw` directly?

dj95:
screenshot_2017-10-29_19-18-13
screenshot_2017-10-29_19-18-28

Now the problem of a too big radius is fixed and capped to the half of one notification :blush:

config.mk
```diff
@@ -22,17 +22,18 @@ pkg_config_packs := dbus-1 \
...
                     x11 \
                     xinerama \
+                    xext \
                     xrandr \
                     xscrnsaver
 
 # check if we need libxdg-basedir
 ifeq (,$(findstring STATIC_CONFIG,$(CFLAGS)))
-	pkg_config_packs += libxdg-basedir
+    pkg_config_packs += libxdg-basedir
```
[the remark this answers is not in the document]
dj95: I did this one because the rest of the indentation in this file uses spaces, not tabs. But I'm going to revert this if you want it.

src/x11/x.c
dj95: I do it in the following commit. I first thought, that its only necessary on the expose event.

src/x11/x.c
```diff
@@ -584,15 +584,46 @@ static dimension_t x_render_layout(cairo_t *c, colored_layout *cl, colored_layou
         int bg_height = MAX(settings.notification_height, (2 * settings.padding) + h);
         double bg_half_height = settings.notification_height/2.0;
         int pango_offset = (int) floor(h/2.0);
+        double degrees = M_PI / 180.0;
```
tsipinakis: This should be a `const`, we don't need to modify it.

src/x11/x.c
```diff
@@ -584,15 +584,46 @@ static dimension_t x_render_layout(cairo_t *c, colored_layout *cl, colored_layou
...
 
-        cairo_set_source_rgb(c, cl->frame.r, cl->frame.g, cl->frame.b);
+        cairo_set_source_rgba(c, cl->bg.r, cl->bg.g, cl->bg.b, 0);
         cairo_rectangle(c, bg_x, bg_y, bg_width, bg_height);
         cairo_fill(c);
 
+        if (settings.frame_width > 0) {
+                if (first && last) {
+                        cairo_new_sub_path (c);
+                        cairo_arc (c, bg_x + bg_width - settings.corner_radius, bg_y + settings.corner_radius, settings.corner_radius, -90 * degrees, 0 * degrees);
```
tsipinakis: Another style nitpick but I think this is borderline unreadable. I'd prefer if it was one parameter per line for long function calls like so:
```c
                cairo_arc (c,
                           bg_x + bg_width - corner_radius,
                           bg_y + corner_radius,
                           corner_radius,
                           -90 * degrees,
                           0 * degrees);
```

Same goes for `XFillArc` and `XFillRectangle`. Though I'm not 100% on this decision. @bebehei What do you think?

(I'm aware that this is the case for other calls in the same function but since it came up now might as-well discuss it and maybe start fixing it.)

bebehei: >  @bebehei What do you think?

Well, I'd like to have parameters in one line. I prefer it.

src/x11/x.c
```diff
@@ -588,10 +590,55 @@ static dimension_t x_render_layout(cairo_t *c, colored_layout *cl, colored_layou
         if (last) bg_height += settings.frame_width;
         else bg_height += settings.separator_height;
 
-        cairo_set_source_rgb(c, cl->frame.r, cl->frame.g, cl->frame.b);
+        cairo_set_source_rgba(c, cl->bg.r, cl->bg.g, cl->bg.b, 255);
```
bebehei: Please use RGB<del>A</del>. Everything else is also based on RGB

src/x11/x.c
bebehei: @dj95 He meant to solely mark it with `const` (instead of defining a new macro).

dj95:
I hope I haven't fucked up the commit history... Maybe you are more experienced to squash those commits.

dj95:
Or shall I squash all commits I made in this PR to one commit?

bebehei:
> Or shall I squash all commits I made in this PR to one commit?

Yes please.

bebehei:
@tsipinakis has shown me this once: https://chris.beams.io/posts/git-commit/

dj95:
And I destroyed the functionality... damn. I'll take a look tomorrow

dj95:
Nevermind. It was just a wrong variable type. Amazing how different the graphical behaviour is :smile: I hope all fixes are now present :)

tsipinakis:
Simply from looking at the diff I see that it still includes irrelevant changes from other commits (looks like parts of #421 are in there).

Can you rebase on master and remove them?

Also it might be beneficial to trim the commit message - we don't need to know what commits got squashed.

bebehei:
@dj95 Before you work further, do this:

```
git clone https://github.com/dj95/dunst dunst-new
cd dunst-new
git checkout master
git pull --ff-only http://github.com/dunst-project/dunst/ master
git checkout feature/round_corners
git reset --hard master
git pull --ff-only http://github.com/bebehei/dunst/ 420-rebase
git push -u origin feature/round_corners
```

dj95:
I'm sorry that you need to wait that long, was a busy last week. I think the PR now contains just the changes that are needed for the round corners feature. Tests on my machine (and hopefully yours, too) are working fine. I'm a bit afraid to squash the commits because last time I made a mistake and squashed a merge from your master branch. I need to practice it a bit with a test-project first(This is my first pull request I've ever made :D ).

bebehei:
@dj95 The two commits are intended.

tsipinakis:
@bebehei I'm not sure about that matrix formatting, it makes it unreadable on smaller windows (equivalent to the old rules formatting)
2017-11-09t12-13-54

bebehei:
@tsipinakis That's the reason, why I have done this rearranging in different commits.

Checkout both and decide, which fits better.

bebehei:
@tsipinakis You decide, whether we should use the matrix commit or not.

bebehei:
There are a few points still to handle (sorry for approving too early):

1. @tsipinakis pointed out in private IRC, that the frame does not work, if `corner_radius` is unset.
2. With this PR, dunst  is damn slow. I don't know, where this problem is, but X lags and consumes much CPU (because of dunst).
    - A small benchmark would be `for i in {0..1000}; do dunstify --raw_icon=anypic.png --replace=$i a b; done`. Just compare the speed of the number rushing up.
3. Icons exceeding the notification are not cut/rounded.
4. There is still the shadow issue (only nice to have).

src/x11/x.c
```diff
@@ -1030,7 +1098,51 @@ void x_setup(void)
         x_win_setup();
         x_cairo_setup();
         x_shortcut_grab(&settings.history_ks);
+}
+
+void x_win_round_corners(void)
```
bebehei: Please add a comment describing the method. Especially, **when** it should be called.

src/x11/x.c
```diff
@@ -584,14 +584,56 @@ static dimension_t x_render_layout(cairo_t *c, colored_layout *cl, colored_layou
...
+
+        int width = win_attr.width + win_attr.border_width;
+        int height = win_attr.height + win_attr.border_width;
+
+        int min_width_height = width < height ? width : height;
+        int half_min_width_height = (int) floor(min_width_height / 2.0);
+
+        settings.corner_radius = MIN(half_min_width_height, settings.corner_radius);
+
+        x_win_round_corners();
```
bebehei: This function call sounds contrary to my understanding of this method. You create/render the layout in here, and then round the whole window (for every notification). But you call it also at the end of `x_win_draw`, where it makes IMHO much more sense.

src/x11/x.c
```diff
@@ -584,14 +584,56 @@ static dimension_t x_render_layout(cairo_t *c, colored_layout *cl, colored_layou
...
+        XWindowAttributes win_attr;
+        XGetWindowAttributes(xctx.dpy, xctx.win, &win_attr);
+
+        int width = win_attr.width + win_attr.border_width;
+        int height = win_attr.height + win_attr.border_width;
+
+        int min_width_height = width < height ? width : height;
+        int half_min_width_height = (int) floor(min_width_height / 2.0);
+
+        settings.corner_radius = MIN(half_min_width_height, settings.corner_radius);
```
bebehei: Is there any reason, touching the settings struct over and over?

src/x11/x.c
```diff
@@ -584,14 +584,56 @@ static dimension_t x_render_layout(cairo_t *c, colored_layout *cl, colored_layou
...
 
         if (first) bg_height += settings.frame_width;
         if (last) bg_height += settings.frame_width;
         else bg_height += settings.separator_height;
 
-        cairo_set_source_rgb(c, cl->frame.r, cl->frame.g, cl->frame.b);
-        cairo_rectangle(c, bg_x, bg_y, bg_width, bg_height);
-        cairo_fill(c);
+        XWindowAttributes win_attr;
+        XGetWindowAttributes(xctx.dpy, xctx.win, &win_attr);
```
bebehei: I'd like to see, that we extract the width from cairo instead of X. Regarding wayland support, this will be much easier to port.

bebehei:
@dj95 I found a great "tutorial" today. It works like: Oh, I want to do this. But I have done that. How can I fix it? https://github.com/k88hudson/git-flight-rules

It might be a great read and it might be also a great handbook to use, when you're trapped in some (weird) git situation and you want to get out of it. These weird git situations happen regularly at the beginning. I have enough experience with these. And also remember: `git status`, `git log (-p)` and `git diff (--staged)` are your friends.

tsipinakis:
Whoops, #448 created conflicts here, sorry! Should've seen that coming.

Friendly ping @dj95 since you've havent given an update here in a while

tsipinakis:
@dj95 Are you still going to work on this or should I take over?

dj95:
It would be very kind if you could do this. I have no time at the moment to work on this pr. Sorry.

tsipinakis:
@bebehei I've ported the changes over (with a few adjustments) to master. The original commits were basically unsalvageable due to merge conflicts (sorry @dj95 !).

Can you give this another review?

bebehei:
> Can you give this another review?

Sorry, but your port went horribly wrong. Attached some screenshots.

screenshot from 2018-05-27 13-12-55
screenshot from 2018-05-27 13-11-36

Config from https://github.com/bebehei/df/blob/0ad21457bc91a0b0d794bb990122b77099840745/i3/dunst/dunstrc (and `-show_age_threshold 1`) and compton running as compositing manager.

tsipinakis:
Oops, there was an if statement leftover that entirely bypassed the background rendering if `frame_width == 0`, fixed!

bebehei:
Looks much better. I'll review it later.

bebehei:
Same configuration as before. Additonally a `corner_radius` set to `2000`.


screenshot from 2018-05-27 15-25-20

screenshot from 2018-05-27 15-26-13


screenshot from 2018-05-27 15-25-35


I think the approach to limit the corner radius to half the height of the notification is a bit counterintuitive, what do you think?

tsipinakis:
> I think the approach to limit the corner radius to half the height of the notification is a bit counterintuitive, what do you think?

It is but I don't see a reasonable alternative. If we force it to use the higher radius it just looks broken.

bebehei:
Instead of mocking about the continuously repeating code, I had a try to touch X11/cairo code for myself. And it was a great learning experience.

> It is but I don't see a reasonable alternative. If we force it to use the higher radius it just looks broken.

I guess we have to document this and give out the hint, that the user shall not choose a radius, which is bigger than half the height (or maybe even the width if the user has got a superweird geometry) of the smallest expected notification.

src/x11/x.c
```diff
@@ -81,18 +81,18 @@ static void x_win_move(window_x11 *win, int x, int y, int width, int height)
         }
 }
 
-void x_win_round_corners(window_x11 *win, const int rad)
+static void x_win_round_corners(window_x11 *win, const int rad)
 {
         const int width = win->dim.w;
         const int height = win->dim.h;
+        const int dia = 2 * rad;
+        const int degrees = 64;
```
tsipinakis: I'd add a comment here to explain *why* 64, or just point to the docs.

bebehei: Err, this is [because of the docs](https://tronche.com/gui/x/xlib/graphics/filling-areas/XFillArc.html) :tm:.

tsipinakis: I'm aware, but someone looking at it a few months or years in the future is going to be really confused at this random value.

bebehei: Good point. It definitely warrants a comment.

bebehei:
> With the latest commits separator_color = frame is broken but I cannot figure out what broke it, bisect says 040be9a. Keep in mind we don't redraw the separator on top if it's the same color as the frame, we just leave a gap when rendering the background (at least that's how it used to work).

Fuck it. Sorry for the strong language. Let's draw the separator actively. I'm forging the commit right now.

src/draw.c
```diff
@@ -393,94 +393,58 @@ static int layout_get_height(colored_layout *cl)
...
                           x + corner_radius,
                           y + height - corner_radius,
                           corner_radius,
-                          90 * degrees,
-                          180 * degrees);
+                          degrees * 90,
+                          degrees * 180);
+        } else {
+                cairo_line_to(c, x + width, y+width);
+                cairo_line_to(c, x, y+width);
```
tsipinakis: > Let's draw the separator actively. I'm forging the commit right now.

While true it's simple to just draw over it and be done. Things not working like this can indicate some type of problem in the implementation. Like here, you're adding the width to the vertical axis.

tsipinakis:
I documented that `corner_radius` is automatically lowered in the man page and dunstrc. I'd call this one ready to be merged if you agree, with one suggestion being to drop c475a0b, if everything is implemented properly that optimisation should work as intended and if not there's a good chance something else was broken alongside so it's worth keeping in IMO.

bebehei:
> with one suggestion being to drop c475a0b,

Simultaneously yes and no. I appreciate dropping this commit. But we're definitely missing here a test. AFAIK, I'm the third person stumbling over this bit in a PR and the only person, who found out about this, had been you.

tsipinakis:
> But we're definitely missing here a test.

We're definitely are, I'm looking into converting out tests to glib right now and after that I'll look into adding tests for the drawing part.

tsipinakis:
Dropped it which dismissed your review, oops!

tsipinakis:
Thanks @dj95 for the initial implementation it's been a while since we've implemented a major feature and this brings a nice change of pace.

(And as usual thanks @bebehei for the help/review!)

PR #425, opened 2017-10-31
Free notifications ignored in history

bebehei:
Notifications marked with `history_ignore = 1` (and therefore transient ones too) are currently never freed.

### DIFF
--- a/src/queues.c
+++ b/src/queues.c
@@ -188,7 +188,6 @@ int queues_notification_close_id(int id, int reason)
                 notification *n = iter->data;
                 if (n->id == id) {
                         g_queue_remove(displayed, n);
-                        queues_history_push(n);
                         target = n;
                         break;
                 }
@@ -198,15 +197,19 @@ int queues_notification_close_id(int id, int reason)
              iter = iter->next) {
                 notification *n = iter->data;
                 if (n->id == id) {
+                        assert(target == NULL);
                         g_queue_remove(waiting, n);
-                        queues_history_push(n);
                         target = n;
                         break;
                 }
         }
 
-        if (reason > 0 && reason < 4 && target != NULL) {
-                notification_closed(target, reason);
+        if (target) {
+
+                if (reason > 0 && reason < 4)
+                        notification_closed(target, reason);
+
+                queues_history_push(target);
         }
 
         return reason;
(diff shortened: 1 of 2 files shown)


src/queues.c
```diff
@@ -183,30 +183,29 @@ int queues_notification_close_id(int id, int reason)
 {
         notification *target = NULL;
 
-        for (GList *iter = g_queue_peek_head_link(displayed); iter;
+        for (GList *iter = g_queue_peek_head_link(displayed);
+             iter && !target;
```
tsipinakis: What's the point of this change? All it's doing it adding a bit of overhead because now we have to check `target` on each iteration and hurts readability.

src/queues.c
```diff
@@ -183,30 +183,29 @@ int queues_notification_close_id(int id, int reason)
...
                 }
         }
 
-        for (GList *iter = g_queue_peek_head_link(waiting); iter;
+        for (GList *iter = g_queue_peek_head_link(waiting);
+             iter && !target;
              iter = iter->next) {
                 notification *n = iter->data;
                 if (n->id == id) {
                         g_queue_remove(waiting, n);
```
tsipinakis: Might be worth adding an `assert(target == NULL)` to catch a possible bug where a notification exists in both queues here.

tsipinakis:
Oops, approved the wrong PR, my review still stands.

src/queues.c
bebehei: I see, this is actually an error and we should use `assert` instead of silently skipping the `waiting` queue.

bebehei:
> Oops, approved the wrong PR, my review still stands.

You can dismiss the review. (What I have done for you).

src/queues.c
bebehei: But actually, to make this check valuable, we should not break after assigning `target = n`. We should further traverse the whole notification queue, because then there are no doubled ids.

src/queues.c
```diff
@@ -198,15 +197,16 @@ int queues_notification_close_id(int id, int reason)
...
                 if (n->id == id) {
+                        assert(target == NULL);
                         g_queue_remove(waiting, n);
-                        queues_history_push(n);
                         target = n;
                         break;
                 }
         }
 
         if (reason > 0 && reason < 4 && target != NULL) {
```
tsipinakis: I missed this earlier, if somehow `reason` is outside of the defined values the notification is also never freed.

bebehei: 🤦‍♂️ Thank you.

PR #428, opened 2017-11-02
Refactor notification init

bebehei:
Refactors `notification_init` to split it into multiple methods.

### DIFF
--- a/src/markup.c
+++ b/src/markup.c
@@ -4,6 +4,8 @@
 
 #include <assert.h>
 #include <stdbool.h>
+#include <string.h>
+#include <stdio.h>
 
 #include "settings.h"
 #include "utils.h"
@@ -42,6 +44,178 @@ static char *markup_br2nl(char *str)
         str = string_replace_all("<br/>", "\n", str);
         str = string_replace_all("<br />", "\n", str);
         return str;
+}
+
+/*
+ * Remove HTML hyperlinks of a string.
+ *
+ * @str: The string to replace a tags
+ * @urls: (nullable): If any href-attributes found, an '\n' concatenated
+ *        string of the URLs in format '[<text between tags>] <href>'
+ */
+void markup_strip_a(char **str, char **urls)
+{
+        char *tag1 = NULL;
+
+        if (urls)
+                *urls = NULL;
+
+        while ((tag1 = strstr(*str, "<a"))) {
+                // use href=" as stated in the notification spec
+                char *href = strstr(tag1, "href=\"");
+                char *tag1_end = strstr(tag1, ">");
+                char *tag2 = strstr(tag1, "</a>");
+
+                // the tag is broken, ignore it
+                if (!tag1_end) {
+                        fprintf(stderr,
+                                "WARNING: Given link is broken: '%s'\n",
+                                tag1);
+                        string_replace_at(*str, tag1-*str, strlen(tag1), "");
+                        break;
+                }
+                if (tag2 && tag2 < tag1_end) {
+                        int repl_len =  (tag2 - tag1) + strlen("</a>");
+                        fprintf(stderr,
+                                "WARNING: Given link is broken: '%.*s.'\n",
+                                repl_len, tag1);
+                        string_replace_at(*str, tag1-*str, repl_len, "");
+                        break;
+                }
+
+                // search contents of href attribute
+                char *plain_url = NULL;
+                if (href && href < tag1_end) {
+
+                        // shift href to the actual begin of the value
+                        href = href+6;
+
+                        const char *quote = strstr(href, "\"");
+
+                        if (quote && quote < tag1_end) {
+                                plain_url = g_strndup(href, quote-href);
+                        }
+                }
+
+                // text between a tags
+                int text_len;
+                if (tag2)
+                        text_len = tag2 - (tag1_end+1);
+                else
+                        text_len = strlen(tag1_end+1);
+
+                char *text = g_strndup(tag1_end+1, text_len);
+
+                int repl_len = text_len + (tag1_end-tag1) + 1;
+                repl_len += tag2 ? strlen("</a>") : 0;
+
+                *str = string_replace_at(*str, tag1-*str, repl_len, text);
+
+                // if there had been a href attribute,
+                // add it to the URLs
+                if (plain_url && urls) {
+                        text = string_replace_all("]", "", text);
+                        text = string_replace_all("[", "", text);
+
+                        char *url = g_strdup_printf("[%s] %s", text, plain_url);
+
+                        *urls = string_append(*urls, url, "\n");
+                        g_free(url);
+                }
+
+                g_free(plain_url);
+                g_free(text);
+        }
+}
+
+/*
+ * Remove img-tags of a string. If alt attribute given, use this as replacement.
+ *
+ * @str: The string to replace img tags
+ * @urls: (nullable): If any src-attributes found, an '\n' concatenated string of
+ *        the URLs in format '[<alt>] <src>'
+ */
+void markup_strip_img(char **str, char **urls)
+{
+        const char *start = *str;
+
+        if (urls)
+                *urls = NULL;
+
+        while ((start = strstr(*str, "<img"))) {
+                const char *end = strstr(start, ">");
+
+                // the tag is broken, ignore it
+                if (!end) {
+                        fprintf(stderr, "WARNING: Given image is broken: '%s'\n", start);
+                        string_replace_at(*str, start-*str, strlen(start), "");
+                        break;
+                }
+
+                // use attribute=" as stated in the notification spec
+                const char *alt_s = strstr(start, "alt=\"");
+                const char *src_s = strstr(start, "src=\"");
+
+                char *text_alt = NULL;
+                char *text_src = NULL;
+
+                const char *src_e = NULL, *alt_e = NULL;
+                if (alt_s)
+                        alt_e = strstr(alt_s + strlen("alt=\""), "\"");
+                if (src_s)
+                        src_e = strstr(src_s + strlen("src=\""), "\"");
+
+                // Move pointer to the actual start
+                alt_s = alt_s ? alt_s + strlen("alt=\"") : NULL;
+                src_s = src_s ? src_s + strlen("src=\"") : NULL;
+
+                /* check if alt and src attribute are given
+                 * If both given, check the alignment of all pointers */
+                if (   alt_s && alt_e
+                    && src_s && src_e
+                    && (  (alt_s < src_s && alt_e < src_s-strlen("src=\"") && src_e < end)
+                        ||(src_s < alt_s && src_e < alt_s-strlen("alt=\"") && alt_e < end)) ) {
+
+                        text_alt = g_strndup(alt_s, alt_e-alt_s);
+                        text_src = g_strndup(src_s, src_e-src_s);
+
+                /* check if single valid alt attribute is available */
+                } else if (alt_s && alt_e && alt_e < end && (!src_s || src_s < alt_s || alt_e < src_s - strlen("src=\""))) {
+                        text_alt = g_strndup(alt_s, alt_e-alt_s);
+
+                /* check if single valid src attribute is available */
+                } else if (src_s && src_e && src_e < end && (!alt_s || alt_s < src_s || src_e < alt_s - strlen("alt=\""))) {
+                        text_src = g_strndup(src_s, src_e-src_s);
+
+                } else {
+                        fprintf(stderr,
+                                "WARNING: Given image argument is broken: '%.*s'\n",
+                                (int)(end-start), start);
+                }
+
+                // replacement text for alt
+                int repl_len = end - start + 1;
+
+                if (!text_alt)
+                        text_alt = g_strdup("[image]");
+
+                *str = string_replace_at(*str, start-*str, repl_len, text_alt);
+
+                // if there had been a href attribute,
+                // add it to the URLs
+                if (text_src && urls) {
+                        text_alt = string_replace_all("]", "", text_alt);
+                        text_alt = string_replace_all("[", "", text_alt);
+
+                        char *url = g_strdup_printf("[%s] %s", text_alt, text_src);
+
+                        *urls = string_append(*urls, url, "\n");
+                        g_free(url);
+                }
+
+                g_free(text_src);
+                g_free(text_alt);
+        }
 }
 
 /*
@@ -96,6 +270,8 @@ char *markup_transform(char *str, enum markup_mode markup_mode)
                 break;
         case MARKUP_FULL:
                 str = markup_br2nl(str);
+                markup_strip_a(&str, NULL);
+                markup_strip_img(&str, NULL);
                 break;
         }
 
--- a/src/menu.c
+++ b/src/menu.c
@@ -95,10 +95,14 @@ char *extract_urls(const char *to_match)
  */
 void open_browser(const char *in)
 {
-        // remove prefix and test url
-        char *url = extract_urls(in);
-        if (!url)
-                return;
+        char *url = NULL;
+
+        // If any, remove leading [ linktext ] from URL
+        const char *end = strstr(in, "] ");
+        if (*in == '[' && end)
+                url = g_strdup(end + 2);
+        else
+                url = g_strdup(in);
 
         int browser_pid1 = fork();
 
--- a/src/notification.c
+++ b/src/notification.c
@@ -40,9 +40,9 @@ void notification_print(notification *n)
         printf("\turgency: %s\n", notification_urgency_to_string(n->urgency));
         printf("\ttransient: %d\n", n->transient);
         printf("\tformatted: '%s'\n", n->msg);
-        printf("\tfg: %s\n", n->color_strings[ColFG]);
-        printf("\tbg: %s\n", n->color_strings[ColBG]);
-        printf("\tframe: %s\n", n->color_strings[ColFrame]);
+        printf("\tfg: %s\n", n->colors[ColFG]);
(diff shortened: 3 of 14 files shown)


src/notification.c
```diff
@@ -280,55 +283,96 @@ char *notification_extract_markup_urls(char **str_ptr) {
...
+ *  - its undefined representation (NULL, -1)
+ *
+ * This function is guaranteed to return a valid pointer.
+ * @Returns: The generated notification
  */
 notification *notification_create(void)
 {
-        return g_malloc0(sizeof(notification));
-}
+        //FIXME: handle OOM
```
tsipinakis: From [glibs docs](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html)
> If any call to allocate memory fails, the application is terminated. This also means that there is no need to check if the call succeeded.

Means we don't need to handle OOM.

src/notification.c
```diff
@@ -280,55 +283,96 @@ char *notification_extract_markup_urls(char **str_ptr) {
...
-
         if (n->raw_icon == NULL && n->icon == NULL) {
                 n->icon = g_strdup(settings.icons[n->urgency]);
         }
 
-        n->urls = notification_extract_markup_urls(&(n->body));
+        /* Color hints */
+        n->colors[ColFG] = n->colors[ColFG]
+                         ? n->colors[ColFG]
+                         : xctx.colors[ColFG][n->urgency];
```
tsipinakis: I *really* don't like using `xctx` inside non-X related functions. Though I guess this will have to stay as-is until the drawing refactor is done.

src/notification.c
```diff
@@ -429,41 +473,25 @@ void notification_init(notification *n)
...
-
-        n->timestamp = g_get_monotonic_time();
-
-        n->redisplayed = false;
-
-        n->first_render = true;
+static void notification_extract_urls(notification *n)
+{
+        // DO markup urls processing here until we split this out correctly
+        n->urls = notification_extract_markup_urls(&(n->body)); //WHAT THE FUCK!?
```
tsipinakis: I'm not sure what the `WTF` is here for? - Sure not the best code but not really weird?

bebehei: Thank you for mentioning this again.

For me, this function call is against every logic. Why do we put a double pointer to that (and not the notification and turn it into a method like every other method).

I also don't see, why we do have to make changes to the body field there. Markup should be extracted somewhere else.

src/notification.c
bebehei: I agree with you. But previously, `notification_init` also had to access this struct. It's bad, but it's not getting worse (and hopefully gets fixed in refactor drawing).

src/notification.c
bebehei: Thanks. This wasn't clear for me. Only [g_try_malloc0](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-try-malloc0) mentions the behavior of `g_malloc0`.

src/notification.c
bebehei: 1. There is a discrepancy between pango and notification spec. Notification spec requires `img` and `a` tags, while pango cannot render these (obviously). As `a` tags happen quite regularly, these have to get replaced. Guess, what this method did, too!?

2. Let's have these two notifications:
    - `notify-send asdf '<a href="https://github.com/">testname</a>'`
        - ` n->urls = 0x5555558675f0 "[#1] https://github.com/"`
    - `notify-send https://summary-url.de https://internal.com`
        - `1: n->urls = 0x555555803980 "https://summary-url.de\nhttps://internal.com"`
    - Mhmm, same field, different format, 👎

bebehei:
@tsipinakis As I'm tired right now and may have introduced some bugs by myself with the lastest commits, I will review this by myself at first. I will "ping" you by re-requesting a review.

---

OT: I would be able to click on merge for my own PRs, while you requested changes in your current review. Would you mind changing this to make a positive review a neccessity?

tsipinakis:
> OT: I would be able to click on merge for my own PRs, while you  requested changes in your current review. Would you mind changing this  to make a positive review a neccessity?

I tweaked the branch restrictions a bit and set it to require a positive review and a passing build to push to master. Thought technically you'd be able to review and approve your own commit, right?

bebehei:
> Thought technically you'd be able to review and approve your own pr, right?

I guess not. Let's test it!

bebehei:
@tsipinakis has enlightened me about the previous functionality of `notification_extract_markup_urls` right now in IRC.

I heavily want to drop this ugly function and replace it with something similar. The specific functionality of adding `[ #linkno]` in front of the URL is quite bogus for me. I don't see why this is neccessary. IMHO, there is `show_indicators` for such functionality. And also having the linknumber included in the urls field makes it quite ugly, as there is no consistency to `extract_urls`.

Currently I have also removed it in this PR and I'd like to evaluate which path we want to go.

bebehei:
> [...] and I'd like to evaluate which path we want to go.

@tsipinakis Could you please give your opinion here?

src/notification.h
```diff
@@ -30,34 +30,44 @@ typedef struct _actions {
 } Actions;
 
 typedef struct _notification {
+        int id;
```
tsipinakis: I'd group this differently: `id` and `dbus_client` together and `appname`, `summary`, `body`, `category`, `urgency` in another group since they're the main notification attributes.

tsipinakis:
> The specific functionality of adding [ #linkno] in front of the URL is quite bogus for me. I don't see why this is necessary.

As I've explained on IRC, as I view it this is because there is no clear indication of which url belongs to which piece of text in the notification when it comes to `a href` tags. For plain urls this is not necessary since it's displayed as text anyway.

> IMHO, there is show_indicators for such functionality.

`show_indicators` shows that there's a url in the notification - it doesn't provide any information as to which piece of text contains which url.

> (Edit: I also found, that notification spec mentions to show it in standard blue underlined text.)

We should probably comply with this but it might be a bit misleading since clicking on it won't work with the current implementation.

Currently I have also removed it in this PR and I'd like to evaluate which path we want to go.

bebehei:
> We should probably comply with this but it might be a bit misleading since clicking on it won't work with the current implementation.

Question to understand this: Does pango support $something to know, that the user clicked on a specific textbit?

src/notification.h
bebehei: Grouped it accordingly now.

tsipinakis:
> Does pango support $something to know, that the user clicked on a specific textbit?

Quick google didn't turn up anything - I have no clue.

bebehei:
> it doesn't provide any information as to which piece of text contains which url.

what's better than a linknumber? Put in the actual text! So, if you open up dmenu it will display `[linktext] URL`.

e.g:
```
notify-send '<a href="https://github.com/dunst-project/dunst/pull/428">GitHub</a>' 'Pull Request updated'
```

will result in:

```
n->urls = "[GitHub] https://github.com/dunst-project/dunst/pull/428"
```

This is also more comfortable.

tsipinakis:
> what's better than a linknumber? Put in the actual text! So, if you open up dmenu it will display [linktext] URL.

While it seems like a good suggestion it could break in some scenarios like having a super long link text, or having two links with the same text

> The first example is _this_ and the second is _this_, which one do you prefer?

would produlce

> [this] URL     [this] URL

While the order would give it away somewhat, it's still confusing.

bebehei:
> [...] or having two links with the same text [...] While the order would give it away somewhat, it's still confusing.

I agree with you. It's actually confusing. But is it dunst's job to make confusing URLs clear? There is no information lost on dunst's behalf.

> While it seems like a good suggestion it could break in some scenarios like having a super long link text, 

Is a long URL currently handled better?

```
[ Hi Mike, I've got you a super long URL,
which renders faulty on your system but
it's good! Check it out! #1]
```

- How long do you need to parse that the URL number 1 takes the whole notification?
- What's the probability of having a second URL in a notification with a superlong URL?

---

I have to admit, I have only searched for a replacement of the current URL replacement, but I think the solution I came up with is actually much better than the current approach.

src/notification.c
tsipinakis: Also, I can't help but feel like all these ternary operators are a bit ugly - what's wrong with a usual `if`?

src/menu.c
```diff
@@ -95,8 +95,18 @@ char *extract_urls(const char *to_match)
  */
 void open_browser(const char *in)
 {
-        // remove prefix and test url
-        char *url = extract_urls(in);
+
+        // remove leading [ linktext ] from URL
```
tsipinakis: If we're removing the prefix here what's the point of still running it through `extract_urls`?

src/markup.c
```diff
@@ -64,7 +64,7 @@ static char *markup_strip_a(char *str)
                 // If we hit the end of the string, delete the trailing tag fully
                 unsigned int len = end ? (end-start)+1 : strlen(start);
 
-                str = string_replace_at(str, start-str, len, "");
+                str = string_replace_at(str, start-str, len, "<span color=\"blue\"><u>");
```
tsipinakis: While it's specified by the spec, I can already see all the customization requests a mile away, especially since it looks ugly even with the default theme.

2017-11-29t18-31-03

src/markup.c
```diff
@@ -44,6 +45,66 @@ static char *markup_br2nl(char *str)
...
+static char *markup_strip_img(char *str)
+{
+        const char *start = str;
+        const char *end = NULL;
+        const char *alt = NULL;
+        char *repl = NULL;
+
+        while ((start = strstr(start, "<img"))) {
+                end = strstr(start, ">");
+                alt = strstr(start, "alt=\""); // use alt=" as stated in the notification spec
```
tsipinakis: How about handling `src` as a URL too?

src/notification.c
```diff
@@ -438,45 +498,31 @@ void notification_init(notification *n)
...
-        n->redisplayed = false;
+static void notification_extract_urls(notification *n)
+{
+        g_clear_pointer(&n->urls, g_free);
 
-        n->first_render = true;
+        char *concat = g_strconcat(n->summary, " ", n->body, NULL);
+        // remove <a href=... to not confuse extract_urls
+        char *concat_plain = markup_transform(g_strdup(concat), MARKUP_FULL);
+        char *urls_plain;
```
tsipinakis: What's the point of this variable? You're assigning and freeing it.

src/notification.c
```diff
@@ -248,95 +255,148 @@ void notification_replace_single_field(char **haystack,
...
+
+                // search contents of href attribute
+                if (href && href < tag1_end) {
+
+                        // shift href to the actual begin of the value
+                        href = href+6;
+
+                        const char *quote = strstr(href, "\"");
+
+                            if (quote && quote < tag1_end) {
```
tsipinakis: What's up with the indentation here?

tsipinakis:
> Is a long URL currently handled better?
> [...]

Good point, is the time it takes to implement it and the switching costs worth it though? What are the upsides?

src/notification.c
bebehei: It **is** actually used. But you're objection has a good reason. The way this variable gets assigned is shit.

src/notification.c
bebehei: When you trust vim....

src/markup.c
```diff
@@ -44,6 +45,66 @@ static char *markup_br2nl(char *str)
...
+        str = string_replace_all("</a>", "", str);
+
+        const char *start = str;
+        const char *end = NULL;
+
+        while ((start = strstr(start, "<a"))) {
+                end = strstr(start, ">");
+
+                // If we hit the end of the string, delete the trailing tag fully
+                unsigned int len = end ? (end-start)+1 : strlen(start);
```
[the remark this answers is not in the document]
bebehei: Yes.

src/markup.c
bebehei: Please, no. Wait for the feature request.

bebehei:
> I feel like this PR has grown from a simple refactor to way beyond scope changes -.-

Simple refactoring of `notification_init` is a first class oxymoron. 😂 

> Good point, is the time it takes to implement it and the switching costs worth it though? What are the upsides?

?

The upsides should be clear and the work is already done!?

Honestly, I don't get what you want to tell here.

src/markup.c
bebehei: Yes, this *is* ugly. I just pushed it as an example how the ugly the recommendation will look like, when we discussed, what the replacement for the link replacement method.

What would you like to do? Remove it fully and mention it to use the indicators? Make it more sensible by just underlining it?

src/menu.c
bebehei: I thought I can catch some weird corner case with this. But after thinking again about it, I've proved myself wrong in this.

src/markup.c
tsipinakis: I would say remove it for now and file an issue about it to look into it later.

tsipinakis:
> The upsides should be clear and the work is already done!?

What I was trying to say is "there is no reason to put in effort to fix what already works" but if it's already implemented I'm fine with merging it.

bebehei:
@tsipinakis  I have reworked the URL handling. It gets now done inside of `markup_strip_(a|img)`. I also added the behavior of adding image URLs. And it gets done over `n->msg` (the shown part of the notification).

But: This is all on my laptop. I won't commit it today. Tiredly reassembling the history is no good.

bebehei:
@tsipinakis I fixed all issues with the URL extraction methods. I added pelnty of tests. Have fun with "Remove a and img tags from msg". ;-)

src/notification.c
```diff
@@ -438,45 +475,25 @@ void notification_init(notification *n)
...
         n->urls = string_append(n->urls, tmp_urls, "\n");
         g_free(tmp_urls);
+        g_free(tmp);
+}
 
+static void notification_dmenu_string(notification *n)
+{
         if (n->actions) {
-                n->actions->dmenu_str = NULL;
+                g_clear_pointer(&n->actions->dmenu_str, g_free);
```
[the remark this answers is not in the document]
bebehei: There is no guard against calling this function a second time. Setting `dmenu_str` solely to `NULL` is irresponsible and makes absolutely no sense.

I've done this in every function, which rewrites some fields.

src/markup.c
```diff
@@ -44,6 +46,189 @@ static char *markup_br2nl(char *str)
...
+
+                        const char *quote = strstr(href, "\"");
+
+                        if (quote && quote < tag1_end) {
+                                plain_url = g_strndup(href, quote-href);
+                        }
+                }
+
+                // text between a tags
+                int text_len = tag2
```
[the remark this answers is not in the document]
bebehei: You marked the exact line to let this code look as confusing as possible. 😂 

I fully agree with you.

src/markup.c
```diff
@@ -44,6 +46,189 @@ static char *markup_br2nl(char *str)
...
+                char *text_src = NULL;
+
+                const char *src_e = NULL, *alt_e = NULL;
+                if (alt_s)
+                        alt_e = strstr(alt_s + strlen("alt=\""), "\"");
+                if (src_s)
+                        src_e = strstr(src_s + strlen("src=\""), "\"");
+
+                // Move pointer to the actual start
+                alt_s = alt_s ? alt_s+5 : NULL;
```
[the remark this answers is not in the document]
bebehei: Oh, yes. That's right.

src/markup.c
```diff
@@ -44,6 +46,189 @@ static char *markup_br2nl(char *str)
...
+
+                // if there had been a href attribute,
+                // add it to the URLs
+                if (text_src && urls) {
+
+                        // If text_alt is not given, the image is
+                        // invisible in the replaced text
+                        if (!text_alt || strlen(text_alt) == 0) {
+                                g_free(text_alt);
+                                text_alt = g_strdup("Invisible Image");
```
tsipinakis: Instead of using `Invisible Image` for everything let's fall back to our previous system, numbers.

`Notification with inline image [image #1]`
And in dmenu
`[#1] url`

test/markup.c
```diff
@@ -45,12 +45,97 @@ TEST test_markup_transform(void)
...
+        g_free(out_urls);
+        g_free(out);
+        g_free(msg);
+
+        PASS();
+}
+
+TEST test_markup_strip_img(void)
+{
+        RUN_TESTp(helper_markup_strip_img, "v <img> img",                                         "v  img",       NULL);
```
tsipinakis: Can tests really be nested like that? I don't see anything indicating so in `greatest` docs.

bebehei: Yes and no. The code is indicating it, but it's nowhere documented.

It works and it's no problem. In functionality it's far better than calling an extra function.

I wish, there would be an additional layer between test and assertion.

src/markup.c
bebehei: I agree. Setting the text to "Invisible image" sounds rather mysterious than helpful.

test/markup.c
bebehei: Ok a small background info on this: When an assertion fails, it immediately returns the current function (which should be a test).

As we test multiple strings (done via multiple assertions), a failed assertion breaks the test function. By calling another subtest, the supertest does not fail and will test further for all strings.

src/markup.c
bebehei: Sorry, I've seen this just now.

> Notification with inline image [image #1]
> And in dmenu
> [#1] url

Please, no. As a user, I'd question the rationale removing it for links and then adding it again for images. As a user, I'd really insult this dev (myself).

Simply changing it to `[Hidden Image]` makes more sense IMO.

bebehei:
@tsipinakis I've rebased the fixes already into their commits. If you want to see the actual diff, please see here: https://github.com/bebehei/dunst/compare/refactor-notification_init...refactor-notification_init-backup

bebehei:
Finally!

Issue #431, opened 2017-11-05
Queue handling with settings.sort


bebehei:
I looked today at #301. On the first sight, the fix seems to be easy (just inserting it at top from `display`). But I found some other issues:

When notifications get sorted, the a full `displayed` queue overruns urgent notifications and lets them stay in `waiting` queue until the normal notifications timed out. Then the noficiation gets shown.

## reproducing

```
for i in {0..5}; do notify-send a $i; done
notify-send -u critical a b
```

The notification "a b" will get shown after all notifications have timed out. "a {1...5}" have timed out.

# Solutions

None of my thoughts are satisfying yet.

### Solution1

Merge `waiting` and `displayed` together. Add another `gint64 last_shown` field in notification struct and then add in `queues.c` a new function for `GSList *get_displayed(gint64 lastshown, uint size)`.

It returns the first `size` notifications of `displayed`. If `n->last_shown != lastshown` (-> not shown before) it does equivalent things to notification as previously moving from `waiting` to `displayed` and also it know, if it has ever shown before (`n->start == 0`)

Of course, `notification_cmp_data` should handle `redisplayed` here, too.

### Solution2

Enhance the waiting insertion process: Make a shallow copy of `displayed` first (call it `displayed'`).  Then insert sorted the nofication to `displayed'`. Check if `displayed'->tail->data == notification` and if so, push it to `waiting` or remove the last element from `displayed` and push then the new notification into it.

PR #439, opened 2017-11-15
Duplicate notifications

### LINKED ISSUE #427: Behaviour of duplicate notifications (again)
As part of #411 there was a change to the way dunst assigns ids to duplicate notifications, instead of assigning a new id for the merged notifications it re-uses the old one. While that seemed like a good idea at the time of the review it makes much less sense now that I think about it and leaves some open problems on duplicate notifications:

* The [notification spec](https://developer.gnome.org/notification-spec/#commands) states that ids should not be re-used unless all ids have been exhausted
    > The return value is a UINT32 that represent the notification. It is unique, and will not be reused unless a MAXINT number of notifications have been generated.

* Misses actions
    If a new notification is sent and stacked with an existing one, the actions of the first notification are used and the second ones are ignored.
    We should look into merging them (and maybe track for which notification each action belongs to so that we can handle them appropriately)

* Confusion with `CloseNotification` signal
    Previously when dunst returned a new id for each stacked notification only the id of the original notification could be used to close it. Now this is fixed since we return the same id but as mentioned above the notification spec has a note against that and it might confuse some clients.

---

For a proposed solution: Keep all merged in an array inside the original notifications struct (and assign a new id to each one). This will allow us to handle each duplicate notification as a separate one while still merging them into one when displaying them.

The downsides:
* Slightly increased memory usage (which can be trimmed by only storing the actions, id and anything else relevant for the duplicates)
* Complicates the iteration/lookup code for the notifications
    If we need to close a notification by id for example not we don't only need to check all queues but every duplicate notification inside each queue as well.

### LINKED ISSUE #438: Crash when replacing and popping too much notifications
The freshly introduced assertion https://github.com/dunst-project/dunst/blob/62e50289f0ead4cc0074b5ed7314302caf72688c/src/queues.c#L196-L200 breaks on a special case.

When replacing notifications, we push the old notification to the queue. So, replacing more notifications than `displayed->length` and then popping th replaced ones all again will result, that at least one replaced notification hangs in `queue`, while `target` is already set.

Related #427 

### Proposed solution

Do not push replaced notification to history. Free them directly.

### Installation info

- Version: `master` after refactor queues merge

bebehei:
So, these would be my two cents for these queue problems. (#431 yet to handle somewhere else)

Fixes #438 
Fixes #427

### DIFF
--- a/src/queues.c
+++ b/src/queues.c
@@ -54,6 +54,7 @@ unsigned int queues_length_history()
 
 int queues_notification_insert(notification *n, int replaces_id)
 {
+
         /* do not display the message, if the message is empty */
         if (strlen(n->msg) == 0) {
                 if (settings.always_run_script) {
@@ -74,6 +75,8 @@ int queues_notification_insert(notification *n, int replaces_id)
 
         if (replaces_id == 0) {
 
+                n->id = ++next_notification_id;
+
                 if (settings.stack_duplicates) {
                         int stacked = queues_stack_duplicate(n);
                         if (stacked > 0) {
@@ -81,8 +84,6 @@ int queues_notification_insert(notification *n, int replaces_id)
                                 return stacked;
                         }
                 }
-
-                n->id = ++next_notification_id;
 
                 g_queue_insert_sorted(waiting, n, notification_cmp_data, NULL);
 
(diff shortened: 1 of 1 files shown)


bebehei:
ping päng pong. @tsipinakis

PR #440, opened 2017-11-16
Use more enums

bebehei:
Use enums for close reasons and also consistently for urgencies.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -303,7 +303,7 @@ static void on_close_notification(GDBusConnection *connection,
 {
         guint32 id;
         g_variant_get(parameters, "(u)", &id);
-        queues_notification_close_id(id, 3);
+        queues_notification_close_id(id, REASON_SIG);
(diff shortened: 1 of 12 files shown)


src/dbus.c
```diff
@@ -324,12 +324,20 @@ static void on_get_server_information(GDBusConnection *connection,
 
 void notification_closed(notification *n, enum reason reason)
 {
+        int valid_reason = REASON_UNDEF;
+
+        if (REASON_INVALID_MIN < reason && reason < REASON_INVALID_MAX) {
+                valid_reason = reason;
```
tsipinakis: Any reason for using a different variable here? Why not just `reason = REASON_UNDEF`?

bebehei: I actually dislike it, changing invocation parameters. But your objection is correct. I've changed it.

bebehei:
ping päng pong. @tsipinakis

PR #442, opened 2017-11-22
Dunstify: Print id before blocking

tsipinakis:
Print the notification id, if requested, before blocking to catch action
return or NotificationClosed signal since once of those events happens
the program will exit without further action, bypassing the id printing
logic.

### DIFF
--- a/dunstify.c
+++ b/dunstify.c
@@ -296,9 +296,10 @@ int main(int argc, char *argv[])
 
     GMainLoop *l = NULL;
 
-    if (block || action_strs)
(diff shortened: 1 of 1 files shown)


dunstify.c
```diff
@@ -296,9 +296,10 @@ int main(int argc, char *argv[])
 
     GMainLoop *l = NULL;
 
-    if (block || action_strs)
+    if (block || action_strs) {
```
bebehei: On the first sight, this looks suspicious but makes fully sense.

Issue #446, opened 2017-11-24
Support image-path


bebehei:
According to the notification spec, dunst should take care of the `image-path` hint. The notification spec [ranks it higher than the normal icon](https://developer.gnome.org/notification-spec/#id2776582). As GApplications [send their icon data via `image-path`](https://github.com/GNOME/glib/blob/b97e5cb21bf97bcdd8c393f032c6eef453f4f410/gio/gfdonotificationbackend.c#L270-L288), dunst should also include this.

bebehei:
Fixed in #447

PR #447, opened 2017-11-24
Support image-path hint

bebehei:
As GApplications expose their icon via image-path hint and not via the
actual app_icon field, GApplication icons haven't been displayed before.

According to the notification spec, image-path hint should also override
the app_icon as follows:

1. image-data hint
2. image-path hint
3. app_icon parameter

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -204,6 +204,13 @@ static void on_notify(GDBusConnection *connection,
                                                 g_variant_unref(dict_value);
                                         }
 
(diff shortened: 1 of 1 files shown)


bebehei:
Fixes #446

bebehei:
> The priorities are a bit backwards here. In x.c the raw icon always takes priority over the path but that's not how it is in the specification

Except for `icon_data`, we handle everything right. IMHO it's ok to "ignore" the spec here and care about more important things.

PR #449, opened 2017-11-26
Fix notification timeouts

### LINKED ISSUE #132: Notifications not timing out/closing by themselves
I'm running dwm with two monitors. Despite using the following config options (I think these are the only relevant ones):

```
idle_threshold = 0 
sticky_history = no

[urgency_low]
    timeout = 2

[urgency_normal]
    timeout = 2

[urgency_critical]
    timeout = 2
```

...I still have notifications linger for much longer than the 2 seconds specified, if X isn't interacted with. If I do one of the following:
- Change window focus between monitors (using mouse or keyboard)
- Change current dwm tag
- Raise/lower volume using keypresses (which seems to cause the current window to briefly lose focus, perhaps an artifact how xbindkeys works)
- Set the rootname (changing the dwm status bar)—this happens every 60 seconds to update a clock display
- Click the notification itself with the mouse

...then the lingering notification will disappear.

I'm using dunst 1.0.0. Let me know if there's any more information I can provide.

### LINKED ISSUE #433: Notifications still shown albeit timed out
This is a separate issue from #132, but may have some overlap.

Sending two notifications with `notify-send -t 1000 asdf asdf1 && notify-send -t 10000 asdf asdf10`, the shorter notification does not get undrawn, when it timed out.

Verifying the behavior of `queues_check_timeouts`, shows that notifications do get removed from the displayed queue, but the notifications are still drawn in dunst's window.

https://github.com/dunst-project/dunst/blob/62e50289f0ead4cc0074b5ed7314302caf72688c/src/queues.c#L289-L292

In the #411, I [removed the `wake_up` call](https://github.com/dunst-project/dunst/commit/7f335b79d25ee3891df6d0207404d522a050481d#diff-19d89f5a0764f788a38b7f01a68d3751L79) from there. If I port it back, it works, but I do not see, why it's neccessary, as we're already in `wake_up`/`run` and reexecuting this, should give an equivalent state.

### Installation info

- Version: `master > 1.2.0`

bebehei:
Notifications did not get undisplayed under certain conditions.

When sending two notifications simultaneously, both notifications get closed properly, but the shorter one gets undisplayed when  the longer one vanishes.

Fixes #433 

When sending notifications in this order: a long one, a short one and a middle one. Then there is no timeout for the middle one created, because there is an existing timeout and the timeout for the short notification fires faster. Then at the next execution of `run` (closing the short notification), `next_timeout` contains still the value of the short nofitication and is still smaller than the timeout for the middle notification. But dunst does not force to set a new timeout, as there is another existing timeout for the long notification.

Fixes #132

### DIFF
--- a/src/dunst.c
+++ b/src/dunst.c
@@ -53,12 +53,7 @@ gboolean run(void *data)
         queues_check_timeouts(x_is_idle());
         queues_update();
 
-        static int timeout_cnt = 0;
         static gint64 next_timeout = 0;
-
-        if (data && timeout_cnt > 0) {
-                timeout_cnt--;
-        }
 
         if (!xctx.visible && queues_length_displayed() > 0) {
                 x_win_show();
@@ -78,10 +73,9 @@ gboolean run(void *data)
                 gint64 timeout_at = now + sleep;
 
                 if (sleep >= 0) {
-                        if (timeout_cnt == 0 || next_timeout < now || timeout_at < next_timeout) {
-                                g_timeout_add(sleep/1000, run, mainloop);
+                        if (next_timeout < now || timeout_at < next_timeout) {
+                                g_timeout_add(sleep/1000, run, NULL);
                                 next_timeout = timeout_at;
-                                timeout_cnt++;
                         }
                 }
         }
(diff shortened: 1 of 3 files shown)


tsipinakis:
> The commit bd8fd8b introduced, that specific notifications did not time

Did you point to the wrong commit here? That's just a whitespace refactor.

bebehei:
@tsipinakis That's obviously wrong. Probably there was something wrong in my clipboard. I found the wrong commit again and have updated the stuff.

PR #453, opened 2017-12-06
Don't emit NotificationClosed signal on redisplayed notifications

tsipinakis:
The `notification_closed` function is called even when a notification has
been pulled from history and re-emits the `NotificationClosed` signal,
this can confuse clients since it's not accounted for in the spec.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -298,7 +298,7 @@ static void on_notify(GDBusConnection *connection,
 
         // The message got discarded
         if (id == 0) {
-                notification_closed(n, 2);
+                signal_notification_closed(n, 2);
                 notification_free(n);
         }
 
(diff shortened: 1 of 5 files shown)


bebehei:
Good catch. But I'd like to have this separated from DBus. I think we should fix it in `queues_notification_close`.

src/dbus.c
```diff
@@ -333,6 +333,9 @@ static void on_get_server_information(GDBusConnection *connection,
 
 void notification_closed(notification *n, enum reason reason)
 {
+        if (n->redisplayed)
```
bebehei: Good catch. But I'd like to have this separated from DBus. I think we should fix it in queues_notification_close.

PR #454, opened 2017-12-06
Distinguish DBus connection and name acquired error

### LINKED ISSUE #451: I don't have any notification on my debian 9.2.
### Installation info

- Version: `v1.2.0-172-g75fddab`
- Install type: `package (apt) & manuel`
- Distro and version: ` Debian 9.2`

I use dunst on my computer (debian sid) and it runs. Now I want to install it on my laptop (debian 9.2). I installed with apt (1.1.0-2+b1) but it doesn't run. After I compilated it form the source (branch: master and debian) but it's same. For the test, I use send-notify.

If I compare with my computer: I have a running process:
  $ ps -eF | grep dunst
  user    2388     1  0 145708 14944  1 10:12 ?        00:00:00 /usr/bin/dunst
I searched in the systemd if I have a service actif where but no. How systemd started it? 

I don't know how to debug my problem. I don't have log or debug's message.

Thank you for your help.
PS: Sorry for my english, i'm a french user

bebehei:
Distinguish between a failure to connect to DBus and a failure to
connect to the actual FDN DBus path.

This is possible, because `on_name_acquired` gets its DBus connection
passed as a parameter. If this parameter is NULL, there is no connection
established and the DBus name is not lost, because another notification
daemon is running.

Fixes #451

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -418,7 +418,11 @@ static void on_name_lost(GDBusConnection *connection,
                          const gchar *name,
                          gpointer user_data)
 {
-        fprintf(stderr, "Name Lost. Is Another notification daemon running?\n");
+        if (connection)
+                fprintf(stderr, "Cannot acquire 'org.freedesktop.Notifications'."
+                                "Is Another notification daemon running?\n");
+        else
+                fprintf(stderr, "Cannot connect to DBus.\n");
         exit(1);
 }
 


src/dbus.c
```diff
@@ -418,7 +418,11 @@ static void on_name_lost(GDBusConnection *connection,
                          const gchar *name,
                          gpointer user_data)
 {
-        fprintf(stderr, "Name Lost. Is Another notification daemon running?\n");
+        if (dbus_conn)
+                fprintf(stderr, "Cannot connect to '/org/freedesktop/Notifications'."
```
tsipinakis: * Minor nitpick but I'd say `Cannot acquire` or `Cannot register` is better wording.

* For a slight more serious mistake: `/org/freedesktop/Notifications` is the name of the object, the name is `org.freedesktop.Notifications`. Look [here](https://dbus.freedesktop.org/doc/dbus-tutorial.html#interfaces) for more details.

src/dbus.c
```diff
@@ -418,7 +418,11 @@ static void on_name_lost(GDBusConnection *connection,
                          const gchar *name,
                          gpointer user_data)
 {
-        fprintf(stderr, "Name Lost. Is Another notification daemon running?\n");
+        if (dbus_conn)
```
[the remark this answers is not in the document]
bebehei: Yes. Full ACK. My thought was that this catches another corner case, when `on_name_acquired` did not work properly and this will ensure, that dunst works better. But actually, the program's dying here. 😂

src/dbus.c
bebehei: > For a slight more serious mistake: /org/freedesktop/Notifications is the name of the object, the name is org.freedesktop.Notifications. Look here for more details.

The slashes are right.  We have to acquire the path (here: name). We do not acquire the interface. We have to acquire the path. If another process is implementing the interface, we don't care. But if another process is acquiring the path, we've got a problem.

bebehei: I proved myself wrong. After reading the code again, I see that these have to be dots. I think, I have to rule out some DBus misconceptions by reading the docs fully.

PR #462, opened 2017-12-19
Add valgrind to travis configuration and check for leaks

bebehei:
So, while we're already starting to improve our testsuite, I'd like to throw this one in, too.

It checks for definite leaks now in our testsuite, and more important, it'll fail if any leaks are found.

For this, we have to ignore that small xdg basedir leak and then tell valgrind to exit with an error code, when a definite leak shows up.


bebehei:
Ouch. There is obviously something wrong with `string_replace_at`.

PR #463, opened 2017-12-21
Add configurable path variables for services

bebehei:
Setting PREFIX to a location different to /usr, the install routine
fails to install the systemd and dbus service files. These are
installed, but in the PREFIX directory and not /usr. DBus and systemd
usually only read their files from /usr/ and ignore files in /usr/local.

Now by default, we're asking pkg-config, where to install it. Mostly,
this will be /usr and this conflicts the FHS. But it's the user's intent
to install dunst and (possibly) override the package manager's files
belonging to dunst.

At the current point, even DBus ignores the PREFIX and installs its
systemd service file to the location specified by pkg-config.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,16 +8,29 @@ ifneq ($(wildcard ./.git/.),)
 VERSION := $(shell git describe --tags)
 endif
 
+ifeq (,${SYSTEMD})
+# Check for systemctl to avoid discrepancies on systems, where
+# systemd is installed, but systemd.pc is in another package 
+systemctl := $(shell command -v systemctl >/dev/null && echo systemctl)
+ifeq (systemctl,${systemctl})
+SYSTEMD := 1
+else
+SYSTEMD := 0
+endif
+endif
+
 SERVICEDIR_DBUS ?= $(shell pkg-config dbus-1 --variable=session_bus_services_dir)
 SERVICEDIR_DBUS := ${SERVICEDIR_DBUS}
 ifeq (,${SERVICEDIR_DBUS})
 $(error "Failed to query pkg-config for package 'dbus-1'!")
 endif
 
+ifneq (0,${SYSTEMD})
 SERVICEDIR_SYSTEMD ?= $(shell pkg-config systemd --variable=systemduserunitdir)
 SERVICEDIR_SYSTEMD := ${SERVICEDIR_SYSTEMD}
 ifeq (,${SERVICEDIR_SYSTEMD})
 $(error "Failed to query pkg-config for package 'systemd'!")
+endif
 endif
 
 LIBS := $(shell pkg-config --libs   ${pkg_config_packs})
@@ -78,10 +91,15 @@ doc: docs/dunst.1
 docs/dunst.1: docs/dunst.pod
 	pod2man --name=dunst -c "Dunst Reference" --section=1 --release=${VERSION} $< > $@
 
-.PHONY: service
-service:
+.PHONY: service service-dbus service-systemd
+service: service-dbus
+service-dbus:
 	@sed "s|##PREFIX##|$(PREFIX)|" org.knopwob.dunst.service.in > org.knopwob.dunst.service
+ifneq (0,${SYSTEMD})
+service: service-systemd
+service-systemd:
 	@sed "s|##PREFIX##|$(PREFIX)|" dunst.systemd.service.in > dunst.systemd.service
+endif
 
 .PHONY: clean clean-dunst clean-dunstify clean-doc clean-tests
 clean: clean-dunst clean-dunstify clean-doc clean-tests
@@ -101,7 +119,10 @@ clean-doc:
 clean-tests:
 	rm -f test/test test/*.o
 
-.PHONY: install install-dunst install-doc install-service uninstall
+.PHONY: install install-dunst install-doc \
+        install-service install-service-dbus install-service-systemd \
+        uninstall \
+        uninstall-service uninstall-service-dbus uninstall-service-systemd
 install: install-dunst install-doc install-service
 
 install-dunst: dunst doc
@@ -114,13 +135,26 @@ install-doc:
 	mkdir -p ${DESTDIR}${PREFIX}/share/dunst
 	install -m644 dunstrc ${DESTDIR}${PREFIX}/share/dunst
 
-install-service: service
+install-service: service install-service-dbus
+install-service-dbus:
 	install -Dm644 org.knopwob.dunst.service ${DESTDIR}${SERVICEDIR_DBUS}/org.knopwob.dunst.service
+ifneq (0,${SYSTEMD})
+install-service: install-service-systemd
+install-service-systemd:
 	install -Dm644 dunst.systemd.service ${DESTDIR}${SERVICEDIR_SYSTEMD}/dunst.service
+endif
 
-uninstall:
+uninstall: uninstall-service
 	rm -f ${DESTDIR}${PREFIX}/bin/dunst
 	rm -f ${DESTDIR}${MANPREFIX}/man1/dunst.1
-	rm -f ${DESTDIR}${SERVICEDIR_DBUS}/org.knopwob.dunst.service
-	rm -f ${DESTDIR}${SERVICEDIR_SYSTEMD}/dunst.service
 	rm -rf ${DESTDIR}${PREFIX}/share/dunst
+
+uninstall-service: uninstall-service-dbus
+uninstall-service-dbus:
+	rm -f ${DESTDIR}${SERVICEDIR_DBUS}/org.knopwob.dunst.service
+
+ifneq (0,${SYSTEMD})
+uninstall-service: uninstall-service-systemd
+uninstall-service-systemd:
+	rm -f ${DESTDIR}${SERVICEDIR_SYSTEMD}/dunst.service
+endif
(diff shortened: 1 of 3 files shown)


bebehei:
Shitty Ubuntu 14.04. I tested it in a container and I still don't know what the Problem is. It seems, that the version of make doesn't support the Shellstatus variable as expected.

bebehei:
Finally got this fixed. Turns out Ubuntu 14 has the old GNU make major version. When reading the docs, it turns out that the `SHELLSTATUS` variable just got introduced in the recent major.

Makefile
```diff
@@ -8,6 +8,18 @@ ifneq ($(wildcard ./.git/.),)
 VERSION := $(shell git describe --tags)
 endif
 
+SERVICEDIR_DBUS ?= $(shell pkg-config dbus-1 --variable=session_bus_services_dir)
```
tsipinakis: I'm not sure about this one, dbus dirs are standardized in `dbus-daemon(1)` and using this ensures that any custom prefix (e.g. `~/.local/share`) won't get the dbus service installed.

bebehei: You can set `SERVICEDIR_DBUS` by yourself and therefore you can specify your own folder.

Makefile
```diff
@@ -103,13 +115,12 @@ install-doc:
 	install -m644 dunstrc ${DESTDIR}${PREFIX}/share/dunst
 
 install-service: service
-	mkdir -p ${DESTDIR}${PREFIX}/share/dbus-1/services/
-	install -m644 org.knopwob.dunst.service ${DESTDIR}${PREFIX}/share/dbus-1/services
-	install -Dm644 dunst.systemd.service ${DESTDIR}${PREFIX}/lib/systemd/user/dunst.service
+	install -Dm644 org.knopwob.dunst.service ${DESTDIR}${SERVICEDIR_DBUS}/org.knopwob.dunst.service
```
tsipinakis: It's a fatal error if one of these install calls fail, the error condition should be handled both here and in `install-systemd` and also in `uninstall-{dbus,systemd}` since it currently stops in the middle of the installation (or uninstallation).

PS: Why is the install target called `install-service` but uninstall `uninstall-dbus`? weird.

bebehei: Naming issue: true, ack.

Permission issue: what!? It's common to start and not check. However it's handled as fatal (like every command in a Makefile).

It's not the makefile's job to check for succession before installation. Also it's (that'll sound hard) not our job to make error messages user friendly. Anyone who's capable to execute make should understand what a permission denied error is.

bebehei:
About that docs thing: I thought about this, too. Undocumented vars won't be ever good. Where should I put this? Can you tell me a reasonable place?

Makefile
tsipinakis: Consider this scenario:
```
$ PREFIX=install/ SERVICEDIR_DBUS=install/dbus make install
install -Dm755 dunst install//bin/dunst
install -Dm644 docs/dunst.1 install//share/man/man1/dunst.1
install -Dm644 dunstrc install//share/dunst/dunstrc
install -Dm644 org.knopwob.dunst.service install/dbus/org.knopwob.dunst.service
install -Dm644 dunst.systemd.service /usr/lib/systemd/user/dunst.service
install: cannot remove '/usr/lib/systemd/user/dunst.service': Permission denied
Makefile:136: recipe for target 'install-service-systemd' failed
make: *** [install-service-systemd] Error 1
```
The fact that this means that the installation was successful but couldn't install a very optional part is weird right off the bat.

```
$ find install/
install/
install/bin
install/bin/dunst
install/share
install/share/man
install/share/man/man1
install/share/man/man1/dunst.1
install/share/dunst
install/share/dunst/dunstrc
install/dbus
install/dbus/org.knopwob.dunst.service
```
All normal here

```
$ PREFIX=install/ SERVICEDIR_DBUS=install/dbus make uninstall
rm -f install/dbus/org.knopwob.dunst.service
rm -f /usr/lib/systemd/user/dunst.service
rm: cannot remove '/usr/lib/systemd/user/dunst.service': Permission denied
Makefile:150: recipe for target 'uninstall-systemd' failed
make: *** [uninstall-systemd] Error 1
```

Looks like something didn't run here and indeed uninstallation finished midway through
```
$ find install/
install/
install/bin
install/bin/dunst
install/share
install/share/man
install/share/man/man1
install/share/man/man1/dunst.1
install/share/dunst
install/share/dunst/dunstrc
install/dbus
```

The only reason the `install` command worked correctly is because `install-systemd` happened to be the last target to be run it'd have exited and might've skipped some steps otherwise.

It's common not to check permissions when we would fail anyway if we can't install that specific part but it's not what's happening here, this is completely optional and it's making it look like something fatal happened.

tsipinakis:
Best place I can think of is the only one we have installation instructions, aka the wiki but since with this PR local installation becomes a bit more complex I'd also add a link to the README pointing that way.

Makefile
bebehei: Sorry, but I have to fully disagree with you. Let me tell you, why:

Let's make an axiom test first.

- (1) In an unparametrized make call: systemd available <=> systemd service file installed,
- (2) If you have systemd installed, you want to get valuable information from `systemctl status --user dunst.service`

Please validate all your statements of your previous comment against those two axioms.

tsipinakis: I was going to argue the other way but then I realised it's the exact same situation with the dbus service which is pretty crucial. So I give it a pass here.

What I still don't like though is the uninstallation failing mid-way, the same way we use -f to ignore errors on non-existent files we should check and only issue a big warning if there is a permission issue with one of the files not stop completely and leave junk files behind.

bebehei: > What I still don't like though is the uninstallation failing mid-way,

Err... Failing the installation midway implies, that the installation did not succeed. So, you can't uninstall dunst successfully if your installation failed.

Packing all uninstall targets into a single one and in combination with some shell foo, this can get solved. This ignores all make principles, but it's the only possible solution. I've pushed a commit to this branch now. Let's discuss it.

Makefile
```diff
@@ -141,17 +140,12 @@ install-service-systemd:
...
-uninstall-service-systemd:
-	rm -f ${DESTDIR}${SERVICEDIR_SYSTEMD}/dunst.service
-endif
+uninstall:
+	( \
+		rm -f ${DESTDIR}${PREFIX}/bin/dunst || ec=1; \
+		rm -f ${DESTDIR}${MANPREFIX}/man1/dunst.1 || ec=1; \
+		rm -rf ${DESTDIR}${PREFIX}/share/dunst || ec=1; \
+		rm -f ${DESTDIR}${SERVICEDIR_DBUS}/org.knopwob.dunst.service || ec=1; \
+		if [ "0" != "${SYSTEMD}" ]; then rm -f ${DESTDIR}${SERVICEDIR_SYSTEMD}/dunst.service || ec=1; fi; \
```
tsipinakis: That wasn't exactly what I meant, more like a "|| echo "WARNING: Failed to remove the {"systemd service file", "dbus service file"}"

bebehei:
So, ok. Let's please remind, how manual installation and removal of software works. I'll come back later on this systemd/dbus stuff.

The software gets configured with `./configure --param1 --param2` and the software gets installed with `make install`. In meantime, the corresponding `Makefile` had been generated, which remembers the configuration, which resulted of the (parametrized) configure call.

To remove the software again, you have to either:

- change to the untouched compilation directory and run `make uninstall`
- If the compilation directory has been touched after installation, you mostly have to run configure with the same parameterset again and then run `make uninstall`
- download the software again, configure it with the same parameters and then run `make uninstall`

In any case, a successful removal can only be assured, when the last configure call in the compilation directory matches the configure call of the installation.

And here, in dunst's case, this is the same. The advantage of GNU autotools is, that you do not have to actively remember the configure parameters for calling `make uninstall`.

When we install dunst we have to give the same parameters for `make`, `(sudo) make install` and `(sudo) make uninstall`. This is a bit sad situation, but well it's how it is. We would have to change the buildsystem for this.

---

Back on the systemd file removal issue:

At this point now, given that make wants to remove the file when the `uninstall` target gets executed is simply, that the previous `make install` call with the same configuration has installed the file there.

So, the fact, that `make uninstall` wants to remove a file, where it does not have permissions, is either, that someone simply forgot sudo (when the person used sudo for `make install`) or has simply used another file location for this.

It makes no sense, treating some failed removals as fatal and some others not. If dunst fails to remove this file, it fails to remove this file, because you either did not specify the params correctly or have forgot sudo.

tsipinakis:
I agree with what you said but there's just one point demonstrated by the example above:
> At this point now, given that make wants to remove the file when the uninstall target gets executed is simply, that the previous make install call with the same configuration has installed the file there.

Or option 2 is that an installation was attempted (but failed because the user forgot to set the right variables, usual mistake given they aren't that common) and now they can't remove the already installed files because uninstallation tries to remove the systemd stuff first (and fails).

Sure, you can just sudo it or dig the Makefile to look for everything that's being installed and remove it manually but that misses the point of the uninstall target.

tsipinakis:
We've already delayed the release of 1.3 by _a lot_ and I've been arguing about a very rare scenario for a few days now so I'm approving this (I've removed the last experimental commit you pushed in response to my argument).

I also realised a lot of this could also easily be resolved with some better documentation so I'd also make it a goal to add some more detailed installation instructions to the repo, maybe in `README` or a separate file.

I've approved this but I'm leaving the merge up to you since I did remove a commit without your approval.

Issue #357, opened 2017-08-10
Xresource colors


tadly:
It's probably pretty obvious what I mean solely by the title.
To put it into a complete sentence though, I'd love to use/reference Xresource colors within the config.

I use [pywal](https://github.com/dylanaraps/pywal) to generate a color pallet and, for the time being, use a wrapper script to start dunst with all the colors applied.
This I'd like to get rid of.

hobarrera:
Dunst is not currently Xorg specific, so this would actually mean maintaining two configuration sources, with one being compile-time optional. 😕

Tadly:
*facepalms*
I even remember reading about dunst not being xorg specific in some other issues I'm following.
Totally forgot about that.

As a developer mysel I see how "ugly" this could/can get...
...as a user I want a central point to configure my color palette without hacking.

Let the brainstrom begin! So far... nothing.

Israel-:
Hi I don't usually comment on dunst issues, I just follow them because I use and really like dunst!
Are the Xresource colors the ones from something like:
/usr/share/X11/rgb.txt
And do you mean the color names?  This would make sense to support since these color names (though X11 specific are also used in the web as well).
Here is a function I use for that...
```c++
#include <X11/Xlib.h>
///etc...
std::string x_color_from_name(const char *colorName){
		Display *dpy;
		Colormap map;
		int scr;
		XColor rgb, nearest_rgb;
		dpy = XOpenDisplay(NULL);
		scr = XDefaultScreen(dpy);
		map = DefaultColormap(dpy, scr);
		XLookupColor(dpy, map, colorName, &rgb, &nearest_rgb);
		int red = (int)rgb.red>>8;
		int green = (int)rgb.green>>8;
		int blue = (int)rgb.blue>>8;
		char tmp[8];
		std::snprintf(tmp, sizeof(tmp), "#%02x%02x%02x", red, green, blue);
		std::string output = tmp;
		return output;
}
```

Israel-:
Just search for Xlibs in the makefile, and set some define and then wrap the function in that define
```c
#ifdef X_EXISTS
//function from above converted to c string
#endif
```

tsipinakis:
> And do you mean the color names? This would make sense to support since these color names (though X11 specific are also used in the web as well).

No, @Tadly is talking about a completely different thing entirely - having the ability to customize the colors from the .Xresources file, you're talking about X11 color value support.

[...]

PR #466, opened 2017-12-23
Interpret - as STDIN config file

bebehei:
As proposed in #357

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -78,7 +78,11 @@ void load_settings(char *cmdline_config_path)
         xdgInitHandle(&xdg);
 
         if (cmdline_config_path != NULL) {
-                config_file = fopen(cmdline_config_path, "r");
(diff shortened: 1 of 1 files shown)


tsipinakis:
While kinda a hack, isn't this already possible by giving `/dev/stdin` as the config file?

bebehei:
... so what's your resolution?

tsipinakis:
> ... so what's your resolution?

I'd say this is unnecessary since there's already a working solution that resolves this usecase so I'd close this but it's up to debate.

bebehei:
The only reason, why I want to add this, is because common unix tools interpret a single dash also as STDIN (e.g.: grep, cat, etc...). When I have to give stdin as a parameter I always want to check first, if the program is able to interpret `-`. I actually never fall back to `/dev/stdin`.

PR #468, opened 2018-01-05
Improve wording of DBus name lost error message

bebehei:
I had been a bit harsh in my IRC memo (dunst-project/dunst-project.org#8). I'm sorry for this.

There's a cosmetic change, which _could_ improve this error message.


tsipinakis:
I was thinking something along the lines of actually getting the name of the running daemon and indicating if dunst itself is already running.

It doesn't look like unix provides an easy way to get the process name from the PID without using `/proc` but since we're already sending a dbus message to get the PID we can use the `GetCapabilities` method to get the daemon name.

PR #471, opened 2018-01-05
Raise the maximum value of the progress hint

### LINKED ISSUE #437: progress values > 100
Currently, dunst doesn't show a progress value, [if it is larger than 100](https://github.com/dunst-project/dunst/blob/62e50289f0ead4cc0074b5ed7314302caf72688c/src/dbus.c#L265). It would be nice to change the behavior for two reasons:

- even with percentages, values larger than 100 are used in the wild. For example `pavucontrol` uses percentages > 100%
- the `%n` format specifier doesn't seem to be tied to a unit, so the restriction to 100 seems arbitrary

tsipinakis:
The progress hint is mostly used for progress bars but as discussed in
 #437 there are cases where values > 100 make sense.

Fixes #437

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -324,7 +324,7 @@ void notification_init(notification *n)
                 n->colors[ColFrame] = g_strdup(xctx.colors[ColFrame][n->urgency]);
 
         /* Sanitize misc hints */
-        if (n->progress < 0 || n->progress > 100)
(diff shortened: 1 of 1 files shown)


tsipinakis:
We're still using `[ %]` to print the progress so in a sense it is a percentage, it just isn't capped at 100.

PR #472, opened 2018-01-06
Implement hiding on fullscreen

### LINKED ISSUE #142: block notifications if there is a window in fullscreen mode
Hi 
Can you implement a feature that block notifications if there is a window in fullscreen mode

I wrote simple app to monitor if there is any
https://gist.github.com/aanatoly/9263530

bebehei:
Implements a rule to hide specific notifications on fullscreen

Fixes #142

### DIFF
--- a/src/notification.h
+++ b/src/notification.h
@@ -9,6 +9,12 @@
 
 #define DUNST_NOTIF_MAX_CHARS 5000
 
+enum behavior_fullscreen {
+        FS_NULL,  //!< Invalid value
+        FS_DELAY, //!< Delay the notification until leaving fullscreen mode
+        FS_SHOW,  //!< Show the message when in fullscreen mode
+};
+
 /// Representing the urgencies according to the notification spec
 enum urgency {
         URG_NONE = -1, /**< Urgency not set (invalid) */
@@ -69,6 +75,7 @@ typedef struct _notification {
         bool first_render;      /**< markup has been rendered before? */
         int dup_count;          /**< amount of duplicate notifications stacked onto this */
         int displayed_height;
+        enum behavior_fullscreen fullscreen; //!< The instruction what to do with it, when desktop enters fullscreen
 
         /* derived fields */
         char *msg;            /**< formatted message */
--- a/src/notification.h
+++ b/src/notification.h
@@ -101,5 +101,14 @@ void notification_update_text_to_render(notification *n);
 void notification_do_action(notification *n);
 
 const char *notification_urgency_to_string(enum urgency urgency);
+
+/**
+ * Return the string representation for fullscreen behavior
+ *
+ * @param in the #behavior_fullscreen enum value to represent
+ * @return the string representation for `in`
+ */
+const char *enum_to_string_fullscreen(enum behavior_fullscreen in);
+
 #endif
 /* vim: set tabstop=8 shiftwidth=8 expandtab textwidth=0: */
--- a/src/notification.h
+++ b/src/notification.h
@@ -10,9 +10,10 @@
 #define DUNST_NOTIF_MAX_CHARS 5000
 
 enum behavior_fullscreen {
-        FS_NULL,  //!< Invalid value
-        FS_DELAY, //!< Delay the notification until leaving fullscreen mode
-        FS_SHOW,  //!< Show the message when in fullscreen mode
+        FS_NULL,      //!< Invalid value
+        FS_DELAY,     //!< Delay the notification until leaving fullscreen mode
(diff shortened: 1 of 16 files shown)


bebehei:
@tsipinakis I've reworked it now, like we have discussed it in IRC. Only new notifications will get delayed and already drawn ones will still get displayed further.

This makes processing of timeouts easier. Previously, when switching between fullscreen and normal, timeouts started always from 0.

bebehei:
@tsipinakis I've also taken the best of two worlds. I added a field in the enum to optionally also push the notifications back.

bebehei:
I guess, I've hit a bug when processing `XEvent`s. Under some circumstances, some notifications get delayed until a new one or a focus change happens.

I'll investigate further, but I've got a hunch, that the last change in the dropgtk PRs about ignoring specific `PropertyChange` events is ignoring the event in too much cases.

bebehei:
Note for myself: When merging together with #475 and #491, it'll fail building. This needs a change in the for the `queue_*` methods, which get an additional parameter `fullscreen`.

bebehei:
> I guess, I've hit a bug when processing XEvents. Under some circumstances, some notifications get delayed until a new one or a focus change happens.

So, I finally found a reproducible event to test it and I had been able to debug and fix it.

While debugging, I also found another small issue with the handling of `xctx.cur_screen`. Further details in the commit.

@tsipinakis I'd kindly request your review.

---

Off-Topic: Something about coveralls: On [Travis Job 678.2](https://travis-ci.org/dunst-project/dunst/jobs/345681723), coveralls failed with a segfault. I have no clue, why this happened, but this is very spooky, we should investigate this.

tsipinakis:
> Off-Topic: Something about coveralls: On Travis Job 678.2, coveralls failed with a segfault. I have no clue, why this happened, but this is very spooky, we should investigate this.

This is not new it's been happening since we started using coveralls and I haven't been able to figure out why. In any case it doesn't affect much since the gcc build still reports results.

src/queues.c
```diff
@@ -331,25 +333,34 @@ void queues_update(void)
         }
 
         /* move notifications from queue to displayed */
-        while (waiting->length > 0) {
+        GList *iter = g_queue_peek_head_link(waiting);
+        while (iter) {
+                notification *n = iter->data;
+                GList *nextiter = iter->next;
```
tsipinakis: [nitpick] Is there a reason for having `nextiter` var rather than directly referring to `iter->next`?

src/x11/x.c
```diff
@@ -894,7 +894,11 @@ gboolean x_mainloop_fd_dispatch(GSource *source, GSourceFunc callback, gpointer
         unsigned int state;
         while (XPending(xctx.dpy) > 0) {
                 XNextEvent(xctx.dpy, &ev);
-                LOG_D("Processing event: %s\n", xevent_type_to_string(ev.type));
+                const char *typename = xevent_type_to_string(ev.type);
+                if (typename)
+                        LOG_D("XEvent: processing '%s'", typename);
```
tsipinakis: You can avoid having to use if logic for a simple debug log (which I dislike) by changing the statement to
```
LOG_D("XEvent: processing '%d: %s'", ev.type, typename);
```
and update `xevent_type_to_string` to return "Unknown" on an unkown event.

src/x11/x.c
```diff
@@ -69,6 +69,48 @@ static void setopacity(Window win, unsigned long opacity);
 static void x_handle_click(XEvent ev);
 static void x_win_setup(void);
 
+/* see x.h */
+const char *xevent_type_to_string(int type)
```
tsipinakis: I'm not sure about having this monster of a function just to log a debug message :/

src/queues.c
bebehei: Must be an older relict of the times, where I wanted to replace all queue iteration things with two pointers and a for loop.

src/x11/x.c
bebehei: We definitely can remove it. It was just super useful while debugging and I left it in on purpose. I wanted to know, what you think about it.

src/notification.h
```diff
@@ -9,6 +9,12 @@
 
 #define DUNST_NOTIF_MAX_CHARS 5000
 
+enum BehaviorFS {
```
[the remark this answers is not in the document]
bebehei: > I'd normally drop reviews like this as it's working perfectly and there are no relevant issues with the code but submitted anyway since @bebehei prodded me to.

This is exactly the reason, why I wanted to get a nitpicky review. One of the things, I dislike about dunst is the hard inconsistency of specific things (like typenames). Because even I fail for myself to stick to one schema.

Issue #412, opened 2017-10-15
Rethinking internal declaration and loading of config values


tsipinakis:
As referenced in #371 having setting parsing and documentation in `settings.c` and splitting default values into `config.h` as global variables is an ugly way to handle setting declaration, I propose the following alternative:

Create a unified `setting` struct that will hold all attributes of a setting, this way all settings can be specified in an array along with possible aliases and ways to handle backwards compatibility.

This might look similar to what already exists in `settings.c` but while it does lose some functionality since we can't add setting-specific code between parsing I would consider that an advantage since it standardizes a way to declare settings without having to go on a case-by-case basis. (Look at all the hacks for backwards compatibility that exist in there!)

Refer to the following data layout:

```c
struct setting {
      char *name; //config key
      char *section; //config section
      char *cmdopt, // cmdline option
      char *description; //description for help menu
      enum setting_type type; // Enum of all setting types - to know what to cast the void* to
      char *default_value;
      void *value;
      void (*parser)(char *cfg_value); // Function pointer for the parser - to be used in case of enums or other special settings
      /* TODO: Think of a way to signify aliases/identical settings 
          for backwards compatibility: i.e. `markup` and `allow_markup`
          maybe also for a way to specify a parser for other deprecated
          settings that affect this one */
}
```

## Storage of values

An idea that I thought about a lot was also migrating the value storage into the same data structure but after consideration I decided to keep the `settings` struct around and simply use a pointer to it in the declaration.

The main argument is that if the storage is done inside the array then each time we want to retrieve the value of a setting a function has to be called which would do a lookup, this has many disadvantages such as extra processing time for the look-up and that there is no convenient way to return a custom type in C so the result would have to be cast from a void pointer each time.

### Default value

Until now default values for settings were set in their proper type inside `config.h` as a global variable (updated to a struct with #371). Both of these are ugly solutions in their own way, ideally both the setting declaration and default value assignment would be together.

The solution that I came up with is to simply store the default value as a string as it would have been set in the config file and have a function call the parser for all settings when dunst starts up.

I've look into (and rejected) the following options

* Void pointer: Can't be easily set to the address of a literal without ugly C hacks
* Union: Possible if we go that route, but has the disadvantage that we have to repeat the type (i.e. `... TYPE_TIME, .time = 1234 ...` and slightly increased memory footprint along with having to add all enums into the union.
* In `settings`: Creates the problem we're trying to solve - setting attributes in different areas). 

## Declaration

[...]

PR #475, opened 2018-01-08
Proposal for settings refactoring

bebehei:
As already made clear in #412, our settings and configuration stuff has got some room to improve.

There are many different bad patterns in the code. But the problem is not the *bad*, the problem is the **many different**.

As proposed in #412, we want to have a big array of `Settings` structs at the end of the day. In my perception, I'm unsure if this can workout.

But anyways, we can prepare for that first. We can improve the code by similarizing it first. Then it'll get much easier to rethink the actual conf structures.

So this PR doesn't do the actual work for #412, it just makes preparations for it.

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -19,10 +19,67 @@
 #include "utils.h"
 #include "x11/x.h"
 
+#define S2P(val, def) string_to_path(val)
+#define S2S(val, def) g_strdup(val)
+
+#define TYPE(type, place, conv, free, equa) type
+#define PLAC(type, place, conv, free, equa) place
+#define CONV(type, place, conv, free, equa) conv
+#define FREE(type, place, conv, free, equa) free
+#define EQUA(type, place, conv, free, equa) equa
+
+#define EQ(a, b) (a != b)
+#define EQ_SEP(a, b) ((a.type == b.type) && (a.type != CUSTOM || 0 == strcmp(a.sep_color, b.sep_color)))
+
+//       typename      c-type                        placeholder               transferfunc            freefunc, equal
+#define  T_INT         int,                          "integer",                string_parse_int,       NULL,     EQ
+#define  T_BOOL        bool,                         "[yes|no]",               string_parse_bool,      NULL,     EQ
+#define  T_TIME        gint64,                       "integer[|s|m|h|d]",      string_to_time,         NULL,     EQ
+#define  T_PATH        char*,                        "path",                   S2P,                    g_free,   EQ
+#define  T_ALIGN       enum alignment,               "[left|center|right]",    parse_alignment,        NULL,     EQ
+#define  T_ICONPOS     enum icon_position_t,         "[left|right|off]",       parse_icon_position,    NULL,     EQ
+#define  T_STRING      char*,                        "string",                 S2S,                    g_free,   EQ
+#define  T_LOGLEVEL    GLogLevelFlags,               "level",                  string_parse_loglevel,  NULL,     EQ
+#define  T_ELLIPSIZE   enum ellipsize,               "[left|middle|right]",    parse_ellipsize,        NULL,     EQ
+#define  T_FOLLOW      enum follow_mode,             "[none|mouse|keyboard]",  parse_follow_mode,      NULL,     EQ
+#define  T_MARKUP      enum markup_mode,             "[no|strip|full]",        parse_markup_mode,      NULL,     EQ
+#define  T_SEPCOLOR    struct separator_color_data,  "<string>",               parse_sepcolor,         NULL,     EQ_SEP
+#define  T_URGENCY     enum urgency,                 "[low|normal|critical]",  parse_urgency,          NULL,     EQ
+#define  T_FULLSCREEN  enum behavior_fullscreen,     "",                       parse_enum_fullscreen,  NULL,     EQ
+
+#define OPTION(type, feld, def, sec, key, cmd, desc) \
+        do { \
+                const char *value = option_get_string(sec, key, cmd, NULL, desc); \
+                if (value) { \
+                        TYPE type  converted = CONV type (value, def); \
+                        if (!EQUA type(converted, def)) { \
+                                SETTINGS.feld = converted; \
+                        } \
+                } \
+        } while(0)
+
+#define DEPREC(type, feld, def, sec, key, cmd, desc, message) \
+        do { \
+                if (ini_is_set(sec, key) || cmdline_is_set(cmd)) { \
+                        OPTION(type, feld, def, sec, key, cmd, desc); \
+                        LOG_W(message); \
+                } \
+        } while(0)
+
+#define RULE(type, field, key) \
+                r->field = CONV type (ini_get_string(cur_section, key, NULL), r->field)
+
+char *string_parse_string(const char* string, const char *def)
+{
+        return string ? g_strdup(string) : g_strdup(def);
+}
+
 settings_t settings;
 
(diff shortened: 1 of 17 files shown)


src/settings.h
```diff
@@ -10,6 +10,7 @@ enum alignment { left, center, right };
 enum ellipsize { start, middle, end };
 enum icon_position_t { icons_left, icons_right, icons_off };
 enum separator_color { FOREGROUND, AUTO, FRAME, CUSTOM };
+struct separator_color_data { enum separator_color type; char *sep_color; };
```
tsipinakis: This hurts readability, structs should always be defined the long way (one item per line).

bebehei: Full ACK. Done. I missed up the enum with a struct. 🙈

bebehei:
@tsipinakis I'd be pleased if you go on and review this PR.

src/settings.c
```diff
@@ -349,26 +378,18 @@ void load_settings(const char *cmdline_config_path)
                 "Transparency. range 0-100"
         );
 
-        {
-                char *c = g_strdup(option_get_string(
+        settings.sep_color = parse_sepcolor(option_get_string(
```
tsipinakis: I don't think having duplicate `option_get_*` calls is a good idea, for once it duplicates the descriptions and general calls but also it should theoretically result in a duplicated entry in the help menu (couldn't make that bug appear in a quick test but I'm not sure why).

bebehei: Ah, thanks for this. That's a good point.

src/settings.c
```diff
@@ -20,32 +20,98 @@
...
+        else if (strcmp(string, "frame") == 0)
+                return FRAME;
+        else
+                return CUSTOM;
+}
+
+
+static enum ellipsize parse_ellipsize(const char *string, enum ellipsize def)
+{
+        if (!string)
```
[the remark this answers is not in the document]
bebehei: The `option_get_string` calls are all called with a default to `NULL`. So, if the value's not set, it returns `NULL`, which then should return the default.

I favor the GLib way of handling `NULL` values. (Interpreting `NULL` as a possible value of a string similar to the empty string, see g_strdup for example)

src/settings.c
```diff
@@ -388,11 +383,11 @@ void load_settings(char *cmdline_config_path)
                 "print notification on startup"
         );
 
-        settings.dmenu = string_to_path(option_get_string(
+        settings.dmenu = string_to_path(g_strdup(option_get_string(
```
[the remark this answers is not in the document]
bebehei: > In this call we're copying a string into a newly allocated space, passing it off to ....

In terms of memory allocation, there is no change. It's now at the point that it's visible that it copied the string one time too often.

> Perhaps making `string_to_path` take a const ...

This wasn't my main focus point yet of this PR. Shall I change it?

src/settings.c
```diff
@@ -11,7 +11,7 @@
 #endif
 
 #include "rules.h" // put before config.h to fix missing include
-#include "config.h"
+#include "../config.h"
```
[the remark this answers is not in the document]
bebehei: There is no point in it. My vim syntax and error checker mocks up, that it can' find `config.h` and the symbols included in it.

I had issues giving additional compiler flags (like `-I.`), so I used it as a quick soultion. I have to go forward and fix the actual issues. Sorry for the noise.

src/settings.c
bebehei: Err, see here: 32431bf1cba938710f05d646eb515f3fdbb7f19b (`Move separator color into single struct`), there it got removed and "fixed".

bebehei:
> Why specifically remove the get_time and get_path functions though? What about get_int and get_bool for which a similar case stands.

¯\_(ツ)_/¯

There's no big rationale in it. This came into my mind, too. But I discarded it as I thought, it might be too much work. When looking at it again now, there's a rationale, which could get derived: boolean, int and string are primitive datatypes, while `gint64` and "path" are not.

src/settings.c
tsipinakis: Whoops you'e right, missed that since I was reviewing each commit individually.

tsipinakis:
What bothers me about leaving those types in is that the `*_get_{int,bool,double}` are basically identical with the only difference being the wrapping or conversion functions that the returned value is being passed from, I'd argue we can get rid of them and stay with `option_get_string`.

2 more points I see worthwhile:

* With the changes this PR introduces we have 2 different default value parameters (one for the ini function, one for the parser). I'd assume the default value is still there because we can't define an `undefined` value for anything other than a string since we aren't returning a pointer. Leaving only the `get_string` function would allow us to get rid of that.

* (This is more against the previous point but still worth mentioning) The way these methods worked they used `cmdline_usage_append` to automatically add the type to the help menu with this changes these data types are now limited to `string`, `int`, `double` or `bool`.

bebehei:
> What bothers me about leaving those types in is that the ...

> With the changes this PR introduces we have 2 different default value parameters ...

That is a very good reasoning. Full ACK.

> I'd argue we can get rid of them and stay with option_get_string.

What about moving a step further? Using only `option(const char *ini, const char *cmdline, const char *def, const char *desc)` and when `cmdline` is `NULL`, it'll behave roughly like `ini_get_string` and when cmdline is NULL, it'll behave like `cmdline_get_string` and when both are set, it'll behave like `option_get_string`.

What's your opinion on this?

> The way these methods worked they used cmdline_usage_append to automatically add the type to the help menu with this changes these data types are now limited to string, int, double or bool.

We should have this fixed before merging the PR.

bebehei:
Ok, I removed the first option now. `option_get_bool`. `int` and `double` will follow.

Hahaha, the coverage in this PR finally drops more than 2 percent and coveralls marks it as bad (albeit it makes things better).

bebehei:
Now the edge cases are kicking in: `--help`/`--version` don't work anymore.

bebehei:
I've updated everything according to the recent merge of #491 and also have addressed the optimisation with `string_to_path`.

bebehei:
*dump*
```
@ rulestring @

identifier r;
identifier type;
identifier feld;
identifier sec;
expression key;
expression def;

@@

- r->feld = type(ini_get_string(sec, key, def));
+ OPTION(type, r->feld, def,
+        sec, key, NULL, NULL);

@ ruleconverter @

identifier r;
identifier type;
identifier feld;
identifier sec;
expression key;
expression def;

@@

- r->feld = type(ini_get_string(sec, key, NULL), def);
+ OPTION(type, r->feld, def,
+        sec, key, NULL, NULL);

@ settingsconverter @

identifier settings;
identifier feld;
expression type;
expression sec;
expression key;
expression cmd;
expression def1;
expression desc;
expression def;

@@

-settings.feld = type(option_get_string(sec, key, cmd, def1, desc), def);
+ OPTION(type, r->feld, def,
+        sec, key, cmd, desc);
```

bebehei:
So, the rebase to current master is done. I'm pretty sure, we can have something reviewable next week.

I'm changing the infrastructure to have the parser functions returning a `GError` instead of a stupid default value. With that we can achieve, that the actual return value is distinguishable if it's good or crap.

I had quite much time thinking about this and I'll probably go on with `gperf` and turn around the settings retrieval from the config-file/parameters. So that we don't walk over the settings struct and search the config value for every single settings-field. Instead we should walk over the config-file/argument list and evaluate the given parameter if it maps to a settings field. If not we're able then to spit out an error.

fwSmit:
Closing this, since I don't think this will be merged

Issue #477, opened 2018-01-09
Refactoring x11


bebehei:
For scheduled refactoring of the X11 code, what should we pay attention to?

- [add ability to center the notifications](https://github.com/dunst-project/dunst/pull/316)
- [transparent colors (rather than a general transparency)](https://github.com/dunst-project/dunst/issues/289)
- [transparent separator colors](https://github.com/dunst-project/dunst/issues/180)
    - (or maybe use a single window per notification!?)
- [Application name based geometry](https://github.com/dunst-project/dunst/issues/265)
---

@tsipinakis please add your thoughts here, too!

bondzula:
Single windows per notifications would be amazing, I would love to have different notifications size's based on the width needed for that notification, and not get resized when the bigger one comes along.

tsipinakis:
> add ability to center the notifications

I wouldn't put that in the X11 refactor that falls more into the settings redesign (#336 )

---

We should also look for:

* Separate X11 with the rest of the codebase as much as possible (_eyeing those `xctx` accesses in `notification.c`_).

* Decoupling X11 API calls from the pango/cairo drawing functions

bebehei:
> I wouldn't put that in the X11 refactor that falls more into the settings redesign (#336 ), the implementation part of it is pretty straight-forward.

Well, however we should look forward to refactor it in a way, that we can adapt such changes later very easy.

IMHO we should actually now decide, where we want to head for with the graphics stack. This avoids double work.

samirettali:
Implementing the transparency for the separators would be great!

bynect:
tracking this in #1018

PR #481, opened 2018-01-12
[X11refactor] Remove color field from xctx

bebehei:
From https://github.com/dunst-project/dunst/issues/477#issuecomment-357219608

> * Separate X11 with the rest of the codebase as much as possible (_eyeing those `xctx` accesses in `notification.c`_).

Looking close at the `xctx.colors` fields, these turn out to be duplicates and are not required. If handled in the settings lookup properly, there is no necessity at all to have the `colors` field and the coupling of `notification.c` and `x.c` can get removed.

### DIFF
--- a/src/notification.h
+++ b/src/notification.h
@@ -34,6 +34,12 @@ typedef struct _actions {
         gsize count;
 } Actions;
 
+typedef struct _notification_colors {
+        char *frame;
+        char *bg;
+        char *fg;
+} NotificationColors;
+
 typedef struct _notification {
         int id;
         char *dbus_client;
@@ -56,7 +62,7 @@ typedef struct _notification {
         enum markup_mode markup;
         const char *format;
         const char *script;
-        char *colors[3];
+        NotificationColors colors;
 
         /* Hints */
         bool transient;     /* timeout albeit user is idle */
--- a/src/notification.h
+++ b/src/notification.h
@@ -2,6 +2,13 @@
 #ifndef DUNST_NOTIFICATION_H
 #define DUNST_NOTIFICATION_H
 
+//TODO: Remove cyclic dependency
+typedef struct _notification_colors {
+        char *frame;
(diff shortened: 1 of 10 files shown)


src/notification.h
```diff
@@ -2,6 +2,13 @@
 #ifndef DUNST_NOTIFICATION_H
 #define DUNST_NOTIFICATION_H
 
+//TODO: Remove cyclic dependency
```
tsipinakis: I think the only thing we can do for the cyclical dependency is convert the `NotificationColor` refernces in settings.h to pointers and forward-define the typedef.

src/notification.h
```diff
@@ -2,6 +2,13 @@
...
+typedef struct _notification_colors {
+        char *frame;
+        char *bg;
+        char *fg;
+} NotificationColors;
```
tsipinakis: While `Actions` and `RawImage` are using CamelCase, I'd prefer if we shifted to camel_case for most things and slowly start shifting the codestyle that way. Though fixing everything right now is not a priority I'd prefer if we didn't introduce new stuff like this.

bebehei: This fits perfect into #483

src/notification.h
bebehei: Yes, this was my thought, too. But like the macro problem of `do { actual code } while (0)`, I think here is also a best solution applicable. And I think we should strive for consistency fix it to not tap into this again.

bebehei: Also, I cannot forward define the type, as I don't save the type as a pointer to a type's allocated memory. So, when defining `settings_t`, the size of `NotificationsColors` has to be known. 🤦‍

bebehei:
Let's put a "nocare" on this PR. This PR is similar in this case to the dropgtk one, where I started put work into the wrong effort. If anyone brings up work, which invalidates this PR, I'll happily close it.

Issue #456, opened 2017-12-13
Getting context hangs dunst


wpovell:
notify-send www.example.com
(Notif appears, ctrl-shift-period to get context, nothing happens)
notify-send test
(Now no notifications work at all)

Tried with browser set to both google chrome stable and firefox nightly.

### Installation info

dunst-git from AUR
OS: Linux arch 4.14.4
notify-send 0.7.7

bebehei:
Could you please run `killall dunst; dunst -print` on the command line, reproduce the stuff and paste the output of dunst?

Dunst should spit out some warnings here.

Also, what's your `dmenu_path` setting (and where does it point to)?

bebehei:
@wpovell Any news on this?

wpovell:
@bebehei Sorry for not getting back to you! My old laptop died so can't check the setup, but if I remember correctly the dmenu_path was `/usr/bin/rofi` and even when running with `-print` no errors were shown before hanging, just the standard message when a browser notification got sent (sorry I don't have it to share, can update again when I get a new laptop on the 3rd).

bebehei:
Ok, I can reproduce the problem. Much easier: `dunst -dmenu "sleep 20"`. This will simply make dunst halt for 20s. After 20s dunst then goes on and processes all incoming messages then responds again.

I guess we have to fork before dmenu and just to open the browser.

tsipinakis:
> I guess we have to fork before dmenu and just to open the browser.

We are forking for both dmenu and the browser but we wait for dmenus result (to issue the proper command) and to collect the exit status to avoid leaving zombie processes around.

To fix this we'd have to move all this logic into the main loop like `check_timeouts` etc.

bebehei:
Well, yeah. Probably we have to do this. As long as dmenu is opened, dunst can't do anything. Because most of the programs on the desktop rely on the notification daemon to be responsive, all other programs hang. Of course, when interactively using dmenu, the user doesn't see this and it's not a tragedy, if the user handles dmenu quickly. But if things go south, the user doesn't know what's going on and also does not see any error message.

I just want to remind you of your IRC client, when you tested something with dunst and your IRC connection timed out, because your client couldn't send notifications.

tsipinakis:
> My old laptop died so can't check the setup, but if I remember correctly the dmenu_path was /usr/bin/rofi

[...]

PR #482, opened 2018-01-13
Handle execvp error in open_browser

### LINKED ISSUE #476: Dunst crashes if browser variable is invalid
No matter what I set the `browser` variable to, Firefox is opened when I open a URL. 

I've tried `browser=xdg-open`, `browser=/usr/bin/xdg-open`, `browser=waterfox`, and `browser=/usr/bin/waterfox`. All of these are ignored.

If I uninstall Firefox, I get the error
```
XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":0"
      after 129 requests (129 known processed) with 0 events remaining.
XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":0"
      after 141 requests (141 known processed) with 0 events remaining.
```
and dunst crashes.

### Installation info

- Version: `1.3.0`
- Install type: `Package`
- Distro and version: `Arch Linux`

bebehei:
The execvp function usually doesn't return and exits by itself. But,
when having an error, it returns. As a faulty browser setting could
trigger this, dunst has to handle this, as the forked child wants to
access the X11 server too, resulting in XIO errors.

Fixes #476

### DIFF
--- a/src/menu.c
+++ b/src/menu.c
@@ -115,8 +115,7 @@ void open_browser(const char *in)
                 if (browser_pid2) {
                         exit(0);
                 } else {
-                        char *browser_cmd =
-                            string_append(settings.browser, url, " ");
+                        char *browser_cmd = g_strconcat(settings.browser, " ", url, NULL);
                         char **cmd = g_strsplit(browser_cmd, " ", 0);
                         execvp(cmd[0], cmd);
                         // execvp won't return if it's successful
--- a/src/menu.c
+++ b/src/menu.c
@@ -119,6 +119,12 @@ void open_browser(const char *in)
                             string_append(settings.browser, url, " ");
                         char **cmd = g_strsplit(browser_cmd, " ", 0);
                         execvp(cmd[0], cmd);
+                        // execvp won't return if it's successful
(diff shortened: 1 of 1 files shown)


src/menu.c
```diff
@@ -119,6 +119,12 @@ void open_browser(const char *in)
                             string_append(settings.browser, url, " ");
                         char **cmd = g_strsplit(browser_cmd, " ", 0);
                         execvp(cmd[0], cmd);
+                        // execvp won't return if it's successful
+                        // so, if we're here, it's definitely an error
+                        fprintf(stderr, "Warning: failed to execute '%s': %s\n",
+                                        settings.browser,
```
tsipinakis: `settings.browser` is being overwritten above by the `string_append` call and can potentially be empty or point to junk here. Perhaps printing `browser_cmd` is better.

bebehei: Albeit not outdated, I've handled this in the `string_append` lines above.

bebehei:
@tsipinakis I've also handled the initial bug for #456.

When the `dmenu` setting is faulty, dunst gets reluctant to signals and it can get only stopped via `SIGKILL`.

PR #488, opened 2018-01-29
Makefile: correct dependencies to fix race condition

stapelberg:
Before this commit, when building dunst in parallel, org.knopwob.dunst.service might get installed as an empty file. See https://bugs.debian.org/888760


bebehei:
@tsipinakis I think we should backport and release it as `1.3.1`, too.

tsipinakis:
Thanks for catching this!

I agree that this is a serious enough issue to warrant a bugfix release.

PR #491, opened 2018-02-20
Parsable markup documentation with doxygen

bebehei:
This is a start to have the documentation with doxygen, which also contains markup to be able to compile it at the end of the day.

With this, one's able to compile a document to get an overview of the internal structures of dunst.

I've moved my own comments of the `queues.c` and `log.c` files to be compliant.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,5 +1,19 @@
 # Important notes on the code
 
+**You can generate an internal overview with doxygen. For this, use `make doc-doxygen` and you'll find an internal overview of all functions and symbols in `docs/internal/html`.**
+
+# Comments
+
+- Comment system is held similar to JavaDoc
+    - Use `@param` to describe all input parameters
+    - Documentation comments should start with a double star (`/**`)
+    - Append `()` to function names and prepend variables with `#` to properly reference them in the docs
+- Add comments to all functions and methods
+- Markdown inside the comments is allowed and also desired
+- Add the comments to the prototype. Doxygen will merge the protoype and implementation documentation anyways.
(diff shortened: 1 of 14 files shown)


bebehei:
Enforcing  the tags with YCM can get very profitable. See the [GetDocs command](https://github.com/Valloric/YouCompleteMe#the-getdoc-subcommand).

HACKING.md
```diff
@@ -1,5 +1,20 @@
...
 
+**You can generate an internal overview with doxygen. For this, use `make doc-doxygen` and you'll find an internal overview of all functions and symbols in `docs/html`.**
+
+# Comments
+
+- Comment system is held similar to JavaDoc
+    - Use `@param` to describe all input parameters
+    - Documentation comments should start with a double star (`/**`)
+    - Look forward to properly link all member types
+        - Functions with an appended `()`
```
tsipinakis: These 2 lines needs some more explanation, I can't make much sense of them.

HACKING.md
```diff
@@ -1,5 +1,20 @@
...
+- Add comments to all functions and methods
+- Markdown inside the comments is allowed and also desired
+- Add the comments to the prototype. Doxygen will merge the protoype and implementation documentation anyways.
+  Except for **static** methods. Add the documentation header to the implementation and *not to the prototype*.
```
tsipinakis: That period id going to confuse people comma would be more appropriate here.

HACKING.md
bebehei: I've added an example and also the way to reference symbols/variables.

PR #496, opened 2018-03-05
[WIP] Drawing refactor

tsipinakis:
_Finally_ re-started work on the drawing refactor. There are still a lot of stuff to improve before calling this finished but the initial phase of moving everything to draw.c is done so submitting this to get some opinions and (hopefully) get this merged faster to avoid conflict hell.


bebehei:
From @tsipinakis via IRC
> Hey, can you take a look at the updated geometry handling in the drawing refactor? Not 100% on some of the decisions I made there

I've got an exam at Friday. I won't review until then. So push/rebase your fixup commits and work on. I'll review on the weekend.

src/x11/x.c
```diff
@@ -53,35 +53,44 @@ cairo_surface_t *x_create_cairo_surface(void)
 
 void x_win_move(int width, int height)
 {
+        // Previous dimensions of the window to avoid calling X11 if no change
+        // is needed
+        static struct {
```
bebehei: Why not `static struct dimension window_dim = { 0 }`?

tsipinakis: `struct dimension` is only visible in `draw.c`. I'm trying to avoid having a lot of gobally available structs. I'll refactor this part more in subsequent commits anyway, I plan to create our own window data structure to allow handling of multiple windows in the future.

config.h
```diff
@@ -16,7 +16,7 @@ settings_t defaults = {
 .icons = { "dialog-information", "dialog-information", "dialog-warning" }, /* low, normal, critical */
 
 .transparency = 0,           /* transparency */
-.geom = "0x0",               /* geometry */
+.geometry = { 0 },           /* geometry */
```
bebehei: IMHO, we shouldn't just set it to `0` to successfully null it in the settings. IMHO the settings should get receive an elaborated initialisation.

src/draw.c
```diff
@@ -554,4 +554,12 @@ void draw(void)
         cairo_surface_destroy(image_surface);
         r_free_layouts(layouts);
 }
+
+void draw_free(void)
```
bebehei: rebase?

src/x11/x.c
```diff
@@ -725,7 +731,7 @@ gboolean x_mainloop_fd_dispatch(GSource *source, GSourceFunc callback, gpointer
                 switch (ev.type) {
                 case Expose:
                         if (ev.xexpose.count == 0 && xctx.visible) {
-                                x_win_draw();
+                                draw();
```
bebehei:
```
src/x11/x.c:734:33: warning: implicit declaration of function ‘draw’; did you mean ‘drem’? [-Wimplicit-function-declaration]
```

Obviously the `draw.h` header is missing here.

bebehei:
I've created tests for the gdk pixbuf functions and refactored them to make clear when `_from_file` and when the former `_from_path` should get used: https://github.com/dunst-project/dunst/compare/refactor-drawing...bebehei:icons

I'd appreciate if you would add these commits to the PR.

config.h
bebehei: That's correct. Yes. But my intention was to split it off for better readability.

tsipinakis:
@bebehei I'm going to call it a wrap here for this PR as it's got way too big and almost unreviewable. There is still a lot of cleanup to be done (draw.c still depends on X stuff) but it's best to avoid merging huge chunks to master at once IMO.

src/draw.c
```diff
@@ -380,6 +380,39 @@ static void free_layouts(GSList *layouts)
         g_slist_free_full(layouts, free_colored_layout);
 }
 
+static cairo_surface_t *render_backdrop(cairo_surface_t *srf, colored_layout *cl, colored_layout *cl_next, int y, int width, int height, bool first, bool last)
```
bebehei: Well, this is a quite long line. I guess, breaking it up should improve readability.

src/draw.c
```diff
@@ -380,6 +380,39 @@ static void free_layouts(GSList *layouts)
...
+                height -= settings.separator_height;
+
+        cairo_set_source_rgb(c, cl->bg.r, cl->bg.g, cl->bg.b);
+        cairo_rectangle(c, x, y, width, height);
+        cairo_fill(c);
+
+        return cairo_surface_create_for_rectangle(srf, x, y, width, height);
+}
+
 static struct dimensions render_layout(cairo_surface_t *srf, colored_layout *cl, colored_layout *cl_next, struct dimensions dim, bool first, bool last)
```
bebehei: As above, this line is quite long. I'd love to break it up, too.

src/draw.c
bebehei: What about naming the function to `render_background`? Personally, I heard "backdrop" the first time today, that I really had to use a translator and I guess it's quite uncommon.

src/draw.c
bebehei: Oh, I have to clarify myself: I would use `*_free(...);` only for structures, which are dynamic elements. I would rename it to `_deinit(...`, because it deallocates only static elements.

bebehei:
> but I guess, I'll merge it right away into my own dunst instance.

.... aaaand I've found a corner case.

I'm locking my screen with [`i3lock`](https://github.com/i3/i3lock/). And before locking my screen, I send a notification to dunst. Previously, this worked fine. Notifications weren't visible. But now dunst shows up above the lockscreen.

bebehei:
this branch is leaking "some" memory. :rofl: After ~11 hours (>6 of them in suspend), dunst consumes about 4GB of RAM.

config.h
```diff
@@ -16,7 +16,15 @@ settings_t defaults = {
...
 .transparency = 0,           /* transparency */
-.geom = "0x0",               /* geometry */
+.geometry = { .x = 0,        /* geometry */
+              .y = 0,
+              .w = 0,
+              .h = 0,
+              .negative_x = 0,
+              .negative_y = 0,
+              .negative_width = 0,
+              .width_set = 0
```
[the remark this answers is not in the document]
tsipinakis: It's required to determine the difference between width being 0 and dynamic width (see `calculate_dimensions`)

bebehei:
Answer to https://github.com/dunst-project/dunst/pull/496#issuecomment-381760930

I've drawn the [status quo](https://github.com/dunst-project/dunst/tree/f8538c9a459ec8274b0b14cd3055ac8ba10d6f5c) in an automaton diagram how dunst's window functions work:

dunst_x11win

IMHO @knopwob has given very clear function names here. Before you draw the window, you actually have to show it. Then you draw into it arbitary many times and then you hide it again.

In your current code, it's [effectively this semaphore check](https://github.com/dunst-project/dunst/pull/496/commits/b0643a23b220003dd6304781a37aec8f97935abb#diff-1878180e15c61cd8ac68995ebf19a28fL498) executing this [`XMapRaised` call over and over](https://github.com/dunst-project/dunst/pull/496/commits/b0643a23b220003dd6304781a37aec8f97935abb#diff-1878180e15c61cd8ac68995ebf19a28fR517) again.

While `x_win_show` contains an [`XMapRaised`](https://tronche.com/gui/x/xlib/window/XMapRaised.html) call and you've removed the semaphore with the visible bool, you start calling `XMapRaised` over and over.

On the other hand, `i3lock` actually listens for `VisibilityNotify` event and usually remaps. I don't get why `i3lock` still can't circumvent it.

tsipinakis:
> Before you draw the window, you actually have to show it. Then you draw into it arbitary many times and then you hide it again.

You're actually right, I dropped that commit since when looking at it again it was based on some false assumptions.

src/x11/x.c
```diff
@@ -485,7 +490,7 @@ static void x_win_setup(void)
                                    (0xffffffff / 100)));
 
         if (settings.f_mode != FOLLOW_NONE) {
-                long root_event_mask = FocusChangeMask | PropertyChangeMask;
+                long root_event_mask = FocusChangeMask | PropertyChangeMask | SubstructureNotifyMask;
```
bebehei: This will introduce a stupid bug. We have to listen to the `SubstructureNotifyMask` anyways, not conditioned to the follow mode.

src/draw.c
```diff
@@ -391,7 +391,11 @@ static int layout_get_height(colored_layout *cl)
         return MAX(h, h_icon);
 }
 
-static cairo_surface_t *render_background(cairo_surface_t *srf, colored_layout *cl, colored_layout *cl_next, int y, int width, int height, bool first, bool last, int *ret_width)
+static cairo_surface_t *render_background(cairo_surface_t *srf,
+                                          colored_layout *cl, colored_layout *cl_next,
```
bebehei: I have to out myself: I'd rather take the overlong line, than a two-dimensional parameter-list. In readability this is actually a disservice.

IMHO think the gnome guys do it best [(example)](https://git.gnome.org/browse/libnotify/tree/libnotify/notification.c?h=0.7.7#n115).

src/draw.c
```diff
@@ -433,6 +435,9 @@ static cairo_surface_t *render_background(cairo_surface_t *srf, colored_layout *
         }
 
         cairo_destroy(c);
+
+        *ret_width = width;
```
bebehei: possible `NULL` deref?

Albeit the function is not intended to get used without a return value of the width, it might be in future and save some debugging ;-)

src/x11/x.c
tsipinakis: Good catch, fixed!

src/draw.c
tsipinakis: I think it takes a bit too much space now but I agree it is more readable that way.

src/icon.c
```diff
@@ -63,8 +63,10 @@ static GdkPixbuf *get_pixbuf_from_file(const char *filename)
         if (is_readable_file(filename)) {
                 GError *error = NULL;
                 pixbuf = gdk_pixbuf_new_from_file(filename, &error);
-                if (!pixbuf)
+                if (!pixbuf) {
+                        LOG_M("%s", error->message);
```
bebehei: - Please use the warning log level. The error is correctable, but it's only correctable by ignoring the data.
- Please do not pipe the error message directly into the log. Give it some context. If `gdk-pixbuf` has got complete control over the message and it would give a generic error message like `Permission Denied`, you'd have only `Permission Denied`, but the user wouldn't know why and where the permission got denied. I'd suggest `LOG_W("Cannot load icon '%s' : %s", filename, error->message);`

src/icon.c
```diff
@@ -115,8 +117,6 @@ GdkPixbuf *get_pixbuf_from_icon(const char *iconname)
                         start = end + 1;
                 } while (*(end) != '\0');
         }
-        if (!pixbuf)
-                LOG_W("Could not load icon: '%s'", iconname);
```
bebehei: This error message is still important. If there is no icon file named as `iconname` in your `settings.icon_path`, then still no icon would appear, but also no message would get logged.

I would refine the message and use `LOG_W("No icon found for: '%s'", iconname);`. Then it would make sense together with the message from `get_pixbuf_from_file`.

tsipinakis:
Agreed, I've pushed some remaining fixes that I had in my local repo and will leave this here for now. Feel free to review it.

src/icon.c
```diff
@@ -0,0 +1,183 @@
...
+
+        int w = gdk_pixbuf_get_width(pixbuf);
+        int h = gdk_pixbuf_get_height(pixbuf);
+        int larger = w > h ? w : h;
+        if (settings.max_icon_size && larger > settings.max_icon_size) {
+                GdkPixbuf *scaled;
+                if (w >= h) {
+                        scaled = gdk_pixbuf_scale_simple(pixbuf,
+                                        settings.max_icon_size,
+                                        (int) ((double) settings.max_icon_size / w * h),
```
bebehei: We're parsing an integer to a `double`, to divide it by an `int` and to multiply it via an `int`.

Wouldn't it be more sophisticated, to **multiply** the `int` first and then divide it afterwards. doesn't need any casts and solely consists of integer arithmetic.

-> same for else condition below.

src/settings.c
```diff
@@ -256,11 +257,24 @@ void load_settings(char *cmdline_config_path)
...
-        );
+        {
+
+                char *c = option_get_string(
+                        "global",
+                        "geometry", "-geom/-geometry", NULL,
+                        "Geometry for the window"
+                );
+
+                if (c == NULL) {
```
bebehei:
```
+               if (c) {
+                        // TODO: Implement own geometry parsing to get rid of
+                        //       the include dependency on X11
+                        settings.geometry = x_parse_geometry(c);
+                        g_free(c);
+                } else {
+                        settings.geometry = defaults.geometry;
+                }
```

src/x11/x.c
```diff
@@ -1124,22 +465,41 @@ static void x_set_wm(Window win)
...
+        static GSourceFuncs x_src_fn = {
+                x_mainloop_fd_prepare,
+                x_mainloop_fd_check,
+                x_mainloop_fd_dispatch,
+                NULL,
+                NULL,
+                NULL
+        };
+
+        struct x11_source *x_src = (struct x11_source*) g_source_new(&x_src_fn,
```
bebehei: err, at least from curiosity: How is this supposed to work? The structure pointer, which gets spit out by `g_source_new` is [opaque](https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#GSource). This may work, but I actually think this is a very bad hack [like it's done in `dunstify`](https://github.com/dunst-project/dunst/blob/7a9187728c8fbd519358e56e3df55a9639025f15/dunstify.c#L178-L188).

src/icon.c
```diff
@@ -0,0 +1,183 @@
...
+        return icon_surface;
+}
+
+static GdkPixbuf *get_pixbuf_from_file(const char *filename)
+{
+        GdkPixbuf *pixbuf = NULL;
+        if (is_readable_file(filename)) {
+                GError *error = NULL;
+                pixbuf = gdk_pixbuf_new_from_file(filename, &error);
+                if (!pixbuf) {
```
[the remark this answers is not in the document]
tsipinakis: From the [docs](https://developer.gnome.org/gdk-pixbuf/stable/gdk-pixbuf-File-Loading.html#gdk-pixbuf-new-from-file)

> Returns A newly-created pixbuf with a reference count of 1, or NULL if any of several error conditions occurred

So checking either one should be fine, I don't see a reason to change it.

src/draw.c
```diff
@@ -0,0 +1,575 @@
...
+                        ellipsize = PANGO_ELLIPSIZE_START;
+                        break;
+                case middle:
+                        ellipsize = PANGO_ELLIPSIZE_MIDDLE;
+                        break;
+                case end:
+                        ellipsize = PANGO_ELLIPSIZE_END;
+                        break;
+                default:
+                        assert(false);
```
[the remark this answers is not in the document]
tsipinakis: I actually haven't touched this piece of code yet, just copy-pasted it. I'll hold onto this for the next PR when I refactor the layout creation.

tsipinakis:
I've pushed the relevant fixes. I'm not sure if you prefer me to rebase after every review or if that messes with your current progress on it so I just left them as fixups.

Looking forward for the full review!

src/x11/x.c
```diff
@@ -45,6 +45,10 @@ bool dunst_grab_errored = false;
 static bool fullscreen_last = false;
 
 /* FIXME refactor setup teardown handlers into one setup and one teardown */
+
+static void x_shortcut_init(keyboard_shortcut *ks);
```
bebehei: The comment above is lost.

src/icon.c
```diff
@@ -163,11 +163,11 @@ cairo_surface_t *icon_get_for_notification(const notification *n)
                 if (w >= h) {
                         scaled = gdk_pixbuf_scale_simple(pixbuf,
                                         settings.max_icon_size,
-                                        (int) ((double) settings.max_icon_size / w * h),
```
bebehei: You're changing logic here and not just copying the code from file a to b. IMHO I'd still separate it into a single commit.

bebehei:
> I've pushed some other small cleanups onto the

I've pushed the commits right now (and not right with my review).

src/x11/x.c
tsipinakis: > But why are we casting it then again to a GSource* back?!?

We could leave it as-is but we're casting it back because all glib methods need the GSource and not the derived struct so it's just easier that way.

[Apparently this is a really weird way to do inheritance in C.](https://stackoverflow.com/questions/1114349/struct-inheritance-in-c)

src/draw.h
```diff
@@ -0,0 +1,14 @@
+#ifndef DUNST_DRAW_H
+#define DUNST_DRAW_H
+
+#include "src/x11/x.h"
+extern window_x11 *win; // Temporary
```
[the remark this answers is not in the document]
tsipinakis: To remove this we still need to figure out how to handle the `visible` field from within `draw.c` so it has to stay for now.

src/x11/x.h
```diff
@@ -23,13 +23,15 @@ typedef struct _keyboard_shortcut {
         bool is_valid;
 } keyboard_shortcut;
 
+// Cyclical dependency
```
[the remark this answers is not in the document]
tsipinakis: For this specific case, the way to handle this would be to move `keyboard_shortcut` to `settings`, but that will have to wait for the next PR when X11 is fully decoupled from the rest of the code to allow for other outputs.

src/icon.c
tsipinakis: You're right, I changed it to add it as a fixup to the the commit that removed some of the previous floating-point hackery that existed in `draw.c`.

tsipinakis:
Rebase done, waiting for the final (and hopefully approving) review :)

PR #502, opened 2018-03-15
Coveragereport and test suite improvements

bebehei:
Here are some improvements about the buildsystem. Especially some improvements for the coverage report.

The coveralls service is a great webservice, but like for Travis it's better to have an independent testsuite which makes it possible to process it locally, which makes it faster and doesn't even require a full push iteration. So there is a gcovr support added.


src/dunst.c
```diff
@@ -23,7 +23,7 @@
 #include "x11/x.h"
 
 #ifndef VERSION
-#define VERSION "version info needed"
+#error "no version info given"
```
tsipinakis: If we're going to fail in this case the entire `ifndef` block is redudant, the compiler will fail because of the undefined VERSION macro when it's used.

test/misc.c
```diff
@@ -0,0 +1,25 @@
+#include "greatest.h"
+
+#ifndef VERSION
+#define VERSION ""
```
tsipinakis: Why cause a fatal error in `dunst.c` if VERSION is not set but gracefully handle it here?

bebehei:
> I'm not sure if I've mentioned this before back when we talked about adding coveralls to the repo but my opinion is that coverage reports should only be used as a ballpark "this part is not tested, lets make it our next target" marker, not as a test suite grader for the reason mentioned above. Adding the ability to generate coverage reports locally in my opinion wrongly enforces the mindset of micro-managing the coverage % and paying less attention to writing proper tests that account for all corner cases.

On the one hand side, I agree with you. I don't want to make this a necessary tool either. But on the other hand, we use this as a service and I don't want to rely on it.

My intent is to get independent of this service.

tsipinakis:
> My intent is to get independent of this service.

As far as service dependency goes, coveralls is undoubtedly the easiest one to replace if it ever goes down, it doesn't need any past data/there is no lock-in and there are tons of other coverage services and programs out there. I am not sure this is worth investment of effort and increased complexity in the makefile this PR adds.

.travis.yml
```diff
@@ -18,6 +18,9 @@ dist: trusty
 sudo: false
 language: c
 
+git:
+  depth: false
```
tsipinakis: What's the point of this? From the travis docs `depth: false` removes the `--depth` argument. Why do a full clone when we're not using it?

bebehei: We have to use it, otherwise the VERSION macro wouldn't work, as `git --describe` would fail otherwise.

src/utils.h
```diff
@@ -16,6 +16,9 @@
 //! Test if string a and b are the same case-insensitively
 #define STR_CASEQ(a, b) (strcasecmp(a, b) == 0)
 
+//! Convert a second into the internal time representation
+#define SECS(s) (((gint64)s) * 1000 * 1000)
```
tsipinakis: Why not use a function?

```c
gint64 seconds2us(gint64 seconds) {
       // do some overflow detection as a bonus
       return seconds * 1000 * 1000;
}
```

If performance is an issue it can also go in the header as `static inline`.

bebehei: Ok.

```
       // do some overflow detection as a bonus
```

Explain me, where do you do overflow detection here?

tsipinakis: I don't in that example! It was a suggestion as an upside to using a function. Though I'm not sure the chances of a 64 bit int overflowing here.

At least the upside for me is that it saves the macro parenthesis madness.

bebehei: Well, I'd guess even just making the `gint64 seconds` const will do the job like compilers would optimize `strlen("asdf")` to its output.

It think it's a perfect opportunity to test the optimisation levels and look into its binary output.

bebehei:
> From 2018-03-15:
> Testing is an important part of development to ensure that future changes don't break existing functionality. This does not hold true for coverage reports, high coverage is not synonymous with a good test suite since it only counts that something actually ran, not that all code paths are accounted for.

---

> My opinion about gcovr hasn't changed since my last comment. I don't see any upside to adding it as a make target.
> I'd drop 8a87827 and 6ea38f1

Ok, I've worked now extensively with gcovr on my own branch for #551. Here's my experience:

I found a bug in existing code. Just because the big red block triggered me: In `notification_format_message` hadn't been any line covered. It was a low hanging fruit for me to add all format symbols in the notification message. **And dunst crashed.**

I found another bug in my tests while watching the coverage. You see what actually got executed. It's a bit like a debugger, but just 100 times faster.

And the coverage also showed visually, which blocks had no testing at all. It did not need to search my editor, which blocks needed testing. I just immediately saw it.

I fully agree, that this doesn't fix corner cases, but it totally helps for basic testing. And yes, it gives you the false belief, that your program is tested good enough, when the green light of your coverage app starts flashing.

But all said above holds true for any coverage application. You'd get the same tricks and treats with coveralls, too.

But why do I want to have gcovr in it:

- You get immediate feedback. No git commit, push, travis build and coveralls processing required. It's a few seconds vs a few minutes (factor 60!).
- You're immune coveralls sufferings. Exactly this PR: [commitA](https://github.com/bebehei/dunst/commit/94f9967d77b89910b7bf165b5aee6f3010ef6e5b) and [build for commitA](https://coveralls.io/builds/20103095) compare it with: [commitA'](https://github.com/bebehei/dunst/commit/ae47fa7bfe30d2dd7392c5c19c01a3a0e77b5163) and [build for commitA'](https://coveralls.io/builds/20096878). Both only differ in commit messages and some names. But 5% coverage diff. Why? I don't know, but honestly would like to know.

src/utils.h
bebehei: Ok, I researched it. Starting with `-O2` it gets optimized. Fine. As expected. Changed it.

Compiled:

```
src/../config.h:15:15: error: initializer element is not constant
 .timeouts = { s2us(10), s2us(10), s2us(0) }, /* low, normal, critical */
               ^~~~
```

:see_no_evil: 

Well, I guess, there is no discussion necessary, if we want a function, right? :grin:

PR #504, opened 2018-03-20
Count monotonic time with Linux specific clocks

### LINKED ISSUE #492: show_age_threshold does not handle suspend/resume
When using `show_age_threshold` the time delta seems to not include time where the computer was suspended.

Relevant code:
https://github.com/dunst-project/dunst/blob/8735c9d3d00faad206eb66d9aa93f7b77787f993/src/notification.c#L514-L515

From `GLib:Date_and_Time_Functions(3)`:
```
   g_get_monotonic_time ()
           gint64              g_get_monotonic_time                (void);

       Queries the system monotonic time, if available.

       On POSIX systems with clock_gettime() and CLOCK_MONOTONIC this call is a very
       shallow wrapper for that. Otherwise, we make a best effort that probably involves
       returning the wall clock time (with at least microsecond accuracy, subject to the
       limitations of the OS kernel).

       It's important to note that POSIX CLOCK_MONOTONIC does not count time spent while
       the machine is suspended.
```

Side note: it would also be good to skip calculating `t_delta` altogether if `show_age_threshold` is not used.

- Version: 1.3.1 (2018-01-30)
- Distro and version: Arch Linux

bebehei:
On Linux, CLOCK_MONOTONIC has got a bug. It does not count
onwards during sleep, albeit required by POSIX. This behavior
is reasoned with the requirement for poll().
Also the GLib people are sticking to this behavior in their
g_get_monotonic_time() function.

So we have to use a drop in replacement, which respects CLOCK_BOOTTIME
on Linux, as this is the clock, what would be CLOCK_MONOTONIC on POSIX
systems.

Fixes #492

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -7,6 +7,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 
 #include "log.h"
 
@@ -166,4 +167,21 @@ gint64 string_to_time(const char *string)
                 return 0;
 }
 
+/* see utils.h */
+gint64 time_monotonic_now(void)
+{
+        struct timespec tv_now;
+
+        /* On Linux, BOOTTIME is the correct monotonic time,
+         * as BOOTTIME counts onwards during sleep. For all other
+         * POSIX compliant OSes, MONOTONIC should also count onwards
+         * during system sleep. */
+#ifdef __linux__
+        clock_gettime(CLOCK_BOOTTIME, &tv_now);
+#else
+        clock_gettime(CLOCK_MONOTONIC, &tv_now);
+#endif
+        return (gint64)tv_now.tv_sec  * G_USEC_PER_SEC
+                     + tv_now.tv_nsec / 1000;
+}
 /* vim: set tabstop=8 shiftwidth=8 expandtab textwidth=0: */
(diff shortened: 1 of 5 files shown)


src/utils.c
```diff
@@ -166,4 +167,24 @@ gint64 string_to_time(const char *string)
...
+        /* On Linux, BOOTTIME is the correct monotonic time,
+         * as BOOTTIME counts onwards during sleep. For all other
+         * POSIX compliant OSes, MONOTONIC should also count onwards
+         * during system sleep. */
+#if __linux__
+        clock_gettime(CLOCK_BOOTTIME, &tv_now);
+#else
+        clock_gettime(CLOCK_MONOTONIC, &tv_now);
+#endif
+        i_now = tv_now.tv_sec * G_USEC_PER_SEC
```
tsipinakis: I have a feeling this multiplication can overflow since `tv_sec` can easily be 32 bits, we should probably cast it to `gint64` first.

bebehei: Me stupid. Why does C not automatically cast to its storage value :angry: ?

tsipinakis: It might I'm not sure, googling around didn't help either. Still, better safe than sorry as it goes.

PR #517, opened 2018-06-11
Fix typos

bebehei:
First commit broken out of the settings refactoring.


tsipinakis:
Feel free to push changes like this directly to master, I don't think a review is necessary.

You should be able to if I haven't messed up something in the protected branch settings, only force pushes should be disallowed.

bebehei:
> Feel free to push changes like this directly to master,

Agreed.

PR #521, opened 2018-06-19
Implement paths for icons

### LINKED ISSUE #520: Path in configs to resolve ( ~ )
Hi

As I have no idea how the notifiication determines what icon to use, so I would like to managed them in my own folder that sits in my home dir. To make my config universal, the expansion of tilde `~` would be awesome as I would not have to change the config per user. 

Resolving a $HOME variable would also be acceptable.
 
I would like to have  this feature, right now I have to set full path.

- Version: `1.3.2 (2018-05-06)`
- Install type: `package`
- Distro and version: ` 4.17.0-2-MANJARO latest stable `

bebehei:
This PR implements tilde support for `icon_path` and absolute icon paths.

Albeit `icon.c` checked the icon, if started with a `~`, the check never had any effect :joy: 

Fixes #520

### DIFF
--- a/test/icon.c
+++ b/test/icon.c
@@ -5,7 +5,7 @@
 #include <gdk-pixbuf/gdk-pixbuf.h>
 #include <glib.h>
 
-#define ICONPREFIX "./data/icons/path"
+#define ICONPREFIX "/data/icons/path"
 
 /* As there are no hints to test if the loaded GdkPixbuf is
  * read from a PNG or an SVG file, the sample icons in the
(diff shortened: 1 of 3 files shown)


test/icon.c
```diff
@@ -14,6 +14,55 @@
...
+{
+        g_assert(chars <= strnlen(string, chars));
+        return string+chars;
+}
+
+TEST test_get_pixbuf_from_file_tilde(void)
+{
+        char *cwd  = g_get_current_dir();
+        const char *home = g_get_home_dir();
+        const char *iconpath = skipchars(ICONPREFIX, 1);
```
tsipinakis: Maybe instead of having this function to hack around literal strings set `ICONPREFIX` to `data/icons/path` and explicitly add the current directory prefix in the suite declaration.

bebehei: Valid point. Changed.

While writing the tests, I wasn't 100% happy about the code either. Thank you for those "nitpicks" I love them and I love to work with you together.

PR #525, opened 2018-06-28
Added dunst xresources theme wrapper

lulivi:
Hello. 

I've been reading some discussions about allowing Command Substitution `$(...)` and X resources access from the configuration file (#357).

Then, I headed to the dunst channel in freenode and `nikos` came up with the idea of creating a script to solve this problem.

So I created a wrapper to change colors and some other attributes according to the Xresources theme. Any tips or improvements?

### DIFF
--- a/contrib/dunst_xr_theme_changer.sh
+++ b/contrib/dunst_xr_theme_changer.sh
@@ -1,0 +1,255 @@
+#!/usr/bin/env bash
+###############################################################################
+##
+## Usage
+##
+##     ./<script_name> [<OPTIONS>]
+##
+##     If it does not run, give execute permissions to the script with
+##     chmod +x <script_name>. Then run ./<script_name>.
+##
+## Options
+##
+##     -h|--help            Optional. Show help message.
+##
+## Description
+##
+##     This script creates a dunst themed user config in $HOME/.config/dunst/
+##     folder, changing dunst basic theming options (like fonts, colors, etc.)
+##     according to your current X resources color palette.
+##
+##     To make this possible, it reads your current user config
+##     ($HOME/.conf/dunst/dunstrc, which is copied from the default config if
+##     it does not exist) and changes the attributes values with  new ones (see
+##     Theming section for more info). Then it dumps the new configuration to
+##     $user_xr_color_conf file.
+##
+## Theming
+##
+##     To change colors and fonts:
+##
+##     * Firstly you have to ensure that those dunst attributes are defined in
+##       the corresponding sections. For example, to change the frame_color
+##       value from the global section, in $HOME/.config/dunst/dunstrc should
+##       be:
+##
+##       ...
+##       [global]
+##           ...
+##           frame_color = <value>
+##           ...
+##
+##     * Then, you can change attribute values as you wish with the
+##       following format in theme_attr_dict:
+##
+##       ["<sec>-<attr>"]="<val>|$(xrdb_get '<X_res>' '<dev_val>')"
+##
+##       Each <variable> means the following:
+##
+##       * sec - current section name e.g.: global.
+##       * attr - current attribute name e.g.: frame-color.
+##       * val - a simple value e.g.: "#fffeee", Monospace, 11...
+##       * X_res - X resource name e.g.: color8.
+##       * dev_val - default value to set if X_res is not found e.g.: #65737e.
+##
+##       Theming example (you can check theme_attr_dict for other values):
+##
+##       ["global-frame_color"]="\"$(xrdb_get 'color8' '#65737e')
+##
+##     The function xrdb_get, searches the first parameter in the X resources
+##     database with appres (command installed from xorg-appres in archlinux
+##     and x11-utils in ubuntu). If the first parameter does not exist, the
+##     function returns the second parameter. For hex colors, is important to
+##     scape " characters for proper functioning of dunst config reader (for
+##     example "#ffeegg").
+##
+##     You can define X_res variables in $HOME/.Xresources file. For in depth
+##     syntax go to https://wiki.archlinux.org/index.php/X_resources.
+##
+###############################################################################
+
+set -e
+
+# Check if appres is installed
+if [ ! "$(command -v appres)" ]; then
+    printf 'Install xorg-appres in archlinux and x11-utils in debian/ubuntu.\n'
+    exit 1
+fi
+
+readonly script_name="$(basename "$0")"
+readonly base_dir="$(realpath "$(dirname "$0")")"
+
+# Show ussage
+usage() {
+    grep -e '^##[^#]*$' "$base_dir/$script_name"  | \
+        sed -e "s/^##[[:space:]]\{0,1\}//g" \
+            -e "s/<script_name>/${script_name}/g"
+    exit 2
+} 2>/dev/null
+
+# Show help
+if [ "$#" -gt 0 ]; then
+    if ! [[ "$1" == "--help" || "$1" == "-h" ]]; then
+        printf '\nUnknown option.\n'
+    fi
+    usage
+fi
+
+# Function to get resource values
+xrdb_get () {
+    output="$(appres Dunst | grep "$1:" | head -n1 | cut -f2)"
+    default="$2"
+    printf '%s' "${output:-$default}"
+}
+
+#
+# Attributes dictionary. Add or remove attributes (see header for more info)
+#
+declare -A theme_attr_dict=(
+    ["global-font"]="$(xrdb_get 'font' 'Monospace') $(xrdb_get '*.font_size:' '11')"
+    ["global-frame_width"]="$(xrdb_get 'border_width' '1')"
+    ["global-frame_color"]="\"$(xrdb_get 'color8' '#65737e')\""
+
+    ["urgency_low-background"]="\"$(xrdb_get 'color0' '#2b303b')\""
+    ["urgency_low-foreground"]="\"$(xrdb_get 'color4' '#65737e')\""
+    ["urgency_low-frame_color"]="\"$(xrdb_get 'color4' '#65737e')\""
+
+    ["urgency_normal-background"]="\"$(xrdb_get 'color0' '#2b303b')\""
+    ["urgency_normal-foreground"]="\"$(xrdb_get 'color2' '#a3be8c')\""
+    ["urgency_normal-frame_color"]="\"$(xrdb_get 'color2' '#a3be8c')\""
+
+    ["urgency_critical-background"]="\"$(xrdb_get 'color0' '#2b303b')\""
+    ["urgency_critical-foreground"]="\"$(xrdb_get 'color1' '#bf616a')\""
+    ["urgency_critical-frame_color"]="\"$(xrdb_get 'color1' '#bf616a')\""
+)
+
+# Attributes dictionary keys.
+readonly valid_keys="${!theme_attr_dict[@]}"
+
+#
+# File paths
+#
+# User config dir and file
+readonly user_conf_dir="${XDG_CONFIG_HOME:-$HOME/.config}/dunst"
+readonly user_conf="$user_conf_dir/dunstrc"
+
+# Default config dir and example file
+example_conf_dir="/usr/share/dunst"
+example_conf="$example_conf_dir/dunstrc"
+
+# User xresources color config file
+readonly user_xr_color_conf="$user_conf_dir/dunstrc_xr_colors"
+
+
+# Check if the user config directory exists
+if ! [ -d "$user_conf_dir" ]; then
+    printf 'Creating folder "%s".\n' "$user_conf_dir"
+    mkdir -p  "$user_conf_dir"
+fi
+
+# Check if the user config file exists
+if ! [ -f "$user_conf" ]; then
+    printf '"%s" file does not exist.\nChecking for config file example.' \
+               "$user_conf"
+
+    if [ -d "/usr/share/dunst" ]; then
+        # Archlinux default dir and example file
+        example_conf_dir="/usr/share/dunst"
+
+        if [ -f "$example_conf_dir/dunstrc" ]; then
+            example_conf="$example_conf_dir/dunstrc"
+        else
+            printf 'Could not find the example config file in "%s".
+               \nPlease, change $example_conf variable in the script.' \
+                   "$example_conf_dir"
+            exit 1
+        fi
+
+    elif [ -d "/usr/share/doc/dunst" ]; then
+        # Debian/Ubuntu default dir
+        example_conf_dir="/usr/share/doc/dunst"
+
+        if [ -f "$example_conf_dir/dunstrc.example.gz" ]; then
+            # Ubuntu <= 17.10 and Debian <= 1.2.0-2 example file:
+            example_conf="$example_conf_dir/dunstrc.example.gz"
+        elif [ -f "$example_conf_dir/dunstrc.gz" ]; then
+            # Ubuntu >= 18.04 and Debian >= 1.3.0-2 example file:
+            example_conf="$example_conf_dir/dunstrc.gz"
+        else
+            printf 'Could not find the example config file in "%s".
+               \nPlease, change $example_conf variable in the script.' \
+                   "$example_conf_dir"
+            exit 1
+        fi
+
+    else
+        printf 'Could not find the example config directory.
+           \nPlease, change $example_conf_dir variable in the script.'
+        exit 1
+    fi
+
+    printf 'Copying example config to "%s".\n' "$user_conf_dir"
+
+    # Get the extension to check if the file is compressed
+    if [[ "${example_conf##*\.}" == "gz" ]]; then
+        # Extract example file to user config file
+        gunzip -c "$example_conf" > "$user_conf"
+    else
+        cp "$example_conf" "$user_conf_dir"
+    fi
+
+fi
+
+# Regular expressions
+readonly re_section_line='^\[(.*)\]$'
+readonly re_empty_comment_line='(^$)|(^[[:space:]]*(\#)|(;))'
+readonly re_attribute_line='^([[:space:]]*)([_[:alnum:]]+)'
+
+# Create an array with the file lines
+mapfile -t conf < "$user_conf"
+
+# Iterate over the file lines
+for idx in "${!conf[@]}"; do
+    # Current line
+    curr_line="${conf[$idx]}"
+    # If we are in a new section:
+    if [[ "$curr_line" =~ $re_section_line ]]; then
+        curr_section="${BASH_REMATCH[1]}"
+        continue
+    fi
+    # Skip the line if it is empty or has a comment
+    if [[ "$curr_line" =~ $re_empty_comment_line ]]; then
+        continue
+    fi
+    # Get the attribute in the current line
+    [[ "$curr_line" =~ $re_attribute_line ]]
+    curr_attr_name="${BASH_REMATCH[2]}"
+    curr_sett_name="${curr_section}-${curr_attr_name}"
+    # If the current attribute is not in our dictionary, continue
+    case "$valid_keys" in
+        *"$curr_sett_name"*)
+            printf -v conf[$idx] '    %s = %s' \
+                   "${curr_attr_name}" \
+                   "${theme_attr_dict[$curr_sett_name]}"
+            ;;
+    esac
+done
+
+# Create a header for the xr_color config file
+user_xr_color_conf_content="\
+###################################
+#
+# Config file created with
+# $script_name wrapper
+#
+###################################
+
+"
+
+# After everything is completed, write the new config to a file
+user_xr_color_conf_content+="$(printf '%s\n' "${conf[@]}")"
+
+printf '%s\n' "$user_xr_color_conf_content" > "$user_xr_color_conf"
+
+printf '"%s" updated successfully.\n' "$user_xr_color_conf"
+


bebehei:
> Any tips or improvements?

First glance:

- Run shellcheck over it and quote everything.
- AFAIK `declare -A` isn't POSIX, so the shebang doesn't match this properly. (have to research it).

Code looks quite good for the rest.

lulivi:
Ok! I'm on it.

lulivi:
@bebehei I've been searching and the simplest idea I have right now is to create a function with a case [from this answer](https://stackoverflow.com/a/22151682) something like this:

```shell
theme_vars () {
    case "$1" in
        "key1") printf '%s' "value1";;
        "key2") printf '%s' "value2";;
        ....
    esac
}

printf 'dict[key1]=%s' "$(theme_vars 'key1')"
```
I've searched for [sh posix standard](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html) and `declare` appears there (don't know it that means anything). It doesn't talk about arrays or associative arrays so... I don't know.

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,115 @@
...
+#
+# It gets your current user configuration file
+# -$XDG_CONFIG_HOME/dunst/dunstrc, which is copied from the default
+# config filein /usr/share/dunst/dunstrc if it does not exists- and
+# changes the defined attributes -declared furder on in the script as
+# $theme_attr_dict dictionary- to the new ones. Then it dumps the new
+# configuration to $user_themed_conf file.
+#
+###############################################################################
+
```
tsipinakis: Using `set -e` would probably be wise in such scripts to catch any errors.

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,115 @@
...
+[ -d $user_conf_dir ] || {
+    printf '%s\n' "$user_conf_dir folder doesn't exist, creating..."
+    mkdir -p  $user_conf_dir
+}
+
+# If it exists, check if the user config file exists
+[ -f $user_conf ] || {
+    printf '%s does not exist.\n.' "$user_conf"
+    printf 'Copying default config to %s...\n' "$user_conf_dir"
+    cp $def_conf $user_conf_dir
```
tsipinakis: Should ensure that `$def_conf` exists in the first place, not every distro has the default dunstrc in `/usr/share/dunst/dunstrc`, in Debian for example it's in `/usr/share/doc/dunstrc.gz`

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,115 @@
...
+}
+
+# Regular expressions
+re_section_line='^\[(.*)\]$'
+re_empty_comment_line='(^$)|(^[[:space:]]*\#)'
```
tsipinakis: Semicolons can also be used as comments.

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,115 @@
...
+re_attribute_line='^[[:space:]]{4}([_[:alnum:]]+)'
```
tsipinakis: Why just 4 spaces? Let's support all indentation types. Something like `^(?:[[:space:]]+)?([_[:alnum:]]+)` should do the trick.

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,115 @@
...
+    ["urgency_critical-foreground"]="\"$(xrdb_get 'color15:' '#eff1f5')\""
+    ["urgency_critical-frame_color"]="\"$(xrdb_get 'color3:' '#ebcb8b')\""
+)
+
+# Iterate over the file lines
+for idx in "${!conf[@]}"; do
+    # Current line
+    curr_line="${conf[$idx]}"
+    # If we are in a new section:
+    [[ "$curr_line" =~ $re_section_line ]] && {
```
tsipinakis: Why use these weird contraptions from logical operators rather than the usual `if/fi` pair? They are confusing at first glance.

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env sh
```
tsipinakis:
```
$ sh ./dunst_xr_theme_changer.sh
./dunst_xr_theme_changer.sh: 56: ./dunst_xr_theme_changer.sh: mapfile: not found
./dunst_xr_theme_changer.sh: 63: ./dunst_xr_theme_changer.sh: Syntax error: "(" unexpected
```

It should probably use `bash` rather than `sh`.

contrib/dunst_xr_theme_changer.sh
lulivi: Is there any other possible default directories to include in the search? Or should I just ask the user to change `$user_conf` path?

lulivi: I did some research and realized about the following:

|Distribution|Directory|file|
|:---:|:---:|:---:|
|Archlinux|`/usr/share/dunst`|`dunstrc`|
|Debian < 1.3.0-2 / Ubuntu < 18.04|`/usr/share/doc/dunst/`|`dunstrc.example.gz`|
|Debian >= 1.3.0-2 / Ubuntu >= 18.04|`/usr/share/doc/dunst/`|`dunstrc.gz`|

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,162 @@
...
+
+if [ -d "/usr/share/dunst" ]; then
+    # Archlinux default dir and file
+    example_conf_dir="/usr/share/dunst"
+
+    if [ -f "$example_conf_dir/dunstrc" ]; then
+        example_conf="$example_conf_dir/dunstrc"
+    else
+        printf 'Could not find the example config file in %s.
+               \nPlease, change \$example_conf variable in the script' "$example_conf_dir"
```
bebehei: As you're using single quotes in the format string, you'll print `\$example_conf` and not solely `$example_conf`.

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,162 @@
...
+
+    if [ -f "$example_conf_dir/dunstrc.example.gz" ]; then
+        # Ubuntu <= 17.10 and Debian <= 1.2.0-2 default file:
+        example_conf="$example_conf_dir/dunstrc.example.gz"
+    elif [ -f "$example_conf_dir/dunstrc.gz" ]; then
+        # Ubuntu >= 18.04 and Debian >= 1.3.0-2 default file:
+        example_conf="$example_conf_dir/dunstrc.gz"
+    else
+        printf 'Could not find the example config file in %s.
+               \nPlease, change \$example_conf variable in the script' "$example_conf_dir"
```
bebehei: As you're using single quotes in the format string, you'll print `\$example_conf` and not solely `$example_conf`.

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,162 @@
...
+        exit 1
+    fi
+
+else
+    printf 'Could not find the example config directory.
+           \nPlease, change \$example_conf_dir variable in the script'
```
bebehei: As you're using single quotes in the format string, you'll print `\$example_conf` and not solely `$example_conf`.

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,162 @@
...
+# config file in /usr/share/dunst/dunstrc if it does not exist- and
+# changes the defined attributes -declared further on in the script as
+# $theme_attr_dict dictionary- to the new ones. Then it dumps the new
+# configuration to $user_xr_color_conf file.
+#
+###############################################################################
+
+set -e
+
+script_name="$(basename $0)"
```
bebehei: `"$0"`

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,162 @@
...
+
+# Function to get resource values
+xrdb_get () {
+    output=$(xrdb -q | grep -e "$1" | cut -f2)
```
bebehei: Quotes for `$(...)`

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,162 @@
...
+# User config dir and file
+user_conf_dir="${XDG_CONFIG_HOME:-$HOME/.config}/dunst"
+user_conf="$user_conf_dir/dunstrc"
+
+# User xresources color config file
+user_xr_color_conf="$user_conf_dir/dunstrc_xr_colors"
+
+# Check if the user config directory exists
+if ! [ -d "$user_conf_dir" ]; then
+    printf '%s\n' "$user_conf_dir folder doesn't exist, creating..."
```
bebehei: `printf 'Creating folder "%s".\n' "$user_conf_dir"`

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,162 @@
...
+# $script_name wrapper
+#
+###################################
+
+"
+
+# After everything is completed, write the new config to a file
+user_xr_color_conf_content+="$(printf '%s\n' "${conf[@]}")"
+
+printf '%s\n' "$user_xr_color_conf_content" > $user_xr_color_conf
```
bebehei: Quotes for the steamname variable? `> "$user_xr_color_conf"`

bebehei:
Oh, let's put a logic test on the script:

You haven't installed dunst into `/usr/bin`, but you already have a working configuration file. Will the script work?

lulivi:
Thank you both @bebehei and @tsipinakis for your nitpicks! I think the code is cleaner now.

Any other change you would like to make to the script?

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,166 @@
...
+            exit 1
+        fi
+
+    elif [ -d "/usr/share/doc/dunst" ]; then
+        # Debian/Ubuntu default dir
+        example_conf_dir="/usr/share/doc/dunst"
+
+        if [ -f "$example_conf_dir/dunstrc.example.gz" ]; then
+            # Ubuntu <= 17.10 and Debian <= 1.2.0-2 example file:
+            example_conf="$example_conf_dir/dunstrc.example.gz"
```
tsipinakis: This won't work, the file in ubuntu and debian is compressed with gzip  you'll need to uncompress it to use it.

bebehei:
> Any other change you would like to make to the script?

The code quality is close to bash perfectness. But the actual point, if the script works is completely unclear to me. I admit, I haven't executed the script and I'm too lazy to do it.

I'd like to ask some guys from #357 to test it.

---

@lulivi You squashed the commit. Thanks. But you also squashed the commit messages. And most of them are invalid now. e.g: "Quoted $0". There is no need to quote `$0` as you already did it right in the commit ;-) It would be great, if you would remove the junk messages.

nhatzHK:
I tested the script. It seems to be working fairly well, except for 2 things:
- [Line 88](https://github.com/lulivi/dunst/blob/24aa046ff6adb2682e1e7bcee0667b89f422466f/contrib/dunst_xr_theme_changer.sh#L88)
    When executed, it crashes on this line with:
    `./dunst_xr_theme_changer.sh: line 88: conditional binary operator expected`
    
    If I comment the whole if/else block out, it runs without error.

- The generated file
    There seem to be some mistakes when putting the colours in the generated file.
    ```
    [urgency_critical]
        background = "#0d170a
    #0d170a"
        foreground = "#9E7263
    #9E7263"
        frame_color = "#9E7263
    #9E7263"
    ```
    The generated file has the right colours, but the string is terminated on a new line, and the colour is repeated in it.

I would suggest to put the colours in variables at the top of the config file, but dunst probably doesn't support variables, and then the user could reuse these variables in the config file. i3wm does the same thing with `set`.
```
set $var "value" # define and set the variable's value
...
...
client.urgent $var # use it to configure other stuff later on
```

Even though this script is going to work pretty well, I think you guys should check out how i3wm lets the user access the [X ressource database](https://i3wm.org/docs/userguide.html#xresources) in the config file.

Anyway this is awesome part from the little things I pointed out.
:+1:

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/env bash
+###############################################################################
+#
+# This script creates a dunst user configuration file in
+# $XDG_CONFIG_HOME/dunst/ folder, changing dunst basic teeming options
```
tadly: `teeming` -> `theming` (I assume)
Just because I noticed it :)

lulivi:
Thank you @nhatzHK for your review!

1. I'll check that if test later.
2. I think the problem with duplicating the color has to do with [this](https://github.com/lulivi/dunst/blob/24aa046ff6adb2682e1e7bcee0667b89f422466f/contrib/dunst_xr_theme_changer.sh#L23) line. It is one of the last changes and maybe didn't test it. 
3. Thanks @tadly for noticing the misspelling!

I have a test tomorrow, so I will start with the debug after it.

lulivi:
@nhatzHK I've tried the final script, and it does not duplicate the colors for me...

* `~/.config/dunst/dunstrc`

    ```ini
    [urgency_normal]
        background = "#285577"
        foreground = "#ffffff"
        frame_color = "#00ff00"
        timeout = 5
        #icon = /path/to/icon
    ```

* `~/.config/dunst/dunstrc_xr_colors`

    ```ini
    [urgency_normal]
        background = "#2b303b"
        foreground = "#a3be8c"
        frame_color = "#a3be8c"
        timeout = 5
        #icon = /path/to/icon
    ```

Don't know why is working bad for you.

bebehei:
> EDIT: @bebehei what do you think about the squashed commits now? I tried to keep the important ones.

I strongly recommend https://chris.beams.io/posts/git-commit/ There is no better thing. @tsipinakis, right? :wink:

A good commit message could be:

```
Add xresources config wrapper
```

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,174 @@
...
+#
+###############################################################################
+
+set -e
+
+script_name="$(basename "$0")"
+
+# Function to get resource values
+xrdb_get () {
+    output="$(xrdb -q | grep -e "$1" | cut -f2)"
```
bebehei: > The generated file has the right colours, but the string is terminated on a new line, and the colour is repeated in it.

I think here is the problem. You have to make sure, that only a single line is retrieved. Currently you grep for anything matching the string `$1`. **It could be even the value, which matches.**. Usually Xresources have `*color0: #abcdef`, but you also could define `XTerm*color0: #123456`. And your regex would match both.

As @nhatzHK has already mentioned: You should use another method to query the X resource database, which also takes care of the precedence of specific values.

@nhatzHK Just to verify my thesis: What does `xrdb -q | grep 'color0:'` give as the output?

nhatzHK: I think this is right. I do have several lines ending with `colorX` in my X ressource database
```sh
$ xrdb -q | grep 'color0:'
*.color0:	#0d170a  ## notice the leading dot
*color0:	#0d170a
```

lulivi: The thing is: should I just get the first line?

e.g.: If you have

```bash
XTerm*color0:	#123456
*.color0:	#0d170a
*color0:	#0d170a
```

then with `head -n1` will get the XTerm color:

```bash
$ xrdb -q | grep "color0:" | head -n1 | cut -f 2
#123456
```

So maybe it is interesting to place colors at the begining or at the end. Other option is to write the exact xresource name:

```bash
$ xrdb -q | grep "*.color0:" | head -n1 | cut -f 2
#0d170a
```

But maybe this option will get the same problem that we have before, something like

```bash
XTerm*.color0:	#123456
*.color0:	#0d170a
```

What do you think?

tsipinakis: How about we throw more regex at it?
```
^(?:\*?\.?[a-zA-Z0-9]+):[[:space:]]?(.*)$
or
^(?:\*?\.?color0):[[:space:]]?(.*)$
```
to catch wildcards and
```
^(?:Dunst\*?\.?[a-zA-Z0-9]+):[[:space:]]?(.*)$
or
^(?:Dunst\*?\.?color0):[[:space:]]?(.*)$
```
for dunst specifically.

lulivi:
> I strongly recommend https://chris.beams.io/posts/git-commit/ There is no better thing. @tsipinakis, right? 😉
> A good commit message could be:
> ```
> Add xresources config wrapper
> ```

Oh sorry, I understood you wanted me to keep relevant commits. So maybe I should follow that commit structure? 

```
Add xresources config wrapper

Created a wrapper script to customize theme variables (colors,
fonts, etc) according with your X resources file (~/.Xresources).

As dunst is not a X exclusive notification daemon, this script
could be a workarround to solve this.

Related with: #357
```

What do you think? Or should I just keep it in oneline as you suggested?

contrib/dunst_xr_theme_changer.sh
bebehei: Hey guys, if I read the docs right, the key should be the `appres` program. No regex involved.

lulivi: Hum... Can't figure out how to make it work :(

tsipinakis: This should be as simple as replacing `xrdb -q` with `appres Dunt` and then using ` | head -n1` as you suggested. Some testing shows that it always puts the more specific match at the top.

lulivi: Thanks Nikos! I was trying something like `Dunst.color0` and wasn't working. The script is fixed now.

lulivi:
> The only other nitpick I'd mention is maybe have theme_attr_dict at the top, under the config variables, so that everything "user editable" is in one place and one doesn't have to dig through the script to find them.

That's so true. I'm on it!

bebehei:
The changes and the script looks good.

I'm putting myself in the role of a user, who just wants to use this wrapper and has no clue about the usage of it:

- How do I seed the X resource database to pass values to dunst?
- How do I run this wrapper?

IMO, these questions should get answered in the header. I think that'll make the difference between "there is something usable but nobody uses it" and "there is something good and everyone uses it".

lulivi:
All right.

Maybe I could do a more extended header. Something like:

```
###############################################################################
##
## Usage
##
##     ./dunst_xr_theme_changer.sh
##
##     If it doesn't execute, give execute permissions to de file with 
##     chmod blabla...
##     And then run ./dunst_xr_theme_changer.sh blab...
##
## Description
##
##     This script blablabla...
##
## Theming
##
##     To theme colors and fonts, you have to change the variables in 
##     theme_attr_dict blabla...
##     Variables are obtained from Xresources, which are defined in 
##     ~/.Xresources this way blablabla...
##
###############################################################################
```

lulivi:
Hey guys, I've made some improvements in the header (you can check the [extended text](https://pastebin.com/VHA0HsmR) and tell me what do you think) and created 2 functions to output : 

* [extended help](https://imgur.com/JV8dDTA) - when `-h` or `--help` are the script options.
* [basic help](https://imgur.com/6mm8jeH) - when the command line option is wrong. It just show usage.

bebehei:
I play the nerd card first: What command line prompt do you have?

---

If you add them, I don't have any objections and I'm glad to add this to dunst!

The only question, which now came to my mind: How will this script come the user's machine. Is this a maintainer's job to put that file manually somewhere in the package or is this our job?

lulivi:
> I play the nerd card first: What command line prompt do you have?

The prompt is a modified version of the bash_git_prompt for fish [check it out here](https://github.com/lulivi/dotfiles/blob/master/.config/fish/config.fish).

> How will this script come the user's machine. Is this a maintainer's job to put that file manually somewhere in the package or is this our job?

Hummm... 🤔 Thats a really good question. This script is clearly an external part of dunst. It is not as integrated as `dunsk_espeak.sh`. I would suggest maybe to show it as a wrapper script you can use to make a connection between Xresources and dunst. So ai think is our job as developers/users to grab the script and run it. Maybe it can be delivered as the default dunstrc file, and only grabbed if needed. Sorry for my english, sometimes its pretty bad.

tsipinakis:
> How will this script come the user's machine. Is this a maintainer's job to put that file manually somewhere in the package or is this our job?

I can only speak for debian since that's what I'm familiar with: contrib directories and similar utility scripts are usually put along with documentation in `/usr/share/doc/<program name>`

bebehei:
@tsipinakis Thanks. I guess my question is answered positively!

@lulivi please add the docs, you've pasted and - if @tsipinakis has no further wishes - I'm glad to merge it!

contrib/dunst_xr_theme_changer.sh
```diff
@@ -0,0 +1,255 @@
...
+##       ["global-frame_color"]="\"$(xrdb_get 'color8' '#65737e')
+##
+##     The function xrdb_get, searches the first parameter in the X resources
+##     database with appres (command installed from xorg-appres in archlinux
+##     and x11-utils in ubuntu). If the first parameter does not exist, the
+##     function returns the second parameter. For hex colors, is important to
+##     scape " characters for proper functioning of dunst config reader (for
+##     example "#ffeegg").
+##
+##     You can define X_res variables in $HOME/.Xresources file. For in depht
```
tsipinakis: `depht` typo

tsipinakis:
Thanks again for the work put into this @lulivi.

PR #528, opened 2018-07-07
Various Improvements

bebehei:
Various small improvements in the code. Too small to fire a PR for each one.


src/option_parser.c
```diff
@@ -181,9 +179,7 @@ const char *next_section(const char *section)
 int ini_get_bool(const char *section, const char *key, int def)
 {
         const char *value = get_value(section, key);
-        if (value == NULL)
-                return def;
-        else {
+        if (value)
```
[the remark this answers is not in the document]
bebehei: Ouch, that hurt. The full switch statement not covered in braces. :face_with_head_bandage: Very good point!

bebehei:
> I'd say 6d07559 is a bit questionable,

Yeah. But it's just a plain assignment code. Nothing more. There is no real logic involved. And the important bits (assignment of `n->dbus_client`) get hidden in the big forest of plain assignments.

Except  `n->actions`. I've put this again back to the end of the function again.

So, plain assignments are now inside the `case`-forest while the logic remains outside.

src/dbus.c
```diff
@@ -240,28 +227,28 @@ static notification *dbus_message_to_notification(const gchar *sender, GVariant
...
-                                                progress = g_variant_get_uint32(dict_value);
+                                                n->progress = g_variant_get_uint32(dict_value);
                                                 g_variant_unref(dict_value);
                                         }
                                 }
                                 break;
                         case 7:
                                 if (g_variant_is_of_type(content, G_VARIANT_TYPE_INT32))
-                                        timeout = g_variant_get_int32(content);
+                                        n->timeout = g_variant_get_int32(content) * 1000;
```
tsipinakis: This is also removing the sanity check that if `timeout < 0` then `timeout = -1`, so timeouts less than -1 are no undefined behaviour.

bebehei: `notification_init` should handle this:

https://github.com/dunst-project/dunst/blob/f1dd772937686d704dc943d4b0e5b074fd76792a/src/notification.c#L327-L329

PR #531, opened 2018-07-31
Refactor typenames

### LINKED ISSUE #483: Internal format of type names
A tiny nitpick: We have a very unclear naming scheme of our type names. Some are written in camel case (`RawImage`, `Actions`), while some others are separated with underscores (`notification`, `cairo_ctx_t`) and may or may not have a `_t` suffix.

We should make this clear, what to use.

bebehei:
Refactor all types into structs and rename it properly to match the discussion in #483 

Fixes #483


tsipinakis:
Looking at the implementation of #483 in code some of the types just don't sit right to me to type `struct` all the time. I think `color`, `notification`, `raw_image`, `window_x11` (maybe also `colored_layout`) should be typedefs since they are either mostly/totally opaque or dead simple structs.

bebehei:
> [...] should be typedefs since they are either mostly/totally opaque [...]

Sorry, but I don't see, why those types should be opaque. As long as a struct's field gets accessed directly, it can't be an opaque type.

> [...]  or dead simple structs.

And here, I'd turn logics 'round: Why should you warrant a dead simple struct a typedef?

---

TBH it looks crazy in this PR, because it's 300 lines change just about adding `struct` and forming the typenames.

bebehei:
Oh, well. It might make sense to convert `x11_window *win` in draw.[ch] into an opaque type (but not in `x11/x.c`).

This would make sense.

tsipinakis:
I have been hovering over the merge button but something is still bothering me in this change and I figure it'd be a mistake to leave it unmentioned.

Looking at the `struct`s again, it feels noisy and part of me feels that the typedefs had better readability which makes me realise that I'm partly changing my mind on this guideline. Why not typedef structs *if they are a simple struct* (i.e. not pointers, unions or anything else). It's usual in C when an unknown type comes along to assume it's a struct.

bebehei:
> Looking at the structs again, it feels noisy and part of me feels that the typedefs had better readability 

Well. Yeah, it's a bit more "noisy". But it's about 80 lines change for the notification type on a project which has got >5000 SLOCs. I wouldn't claim this is noisy.

And most of the times, you actually interact with a notification in the code, you use the variable's name using `n->field`. So most of the lines, where you interact with a notification in the code, are still unchanged.

> Why not typedef structs if they are a simple struct (i.e. not pointers, unions or anything else).

Because it's a structure. Nothing more, nothing less. There is no need to make abstractions. It's not used anywhere outside.

> It's usual in C when an unknown type comes along to assume it's a struct.

Sorry, but the applied logic in here is a bit crazy. How am I supposed to know the structure of the struct, if it's unknown!?!

---

My point of this PR is not to change anything to a struct, just because structs are better. My main goal is to achieve consistency. How we achieve this, is secondary. If we `typedef` everything, I'd be ok with that. But I thought we already agreed on not typedeffing everything.

So, could you explain me please the logics, why should we use a `typedef` for the `notification` but not for the other structs?

bebehei:
Hey I've rebased the commits and took care of your `struct color color` example.

I hope we can finally resolve the typenames and stick them to a simple rule.

tsipinakis:
> Hey I've rebased the commits and took care of your struct color color example.

Thanks, that indeed looks much better.

Looking at this almost a month later I realize I was probably going at this the wrong way, trying to solve confusingly named variables by using the ambiguous typedef syntax. This is indeed a really good step forward in terms of readability, merged!.

PR #532, opened 2018-08-06
Added notification logging.

PaddiM8:


config.h
```diff
@@ -38,6 +38,9 @@ settings_t defaults = {
 .show_indicators = true,
 .word_wrap = false,
 .ellipsize = middle,
+.log_path = "~/dunst.log",
```
bebehei: Obviously, there's a more sensible default path.

dunstrc
```diff
@@ -144,6 +144,15 @@
     # Possible values are "start", "middle" and "end".
     ellipsize = middle
 
+    # Where to save notification the notification log
```
bebehei: Add these changes to the manpage, too!

log-patch.diff
```diff
@@ -0,0 +1,129 @@
+diff --git a/config.h b/config.h
```
bebehei: NO!!!

src/dunst.c
```diff
@@ -110,7 +110,7 @@ gboolean quit_signal(gpointer data)
 static void teardown(void)
 {
         regex_teardown();
-
+        g_free(settings.log_path);
```
bebehei: Obviously the wrong place to free it.

src/notification.c
```diff
@@ -282,6 +282,29 @@ void notification_init(notification *n)
         n->body     = n->body     ? n->body     : g_strdup("");
         n->category = n->category ? n->category : g_strdup("");
 
+        /* log */
+        if (settings.log_notifications) {
```
bebehei: Sorry, but indentation is 8 spaces.

bebehei: Oh, and of course: Why do you do this stuff **before** sanitation!?

src/settings.c
```diff
@@ -231,6 +231,29 @@ void load_settings(char *cmdline_config_path)
                 g_free(c);
         }
 
+        settings.log_path = option_get_string(
+                "global",
+                "log_path", "-log_path", defaults.log_path,
+                "Log file location"
+        );
+
+        char *home = getenv("HOME");
```
bebehei: There is `string_to_path` and `option_get_path`.

src/notification.c
bebehei: And thirdly, there's `notification_format_message`, which does the job with a few modifications much better IMHO.

dunstrc
```diff
@@ -144,6 +144,15 @@
...
+    log_path = "~/.dunst.log"
+
+    # Log all notifications and save them in a FILE
+    log_notifications = no
```
bebehei: Oh, please make the defaults consistent to `config.h`.

bebehei: And why do you put the `log_path` as the **first** option, while there is a boolean `log_notifications`!?

src/notification.c
```diff
@@ -282,6 +282,29 @@ void notification_init(notification *n)
...
+            FILE *log_file;
+            log_file = fopen(settings.log_path, "a");
```
bebehei: What about Permission denied?!?

src/dunst.c
bebehei: See travis failure.

src/notification.c
```diff
@@ -282,6 +282,29 @@ void notification_init(notification *n)
...
+
+            for(char *c = settings.log_format; *c; c++) {
```
tsipinakis: Don't reinvent the wheel, we already have `notification_format_message` that does this exact thing and can be easily repurposed (hint: make `format` a parameter).

log-patch.diff
tsipinakis: What's up with this patch file? Git already keeps track of changes so what's the point in having it here?

PaddiM8: Forgot to delete that.

config.h
PaddiM8: I made it because me and some other people needed the feature. My friends switched to another notification daemon because dunst doesn't have notification logging actually. Forked it for personal usage, made a pull-request because eh why not maybe you'd find the feature useful (even though the code isn't the best since it's the first time I use C hehe). Maybe at least implement it yourself if you found it useful. Thank you for the code review, I'll see and learn. 

Is just since to have notification logging included in the actual notification daemon.

src/dunst.c
PaddiM8: Not sure where else though.

src/notification.c
PaddiM8: Oo nice.

src/notification.c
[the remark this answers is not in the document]
PaddiM8: Good point!

He-Ro:
I want to note that having the log output spread across multiple lines makes it hard to process. Let's assume you want to only see the entries corresponding to one application. With the current way you cannot grep for the application name (at least not with greps default options).
I think every log entry should therefore be on one line.

bebehei:
> I made it because me and some other people needed the feature.

Please tell us **why** you need the feature! What does it enable for you for that way?

bebehei:
@He-Ro Are you interested in this feature, too!? What's your desired usecase for it?

config.h
PaddiM8: We use notification logs to catch up on messages and notifications we miss. Very handy if you're away for a while and  your computer is still on, or if you need to focus for a while, then you can read the notifications later on(also it's a handy way to see messages people delete, which happens quite a bit in our case). It's very nice to always know that your notifications are saved somewhere.

PaddiM8:
@He-Ro Very good point! I made it possible to change the log format, so it is possible to fix that yourself. Although I should probably make it default.

config.h
PaddiM8: Oops it came all the way up here.

He-Ro:
@bebehei I do not have a use case for the logging. My thoughts were more general about the format of any log file.

bebehei:
Ok, it's obvious, this PR won't get merged into master.

But while seeing this PR, I once came up with a similar idea to log notifications, while introducing the logging framework. There are structured log messages available in systemd, and [GLib supports it as a backend](https://developer.gnome.org/glib/stable/glib-Message-Logging.html#using-structured-logging). IIUC, you can pass any data like a dictionary and a format string. So you could crawl any data with it and can still stringformat later the actual message. 

IDK if this works out, but it's a more sophisticated approach. And also, the only remaining `printf` calls in dunst are actually the ones from `notification_print`. It might be a good opportunity to remove them, too!

Issue #533, opened 2018-08-07
Every message to stdout / stderr should show program name for better troubleshooting


willbprog127:
Greetings! :smile: 

For a while now I've been receiving 'anonymous' error and warning messages about 'the frame section is deprecated' and ''allow_markup' is deprecated' printed on the console and also in `.xsession-errors`.  Because there is no program name before the messages, it makes it difficult to troubleshoot which program is having the problem.  I finally determined it was dunst.

Can you please prefix any printed errors or warnings with '[dunst]: '?

An example from `.xsession-errors`:

```
Warning: 'allow_markup' is deprecated, please use 'markup' instead.   <<??????????????
Warning: 'icon_folders' is deprecated, please use 'icon_path' instead.   <<??????????????
Warning: The frame section is deprecated, width has been renamed to frame_width and moved to the global section.   <<??????????????
Warning: The frame section is deprecated, color has been renamed to frame_color and moved to the global section.   <<??????????????
```
Most users would ask themselves "Where are these messages coming from?  How can I fix it if I don't know which program is having the problem?"

So 
`Warning: 'allow_markup' is deprecated, please use 'markup' instead.`
should be
`[dunst] Warning: 'allow_markup' is deprecated, please use 'markup' instead.`

Thanks! :+1: 

### Installation info

- Version: `1.3.0 (2018-01-05)`
- Install type: `Debian .deb package - automatic install`
- Distro and version: `Debian Testing x86_64`

bebehei:
Full ACK. AFAIK, there is `G_LOG_DOMAIN`, which should be settable during runtime. I'm working on a solution right now.

I don't like a fixed prefix. This would make no sense, if it's running inside a systemd unit.

willbprog127:
@bebehei  
> "I don't like a fixed prefix. This would make no sense, if it's running inside a systemd unit."

That's fine, whatever you want.  It just would be nice to know where the messages are coming from since most other entries in `.xsession-errors` and on the console include the app's name.

tsipinakis:
Looking at this again I'm not sure this is something for us to deal with.

For one we would need some way of detecting whether the output is going to `.xsession-errors` which I'm not sure how to achieve.

Secondly, taking a look at my 300k+ lines huge `.xsession-errors` practically no other program does this by default. This should be something that's up to the logger to do albeit making a change to something as archaic as X11 is downright impossible.

Systemd does it right with the journal feature, I'm not sure what else to do to improve the situation.
[...]

PR #534, opened 2018-08-07
added dunst: identifier to log messages

fwarren:
Hi,
In response to https://github.com/dunst-project/dunst/issues/533 I added dunst: before the error level text in log.c


bebehei:
Thanks for firing a pull request. I'll close this PR, as the best fix would be in the log function itself, respecting the `log_domain` parameter. I'm currently figuring out if it's actually doable, that it's set during runtime.

The `log_level_to_string` should be a plain mapping function working coherent to the function `log_set_level_from_string`.

PR #541, opened 2018-09-06
Concatenated queues

### LINKED ISSUE #431: Queue handling with settings.sort

### LINKED ISSUE #467: Five calls to notify-send produce six notifications on screen
Hi!

Sometimes there are exactly five new messages in my inbox and Thunderbird + GNotifier addon make dunst display something like this:

2017-12-24-130406_1280x800_scrot

And when those five notifications disappear, then the fifth message appears.

The thing is, if it's okay to display five notifications at the same time, dunst should just display the five messages it received. If five notifications are too many, then it should only display four, the fourth being "(2 more)".

The above picture was generated with the command ``for i in {1..5}; do notify-send asdf "the quick brown fox jumps over the $i lazy dogs"; done``

### Installation info

- Version: 1.2.0, but also tried latest from master
- Install type: "dunst" package from Arch Linux's "community" repository, or "dunst-git" package from AUR
- Distro and version: 64 bit Arch Linux

bebehei:
Let important notifications seep into displayed even if the `display` queue is full.

By checking the joint of both queues, if you concatenate now the `waiting` and `display` queues, the concatenated queue is still sorted. A push to waiting and `queues_update()` should show the important notifications even if there if display is full.

Previously, this would have required a timeout of a notification in the displayed queue. Now it works immediately.

Fixes #431 
Fixes #467

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -26,7 +26,7 @@ typedef struct {
         char *text;
         PangoAttrList *attr;
         cairo_surface_t *icon;
-        notification *n;
+        const notification *n;
 } colored_layout;
 
 window_x11 *win;
@@ -245,7 +245,7 @@ static PangoLayout *layout_create(cairo_t *c)
         return layout;
 }
 
-static colored_layout *layout_init_shared(cairo_t *c, notification *n)
+static colored_layout *layout_init_shared(cairo_t *c, const notification *n)
 {
         colored_layout *cl = g_malloc(sizeof(colored_layout));
         cl->l = layout_create(c);
@@ -301,7 +301,7 @@ static colored_layout *layout_init_shared(cairo_t *c, notification *n)
         return cl;
 }
 
-static colored_layout *layout_derive_xmore(cairo_t *c, notification *n, int qlen)
+static colored_layout *layout_derive_xmore(cairo_t *c, const notification *n, int qlen)
 {
         colored_layout *cl = layout_init_shared(c, n);
         cl->text = g_strdup_printf("(%d more)", qlen);
@@ -350,12 +350,10 @@ static GSList *create_layouts(cairo_t *c)
         int qlen = queues_length_waiting();
         bool xmore_is_needed = qlen > 0 && settings.indicate_hidden;
 
-        notification *last = NULL;
         for (const GList *iter = queues_get_displayed();
                         iter; iter = iter->next)
         {
                 notification *n = iter->data;
-                last = n;
 
                 notification_update_text_to_render(n);
 
@@ -371,7 +369,7 @@ static GSList *create_layouts(cairo_t *c)
         if (xmore_is_needed && settings.geometry.h != 1) {
                 /* append xmore message as new message */
                 layouts = g_slist_append(layouts,
-                        layout_derive_xmore(c, last, qlen));
+                        layout_derive_xmore(c, queues_get_head_waiting(), qlen));
         }
 
         return layouts;
--- a/src/queues.c
+++ b/src/queues.c
@@ -258,7 +258,7 @@ void queues_history_pop(void)
         n->redisplayed = true;
         n->start = 0;
         n->timeout = settings.sticky_history ? 0 : n->timeout;
-        g_queue_push_head(waiting, n);
+        g_queue_insert_sorted(waiting, n, notification_cmp_data, NULL);
 }
 
 /* see queues.h */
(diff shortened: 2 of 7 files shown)


bebehei:
So, while working on the queues, I also had an idea for #467. So, I've pushed additionally a fix for #467 

Now, two issues are addressed, which are which are important to the 1.4 release! :muscle:

bebehei:
@tsipinakis I'd love to see your review :wink:

tsipinakis:
Sorry, I've tried to review this twice already but both times there's something bothering me with the way this is implemented but at the same time I can't think of a better way to do it. I'll look into it more tomorrow.

src/queues.c
```diff
@@ -377,6 +377,40 @@ void queues_update(bool fullscreen)
...
+        /* If displayed is actually full, let the more important notifications
+         * from waiting seep into displayed.
+         */
+        if (settings.sort) {
+                GList *i_waiting, *i_displayed;
+
+                while (   (i_waiting   = g_queue_peek_head_link(waiting))
+                       && (i_displayed = g_queue_peek_tail_link(displayed))) {
+
+                        while (   fullscreen
```
tsipinakis: I'd break it down into a `queues_notification_ready` helper function to make this clearer.

src/queues.c
```diff
@@ -377,6 +377,40 @@ void queues_update(bool fullscreen)
...
+                               && i_waiting->data
+                               && ((notification*) i_waiting->data)->fullscreen != FS_SHOW) {
+                                i_waiting = i_waiting->prev;
+                        }
+
+                        if (i_waiting && notification_cmp(i_displayed->data, i_waiting->data) > 0) {
+                                notification *towait = i_displayed->data;
+                                notification *todisp = i_waiting->data;
+
+                                g_queue_delete_link(displayed, i_displayed);
```
tsipinakis: This can also be abstracted into `queues_swap_notification`

src/queues.c
```diff
@@ -355,9 +355,21 @@ void queues_update(bool fullscreen)
...
-                        /* the list is full */
-                        break;
+                if (displayed_limit > 0) {
+                        int cur_displayed_limit = displayed_limit-1;
+
+                        if (!settings.indicate_hidden || waiting->length == 1)
+                                cur_displayed_limit++;
+
+                        while (displayed_limit <= displayed->length) {
+                                notification *n = g_queue_pop_tail(displayed);
```
tsipinakis: Can't we can simplify this entire block like so?:
```c
if (   displayed->length == displayed_limit - 1
    && queues_notification_ready() > 1)
        break; // queue is full, last slot will be used by xmore
```

src/draw.c
```diff
@@ -347,31 +347,34 @@ static GSList *create_layouts(cairo_t *c)
 {
         GSList *layouts = NULL;
 
-        int qlen = queues_length_waiting();
-        bool xmore_is_needed = qlen > 0 && settings.indicate_hidden;
+        int waiting_len = queues_length_waiting();
+        bool xmore_is_needed = waiting_len > 0 && settings.indicate_hidden;
 
         notification *last = NULL;
+        int slots_avail = settings.geometry.h;
```
tsipinakis: What's the point of this variable? If there are notifications in `waiting` we definitely want to have an `xmore`

tsipinakis:
Somehow I submitted the review before I was actually done, oops!

Works fine from what I can tell, so my nitpicks this time are about simplifying the code :p

src/queues.c
```diff
@@ -355,9 +363,21 @@ void queues_update(bool fullscreen)
                 notification *n = iter->data;
                 GList *nextiter = iter->next;
 
-                if (displayed_limit > 0 && displayed->length >= displayed_limit) {
-                        /* the list is full */
-                        break;
+                if (displayed_limit > 0) {
```
bebehei: Note to myself: What to do, if `displayed_limit = 1` (the case when the user only wants to show a single notification at all times)?

src/draw.c
bebehei: I guess, based on the current logic, I can remove it again. But while writing the initial code, I had a constraint and it was necessary to check it.

bebehei:
@tsipinakis I've simplyfied the code now and polished the commits. I'd consider it final now. A review would be great! Thanks!

src/queues.c
```diff
@@ -382,6 +385,12 @@ void queues_update(bool fullscreen)
                 iter = nextiter;
         }
 
+        /* if necessary, push the overhanging notifications from displayed to waiting again */
+        while (displayed->length > cur_displayed_limit) {
```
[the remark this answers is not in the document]
bebehei: This is necessary. Regard the following case:

- `settings.geometry.h == 5`.
- Send 5 Notifications
    - => `cur_displayed_limit = 5`
- Send 6th notification
    - => `cur_displayed_limit = 4`
    - but `displayed->length == 5`

So we have to make space for the xmore notification, which will get generated inside the draw stuff.

src/queues.c
```diff
@@ -75,6 +75,49 @@ unsigned int queues_length_history(void)
...
+                                      GQueue *queueB,
+                                      GList  *elemB)
+{
+        notification *toB = elemA->data;
+        notification *toA = elemB->data;
+
+        g_queue_delete_link(queueA, elemA);
+        g_queue_delete_link(queueB, elemB);
+
+        if (toA) g_queue_insert_sorted(queueA, toA, notification_cmp_data, NULL);
```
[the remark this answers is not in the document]
bebehei: Proper point!

src/queues.c
```diff
@@ -353,13 +346,23 @@ void queues_update(bool fullscreen)
                 }
         }
 
+        int cur_displayed_limit;
+        if (settings.geometry.h == 0)
```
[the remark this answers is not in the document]
bebehei: The most important bit is the `displayed->length + waiting->length > settings.geometry.h`. This can't be done in `queues_init`. Putting this code into exactly this position is the actual fix for #467.

src/queues.c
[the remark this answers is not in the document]
bebehei: Update pushed!

PR #545, opened 2018-09-13
Draw event optimizations

tsipinakis:


src/x11/x.c
```diff
@@ -326,20 +326,14 @@ gboolean x_mainloop_fd_dispatch(GSource *source, GSourceFunc callback, gpointer
...
-                        wake_up();
-                        break;
                 case CreateNotify:
                         LOG_D("XEvent: processing 'CreateNotify'");
                         if (win->visible &&
                             ev.xcreatewindow.override_redirect == 0)
                                 XRaiseWindow(xctx.dpy, win->xwin);
                         break;
+                case FocusIn:
+                case FocusOut:
```
bebehei: I don't get the point of removing the `FocusOut`/`FocusIn` cases. I don't get the point why we even covered these cases.

But at least, the following `LOG_D` call becomes misleading in this commit.

tsipinakis: According to my tests not checking for focus events breaks i3 when using the keyboard to switch between monitors. I'll chalk it up to "black magic" and leave it as is, that if statement stops draw from being called more than necessary anyway.

bebehei: Do you have a reproducer for this? I've tested now dunst with the cases `FocusIn`/`FocusOut` both removed. And it doesn't even make any difference on my system.

Because the window associated to the `Focus(In|Out)` event never matches our window and when receiving a FocusChange event, there's always another `PropertyNotify` event in the queue, which has an atom `_NET_ACTIVE_WINDOW`, which resembles exactly the `Focus(In|Out` event.

bebehei: Ouch. This hurts! I found the bug :see_no_evil: :angry: 

I guess, Captain Obvious hasn't called us, but we have to update the window's current screen more often: https://github.com/bebehei/dunst/commit/1faf7c671df379ec3c87778f92ac225cf7612d1e

And to debug the error, use this commit: https://github.com/bebehei/dunst/commit/a423c204e1524d2b57ca83f72fd1eede5b29337e

Look at the debug output while processing the events. The `get_screen_active()` when using the keyboard to switch has got a HUGE delay. And when listening to the `FocusChange` event, it does exactly that. Can you also see, that the window gets moved between the monitors with a longer delay, when using the keyboard?

tsipinakis: Yikes, where is that delay coming from though? I can still reproduce the issue with the commit that updates the active screen more often, the debug commit is the one that fixes it (specifically, the delay).

bebehei:
Ok, I understand the problem now:

- We determine the current desktop, by crawling our screen data and looking forward to calculate the some x coordinates
- We calculate the x coordinates are either
    - the position of the current window focused window (`FOLLOW_KEYBOARD`)
    - the result of calling `XQueryPointer` on the root window (`FOLLOW_MOUSE`)
- Then we look forward if these coordinates are in the rectangle of one of our screen configs and return the index of it.
- -> **This is the result of `get_active_screen()`

The problem: `XQueryPointer` is slow and if I understand [this thread](https://lists.freedesktop.org/archives/xorg/2012-January/053818.html) correctly, we're to fast during processing (what the debug commit showed).

So, we're actually relying on the fact, that multiple events are sent on a focus change and while processing one of those events, the `XQueryPointer` will return the new position and we can switch the screen.

This is shitty, but we can't change it.

----

For the future, I'd suggest a more radical approach: There is [`_NET_CURRENT_DESKTOP` from ewmh](https://standards.freedesktop.org/wm-spec/wm-spec-1.3.html#idm140130317686320). Usually window managers send with the `PropertyNotify` the atom, which changed. And if `_NET_CURRENT_DESKTOP` changed, you just change the screen to the new value.

In case of i3, the option `focus_follows_mouse <yes|no>` is always aligned with dunst's `follow = <mouse|keyboard>`.

PR #547, opened 2018-09-14
Random fixes

knopwob:
This fixes minor stuff that I came across while refamiliarizing myself with the code and running clangs scan-build over it.

### DIFF
--- a/src/dbus.h
+++ b/src/dbus.h
@@ -17,7 +17,6 @@ enum reason {
 
 int initdbus(void);
 void dbus_tear_down(int id);
-/* void dbus_poll(int timeout); */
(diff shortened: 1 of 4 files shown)


tsipinakis:
> that I came across while refamiliarizing myself with the code

Sounds promising, do you intend to ~~start~~, well, come back to working on dunst regularly? There are quite a few still open issues that could use some work :).

knopwob:
I'm not sure yet. Initially I just wanted to fix the flickering bug and realized how little of the code I still recognize. 
Maybe a few fixes here and there but work and other commitments don't leave me enough time to get back into development like the "good old times"(tm).

Issue #540, opened 2018-09-04
dunst does not display JPEG icons without a filename extension


eadmund:
`dunst` correctly displays icons from files whose names end in '.jpeg' or '.jpg,' but it does _not_ display them when they do not.

To reproduce:

```
$ notify-send --icon /path/to/jpeg-image Test "no image displays!"
```

### Installation info

- Version: `v1.3.2-143-g896c008`
- Install type: `manually`
- Distro and version: `Debian 9 (stretch)`

tsipinakis:
I cannot reproduce this on debian buster. I have a feeling this is something to do with `gdk-pixbuf` which we use to load icons.

Perhaps in the version that's in stretch it only looks at the extension? I will set up a stretch VM later this week to test.

bebehei:
I can't reproduce it either. And I've tested it on Xenial, trusty and bionic now.

- What's your gdb pixbuf version?
- What does `gdk-pixbuf-query-loaders` spit out on your system? (Watch out, this program spits out much stuff. I'm only interested in the lines for the jpeg loader!)

@eadmund And to rule out a possible misbehavior[0] of dunst: How have you tested it? Have you executed any `mv` call while the notification had been shown?
----

[0]: Dunst reloads the notification icon whenever it's drawing a notification complete freshly. So if you `mv` the file and trigger a reload, the icon disappears.

eadmund:
Looks like `gdk-pixbuf` is at 2.36.5-2+deb9u2

I get this from `gdk-pixbuf-query-loaders`:

```
…
"/usr/lib/x86_64-linux-gnu/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-jpeg.so"
"jpeg" 5 "gdk-pixbuf" "JPEG" "LGPL"
"image/jpeg" ""
"jpeg" "jpe" "jpg" ""
"\377\330" "" 100
…
```

I ran `dunst`, then used both `notify-send` and D-Bus directly to send notifications, using different JPEG files, with & without `.jpeg`/.`jpg`. I did not move any files in between sending the notification and seeing it drawn. I _did_ confirm that a file missing the final `.jpeg` works _with_ it (so it's _probably_ not a malformed file issue).

bebehei:
@eadmund Thanks for the answer. Sadly nothing, which enlightens me.

I've written a small helper here: https://gist.github.com/bebehei/d1248deab8199ea93d877fdf126be588

[...]

PR #548, opened 2018-09-17
limit icon size when we have a fixed width

knopwob:
This is a proposed fix for #540.

I'm not reallty sure/happy with this fix but it seems to be the easiest.

I'm not sure wether we should overwrite `max_icon_size` if it's not 0, or should we assume the user knows what he's doing?

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -429,6 +429,24 @@ void load_settings(char *cmdline_config_path)
                 "Scale larger icons down to this size, set to 0 to disable"
         );
 
+        // restrict the icon size to a reasonable limit if we have a fixed width.
+        // Otherwise the layout will be broken by too large icons.
+        // See https://github.com/dunst-project/dunst/issues/540
+        if (settings.geometry.width_set) {
+                const int icon_size_limit = settings.geometry.w / 2;
+                if (   settings.max_icon_size == 0
+                    || settings.max_icon_size > icon_size_limit) {
+                        if (settings.max_icon_size != 0) {
+                                LOG_W("Max width was set to %d but got a max_icon_size of %d, too large to use. Setting max_icon_size=%d",
+                                        settings.geometry.w, settings.max_icon_size, icon_size_limit);
+                        } else {
+                                LOG_I("Max width was set but max_icon_size is unlimited. Limiting icons to %d pixels", icon_size_limit);
+                        }
+
+                        settings.max_icon_size = icon_size_limit;
+                }
+        }
+
         // If the deprecated icon_folders option is used,
         // read it and generate its usage string.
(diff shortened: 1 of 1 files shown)


src/settings.c
```diff
@@ -474,6 +474,17 @@ void load_settings(char *cmdline_config_path)
...
         );
 
+        // restrict the icon size to a reasonable limit if we have a fixed width.
+        // Otherwise the layout will be broken by too large icons.
+        // See https://github.com/dunst-project/dunst/issues/540
+        if (settings.geometry.width_set) {
+                int icon_size_limit = settings.geometry.w / 2;
+                if (settings.max_icon_size == 0 ||
+                    settings.max_icon_size > icon_size_limit) {
+                        settings.max_icon_size = icon_size_limit;
```
bebehei: For consistency: In the past we've put the logic concatenators of if conditions in front of the new line: 

```
                if (   settings.max_icon_size == 0
                    || settings.max_icon_size > icon_size_limit) {
                        settings.max_icon_size = icon_size_limit;
```

knopwob:
Shrink will still work. The image is limited to the max_win_width/2, but when the rest of the notification is smaller than half the width it will still shrink the window. It looks rather weird but that's always the case when you have a huge image.

I agree that there are probably better solutions to this if you factor this into the drawing/icon code but
I'm not sure it is worth it to complicate that code even further.

Another way to see it is that the user created a configuration that is broken.
He wants a window that is limited to a max width but on the other hand wants to see
icons in their full size (or he set max_icon_size to a value that is unreasonably high for his max window width). So one could argue that we're just sanitizing the configuration. :D 

It would be nice to have the option to resize the image to a size that's fit the rest of the notification. In other words layout the text first and derive the icon size from the resulting text size.
But I think this is a different feature.

Anyway you know the code better than me at this point so I leave the decision up to you wether you
want this fixed within the settings code or in the drawing/icon code.

tsipinakis:
@knopwob seems to be inactive here so I took over in order to unblock 1.4.0.

Can't see a better solution so I just added some warning messages to call it a day.

bebehei:
Could you please rebase on current master, to run through CircleCI, too?

tsipinakis:
Done!

genofire:
It does not work for me correct,
i run it with: 
```
geometry             = 0x0-4+24
max_icon_size        = 32
```

But i get:
```
 WARNING: Max width was set to 0 but got a max_icon_size of 32, too large to use. Setting max_icon_size=0
```

Does i something wrong?

tsipinakis:
@genofire See #614 it's a bug when dynamic width is used. It's fixed in master and I'll put out a point release once the few other bugs that were discovered after 1.4 are fixed.

PR #550, opened 2018-10-01
Misc patches

bebehei:
Just some miscellaneous changes, accumulated over the past weeks.

TBH, the libxdg removal came out of the blue. I found out today about the [`g_get_user_config_dir` function](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-get-user-config-dir), which makes `libxdg-basedir` irrelevant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Remove libxdg-basedir dependency (GLib's function is used instead)
 - `fullscreen` rule to hide notifications when a fullscreen window is active
 - When new notifications arrive, but display is full, important notifications don't
   have to wait for a timeout in a displayed notification (#541)
--- a/src/icon.c
+++ b/src/icon.c
@@ -161,18 +161,18 @@ cairo_surface_t *icon_get_for_notification(const struct notification *n)
         int h = gdk_pixbuf_get_height(pixbuf);
         int larger = w > h ? w : h;
         if (settings.max_icon_size && larger > settings.max_icon_size) {
-                GdkPixbuf *scaled;
-                if (w >= h) {
-                        scaled = gdk_pixbuf_scale_simple(pixbuf,
-                                        settings.max_icon_size,
-                                        (settings.max_icon_size * h) / w,
-                                        GDK_INTERP_BILINEAR);
-                } else {
-                        scaled = gdk_pixbuf_scale_simple(pixbuf,
-                                        (settings.max_icon_size * w) / h,
-                                        settings.max_icon_size,
-                                        GDK_INTERP_BILINEAR);
-                }
+                int scaled_w = settings.max_icon_size;
+                int scaled_h = settings.max_icon_size;
+                if (w >= h)
+                        scaled_h = (settings.max_icon_size * h) / w;
+                else
+                        scaled_w = (settings.max_icon_size * w) / h;
+
+                GdkPixbuf *scaled = gdk_pixbuf_scale_simple(
+                                pixbuf,
+                                scaled_w,
+                                scaled_h,
+                                GDK_INTERP_BILINEAR);
                 g_object_unref(pixbuf);
                 pixbuf = scaled;
         }
--- a/src/icon.c
+++ b/src/icon.c
@@ -76,7 +76,7 @@ GdkPixbuf *get_pixbuf_from_file(const char *filename)
 
 GdkPixbuf *get_pixbuf_from_icon(const char *iconname)
 {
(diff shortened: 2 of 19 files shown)


src/settings.c
```diff
@@ -252,14 +255,12 @@ void load_settings(char *cmdline_config_path)
         {
                 char *c = option_get_string(
                         "global",
-                        "follow", "-follow", "",
+                        "follow", "-follow", NULL,
```
tsipinakis: Default should be `defaults.f_mode`.

bebehei: Well, this wasn't the case before either, but I fully agree with you. New commit fixes it.

src/icon.c
```diff
@@ -161,18 +161,18 @@ cairo_surface_t *icon_get_for_notification(const struct notification *n)
...
-                } else {
-                        scaled = gdk_pixbuf_scale_simple(pixbuf,
-                                        (settings.max_icon_size * w) / h,
-                                        settings.max_icon_size,
-                                        GDK_INTERP_BILINEAR);
-                }
+                int scaled_w = settings.max_icon_size;
+                int scaled_h = settings.max_icon_size;
+                if (w>=h)
+                        scaled_w = (settings.max_icon_size * h) / w;
```
tsipinakis: Opps! Sorry I missed this earlier: You've reversed this, if `width >= height` then the width should be scaled down to `max_icon_size`, and we have to calculate the appropriate height.

And do the reverse if `height > width`.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Remove libxdg-basedir dependency (GLib's function is used instead)
```
tsipinakis: > GLib's g_get_user_config_dir function does exactly the same thing

Unfortunately for us, it doesn't. Debian places the default config in /etc/xdg/dunst/dunstrc which `libxdg-basedir` loads properly, but `g_get_user_config_dir` does not.

src/utils.h
```diff
@@ -4,6 +4,9 @@
 
 #include <glib.h>
 
+#define STR_EMPTY(s) !STR_FULL(s)
+#define STR_FULL(s) (s && strnlen(s, 1))
```
tsipinakis: Isn't `strnlen(s,1)` the same as `s[0] != '\0'`?

tsipinakis: I mentioned it because it does the exact same thing and it saves a function call in the meantime. Its just a nitpick but good change to have IMO.

CHANGELOG.md
bebehei: Good point. I've rewritten it and it'll respect now the system dirs, too!

src/utils.h
```diff
@@ -4,6 +4,10 @@
 
 #include <glib.h>
 
+#define STR_EMPTY(s) (*s == '\0')
+#define STR_NULL_OR_EMPTY(s) (!s || STR_EMPTY(s))
```
tsipinakis: At this point I'm not sure about the usefulness of this macro, it's longer than typing out `(!s || *s == '\0')` directly.

bebehei: Well yeah, that's true. But it's definitely easier to read and understand.

Do you have a better solution to cope GCC's weird understanding of NULL-pointer guards? https://travis-ci.org/dunst-project/dunst/jobs/437820396#L730-L735

tsipinakis: To be fair, there's nothing weird about it, `buf` in that warning is on the stack so it will always exist.

I don't think there's a better solution.

bebehei: I wasn't satisfied with this either.

We'll keep only `STR_EMPTY` and `STR_FULL` and throw `STR_NULL_OR_EMPTY` away.

I got the insight, that most standard library methods work only on a char buffer and not just do something with the buffer, but also propagate the length of the buffer they used.

So testing your stack buffer via `STR_EMPTY` is actually pointless, because it's a char buffer and not a string.

src/log.c
```diff
@@ -32,23 +32,23 @@ void log_set_level_from_string(const char *level)
         if (!level)
                 return;
 
-        if (g_ascii_strcasecmp(level, "critical") == 0)
+        if (STR_EQ(level, "critical"))
```
tsipinakis: This breaks the case insensitive comparison, if I call `log_set_level_from_string("CRITICAL")` it'll fail for example. I'm not sure if we depend on that, but it's good to have.

bebehei: Well, we didn't depend on it, but it makes perfect sense using italics for the loglevel.

Issue #445, opened 2017-11-24
New command line control


bebehei:
We should write a command line control for dunst.

Currently, control takes place only via keyboard and a some [hackish way via notifications](https://github.com/dunst-project/dunst/blob/6e9a805017a51d3628d64bf47632d4947f3d66c8/src/queues.c#L57-L73). These notifications are effective, but quite ugly and the keyboard shortcuts are not working on some environments.

I propose to add a new binary, which can send commands to the dunst server via an additional DBus interface.

The features, this command line tool should have:

- close one/all notifications
- pop history
- pause/unpause/toggle dunst
- retrieve information about notifications
- ... ?

This would in addition fix/invalidate all of these issues.

#308 #307 #216 #171 #103 #112 #138 #284 #241

tsipinakis:
Duplicate/relevant: #284

bebehei:
I started work on it: https://github.com/bebehei/dunst/tree/dunstcmd

`dunstcmd status` is already working in a preliminar form.

ckipp01:
Hey! Thanks for all the work on this. I've seen this issue linked from quite a few others. Is there a status update on the ability to be able to query whether or not dunst is paused or not? I've recently switched to dunst and normally just have a toggle switch to switch on or off notifications. It's sort of killing me that there seemingly isn't a good way to query for that info in dunst.

tsipinakis:
Not much to update, it's on the roadmap but the progress is only what you see on the branch above. I'm not sure if @bebehei is still planning to work on this.

bebehei:
Yes, I had been on it. But in the meantime, I actually don't know whether I'm finishing the client side in C. The current architecture for the client is ugly. I'd actually prefer Python for the client (or have a try with Rust). 

On the server side, the problems are solved and the architecture is clear. The intended feature set is not yet implemented on the server side, but that's an easy game.

bebehei:
The server side is "finished". I dropped the C-client. @stunkymonkey will finish the client side. We've settled with Rust.

tsipinakis:
That's awesome, I've been trying to learn rust recently and this is yet another excuse to get to it.

Is there a repo set up somewhere for it? (Or are we putting it in here along with everything else?)

bebehei:
> Is there a repo set up somewhere for it? (Or are we putting it in here along with everything else?)

There is a project, but it's currently hidden and just for test usage.

[...]

Issue #216, opened 2015-02-22
Allow to check whether dunst is paused


roomcays:
I'd like to check if dunst is paused within my Conky script. Something like

```
killall -SIGUSR2 dunst
```

would be OK, if it can print something out to STDOUT, but then command

```
killall -SIGUSR1 dunst
```

should toggle pause/resumed mode.
Alternatively one might use notifications with pre-defined summaries.

hughdavenport:
Hmm,

I think changing the signals may cause a regression in a lot of user scripts.

Sending a message with summary of say DUNST_COMMAND_STATUS should be possible, but it won't return to stdout, but instead show a notification regardless of paused status? What are your thoughts here?

Cheers,

Hugh

arnegoeteyn:
I would also like to check the status of dunst in my lemonbar config. Is there already a way to check check this? Or a dunst toggle would be great as well.
Cheers

raspbeguy:
I think the easiest way that wouldn't cause any regression (and IMHO the most UNIX-compliant one) would be to keep the status written in a file, for instance `/tmp/dunst-status`, with the same idea than what you can find in `/sys` or `/proc`. A single `cat` to this file would give the required information.

raspbeguy:
Or it could throw an info text (easily parseable) when you send a `SIGINFO`. Could be handier.

tsipinakis:
Implemented with #651.

PR #551, opened 2018-10-01
Queues tests

### LINKED ISSUE #563: Dunst doesn't respect `-r` (replace) anymore
I have fetched the latest version of Dunst today and built it. It seems that the replace functionality is not working anymore although it was working before perfectly.

``` sh
$ dunstify -r 2593 -p "test volume"
13
$ dunstify -r 2593 -p "test volume"
14
$ dunstify -r 2593 -p "test volume"
15
```
It seems that it no longer replaces the previous id and my volume notifications are stacking up on top of each other. As it's showed in the snippet every time it runs, A new id is generated and doesn't respect the `-r $id` anymore.

**The last working commit** is: `f0b047497eabf3`, so whatever has caused this issue is in the following commit that was merged on `2018-11-11`:
``` sh
d879d70da060ea 2018-10-25 12:32 +0200 Jordan Galby Implement stack_tag, implementing x-canonical-private-synchronous
```
Before this commit the result of the above snippet was as expected:
``` sh
$ dunstify -r 2593 -p "test volume"
2593
$ dunstify -r 2593 -p "test volume"
2593
$ dunstify -r 2593 -p "test volume"
2593
```
### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon v1.3.2-235-gd786381`
- Install type: `manually`
- Distro and version: ` Ubuntu 18.04.1 LTS x86_64`

bebehei:
---

Edit: Fixes #563

### DIFF
--- a/src/dunst.c
+++ b/src/dunst.c
@@ -68,7 +68,6 @@ static gboolean run(void *data)
         dunst_status(S_FULLSCREEN, have_fullscreen_window());
         dunst_status(S_IDLE, x_is_idle());
 
-        queues_check_timeouts(status);
         queues_update(status);
 
         static gint64 next_timeout = 0;
--- a/src/dunst.c
+++ b/src/dunst.c
@@ -2,6 +2,7 @@
 
 #include "dunst.h"
 
+#include <assert.h>
 #include <glib.h>
 #include <glib-unix.h>
 #include <signal.h>
@@ -24,6 +25,34 @@
 
 GMainLoop *mainloop = NULL;
 
+static struct dunst_status status;
+
+/* see dunst.h */
+void dunst_status(const enum dunst_status_field field,
+                  bool value)
+{
+        switch (field) {
+        case S_FULLSCREEN:
+                status.fullscreen = value;
+                break;
+        case S_IDLE:
+                status.idle = value;
+                break;
+        case S_RUNNING:
+                status.running = value;
+                break;
+        default:
+                LOG_E("Invalid %s enum value in %s:%d", "dunst_status", __FILE__, __LINE__);
+                break;
+        }
+}
+
+/* see dunst.h */
+struct dunst_status dunst_status_get(void)
+{
+        return status;
+}
+
 /* misc functions */
 static gboolean run(void *data);
 
@@ -36,10 +65,11 @@ static gboolean run(void *data)
 {
         LOG_D("RUN");
 
-        bool fullscreen = have_fullscreen_window();
+        dunst_status(S_FULLSCREEN, have_fullscreen_window());
+        dunst_status(S_IDLE, x_is_idle());
 
-        queues_check_timeouts(x_is_idle(), fullscreen);
-        queues_update(fullscreen);
+        queues_check_timeouts(status);
+        queues_update(status);
 
         static gint64 next_timeout = 0;
 
@@ -76,7 +106,7 @@ static gboolean run(void *data)
 
 gboolean pause_signal(gpointer data)
 {
-        queues_pause_on();
+        dunst_status(S_RUNNING, false);
         wake_up();
 
         return G_SOURCE_CONTINUE;
@@ -84,7 +114,7 @@ gboolean pause_signal(gpointer data)
 
 gboolean unpause_signal(gpointer data)
 {
-        queues_pause_off();
+        dunst_status(S_RUNNING, true);
         wake_up();
 
         return G_SOURCE_CONTINUE;
@@ -108,6 +138,9 @@ static void teardown(void)
 
 int dunst_main(int argc, char *argv[])
 {
+
+        dunst_status(S_RUNNING, true);
+        dunst_status(S_IDLE, false);
 
         queues_init();
 
--- a/src/dunst.c
+++ b/src/dunst.c
@@ -63,14 +63,14 @@ void wake_up(void)
 
 static gboolean run(void *data)
 {
+        static gint64 next_timeout = 0;
+
         LOG_D("RUN");
 
         dunst_status(S_FULLSCREEN, have_fullscreen_window());
         dunst_status(S_IDLE, x_is_idle());
 
         queues_update(status);
-
-        static gint64 next_timeout = 0;
 
         bool active = queues_length_displayed() > 0;
 
--- a/src/dunst.c
+++ b/src/dunst.c
@@ -101,7 +101,7 @@ static void teardown(void)
 {
         regex_teardown();
 
-        teardown_queues();
+        queues_teardown();
 
         draw_deinit();
 }
--- a/src/notification.c
+++ b/src/notification.c
@@ -376,7 +376,7 @@ static void notification_format_message(struct notification *n)
 
         /* replace all formatter */
         for(char *substr = strchr(n->msg, '%');
-                  substr;
+                  substr && *substr;
                   substr = strchr(substr, '%')) {
 
                 char pg[16];
@@ -450,6 +450,7 @@ static void notification_format_message(struct notification *n)
                 case '\0':
                         LOG_W("format_string has trailing %% character. "
                               "To escape it use %%%%.");
+                        substr++;
                         break;
                 default:
                         LOG_W("format_string %%%c is unknown.", substr[1]);
--- a/src/queues.c
+++ b/src/queues.c
@@ -113,12 +113,46 @@ static void queues_swap_notifications(GQueue *queueA,
  */
 static bool queues_notification_is_ready(const struct notification *n, struct dunst_status status, bool shown)
 {
+        if (!status.running)
+                return false;
         if (status.fullscreen && shown)
                 return n && n->fullscreen != FS_PUSHBACK;
         else if (status.fullscreen && !shown)
                 return n && n->fullscreen == FS_SHOW;
         else
                 return true;
+}
+
+/**
+ * Check if a notification has timed out
+ *
+ * @param n the notification to check
+ * @param status the current status of dunst
+ * @returns true, if the notification is timed out, otherwise false
+ */
+static bool queues_notification_is_finished(struct notification *n, struct dunst_status status)
+{
+        assert(n);
+
+        if (n->timeout == 0) // sticky
+                return false;
+        if (n->start == 0)   // hidden // TODO: is this really the implication of hidden?
+                return false;
+
+        bool is_idle = status.fullscreen ? false : status.idle;
+
+        /* don't timeout when user is idle */
+        if (is_idle && !n->transient) {
+                n->start = time_monotonic_now();
+                return false;
+        }
+
+        /* remove old message */
+        if (time_monotonic_now() - n->start > n->timeout) {
+                return true;
+        }
+
+        return false;
 }
 
 /* see queues.h */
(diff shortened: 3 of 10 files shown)


tsipinakis:
That's a very good catch! We hadn't had a discussion on it but previously we assigned the ID asked anyway. The notification spec doesn't specify what to do if the ID is invalid but we should preserve the previous behaviour to not break those that use a static id replacement as the old workaround for `stack_tag`.

I consider this a bug and it was probably introduced in #552.

tsipinakis:
I pushed a fix @bebehei

test/queues.c
```diff
@@ -529,6 +529,67 @@ TEST test_queues_update_paused(void)
         PASS();
 }
 
+TEST test_queues_update_sicker(void)
```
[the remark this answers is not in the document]
bebehei: `s/sicker/seep/g`.

Sorry, I totally missed that this isn't even an english word :grin:

test/queues.c
```diff
@@ -355,6 +355,65 @@ TEST test_datachange_ttl(void)
...
+        QUEUE_LEN_ALL(0, 1, 0);
+        NOT_LAST(n2);
+
+        ACT_DEINIT();
+        PASS();
+}
+
+TEST test_queue_stacktag(void)
+{
+        const char *stacktag = "THIS IS A SUPER WIERD STACK TAG";
```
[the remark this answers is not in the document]
bebehei: Intended ;-)

test/queues.h
```diff
@@ -0,0 +1,97 @@
...
+#define QUEUE(q) QUEUE_##q
+
+#define ACT_INIT() queues_init()
+#define ACT_DEINIT() teardown_queues()
+#define ACT_INS(n) queues_notification_insert(n)
+#define ACT_CLS(n) queues_notification_close(n, REASON_UNDEF)
+#define ACT_CLS_ID(id) queues_notification_close_id(id, REASON_UNDEF)
+#define ACT_RPL(n) queues_notification_replace_id(n)
+#define ACT_RUN(data) queues_update(data)
+#define ACT_REDISPLAY(data) queues_upodate(data)
```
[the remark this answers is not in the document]
bebehei: Full ACK. I first started the tests, I basically packed everything into a macro. In the meantime, I've learnt, that `ASSERTm` does a pretty awesome job in most cases.

But I hadn't been able to remove all the unused (and unnecessary) macros.

bebehei:
Most of the stuff is now fixed.

> I also found a bug: After a notification is shown and closed the dunst window isn't hidden, a small black rectangle keeps being displayed (and moves around the screens with the cursor as would be expected). Bisect showed 052d9f7 as the broken commit.

This is now fixed. The `visible` field of the `dunst_status` structure was absolutely redundant. I tripped too much over the word `visible`, which had different meanings in different contexts.

But shame on me, that I requested your review, but did not even bother to run dunst :face_with_head_bandage:

src/queues.c
```diff
@@ -121,6 +123,36 @@ static bool queues_notification_is_ready(const struct notification *n, struct du
...
+        assert(n);
+
+        if (n->timeout == 0) // sticky
+                return true;
+        if (n->start == 0)   // hidden // TODO: is this really the implication of hidden?
+                return true;
+
+        /* don't timeout when user is idle */
+        if (status.idle && !status.fullscreen && !n->transient) {
+                n->start = time_monotonic_now();
```
tsipinakis: Why are you modifying `n->start` in a query function? What's the point of this if block entirely?

src/queues.c
```diff
@@ -343,68 +375,32 @@ void queues_history_push_all(void)
 }
 
 /* see queues.h */
-void queues_check_timeouts(struct dunst_status status)
+void queues_update(struct dunst_status status)
```
tsipinakis: These are 2 huge and complicated functions, let's break them down rather than merge them.

src/notification.c
```diff
@@ -450,6 +450,7 @@ static void notification_format_message(struct notification *n)
                 case '\0':
                         LOG_W("format_string has trailing %% character. "
                               "To escape it use %%%%.");
+                        substr++;
```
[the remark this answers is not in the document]
bebehei: If I set `substr` to `NULL`, the `substr = strchr(substr, '%')` expression, which will get executed at the end of the loop will segfault.

src/queues.c
```diff
@@ -455,7 +455,7 @@ void queues_update(struct dunst_status status)
                 while (   (i_waiting   = g_queue_peek_head_link(waiting))
                        && (i_displayed = g_queue_peek_tail_link(displayed))) {
 
-                        while (i_waiting && ! queues_notification_is_ready(i_waiting->data, status, true)) {
+                        while (i_waiting && ! queues_notification_is_ready(i_waiting->data, status, false)) {
```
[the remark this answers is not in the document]
bebehei: > From here, the bug and the solution becomes obvious: We can simplify the entire function to
> 
> ```
> return n->fullscreen == FS_SHOW || !fullscreen;
> ```

I'd love to simplify it as that. But it's exactly that corner case you're pointing out. Let's rephrase it in language:

fullscreen, visible = (1,1) means that the notification got inserted and shown while in a "normal" mode, but afterwards you switched a window to fullscreen.

And that's the exact point between delay and pushback:

https://github.com/dunst-project/dunst/blob/d879d70da060ea78fe735d62249a0afdf3e61bc8/dunstrc#L318-L323

src/queues.c
bebehei: I don't see any benefit of splitting this method.

There are 4 parts, which are required to be executed in exactly this order and the logics involved is nowhere else necessary.

Additonally, the complex testing logic is already splitted out into `queues_notification_is_(ready|finished)`.

src/queues.c
```diff
@@ -121,6 +123,36 @@ static bool queues_notification_is_ready(const struct notification *n, struct du
                 return true;
 }
 
+/**
+ * Check if a notification has timed out
+ *
+ * @param n the notification to check
+ * @param status the current status of dunst
+ * @returns true, if the notification isn't timed out, otherwise false
```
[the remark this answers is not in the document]
bebehei: Thanks for pointing out such a stupid nobrainer.

src/queues.c
bebehei: You're never considered idle, if you're in fullscreen. Current master:

https://github.com/dunst-project/dunst/blob/ed341dfca30ed496bb056a00251af6a17d1f7a07/src/queues.c#L346

I think, I should use exactly this line in there also in the new function.

src/queues.c
```diff
@@ -134,15 +134,15 @@ int queues_notification_insert(struct notification *n)
         }
         /* Do not insert the message if it's a command */
         if (STR_EQ("DUNST_COMMAND_PAUSE", n->summary)) {
-                pause_displayed = true;
+                dunst_status(S_RUNNING, false);
```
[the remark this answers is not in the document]
tsipinakis: I don't think we'd ever need to actively signal a command line tool as it usually doesn't hang around, makes a request and exits.

But either case aside IMO it's best to avoid needless abstraction on a "might need this" basis, especially one as simple to implement later as this.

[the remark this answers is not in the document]
bebehei: I agree with you IMO, but I'm already working on #445 and to have the ability for a signal is mandatory for #216.

Also the status field is declared static in dunst.c.

test/queues.h
```diff
@@ -0,0 +1,57 @@
...
+#define QUEUE_CONTAINSm(msg, q, n) ASSERTm(msg, g_queue_find(QUEUE(q), n))
+
+#define NOT_LAST(n) do {ASSERT_EQm("Notification " #n " should have been deleted.", 1, notification_refcount_get(n)); g_clear_pointer(&n, notification_unref); } while(0)
+
+static inline struct notification *test_notification(const char *name, gint64 timeout)
+{
+        struct notification *n = notification_create();
+
+        if (timeout != -1)
+                n->timeout = S2US(timeout);
```
tsipinakis: Huh, here's a case I didn't consider: What if timeout is `-1`? It will be multiplied anyway and there are some places in the code we explicitly check for `-1` like [`rules.c`](https://github.com/dunst-project/dunst/blob/master/src/rules.c#L17), need to check if there are breakages with this.

test/queues.h
```diff
@@ -0,0 +1,57 @@
...
+        } while (0)
+
+#define QUEUE_IS(q, exp) QUEUE_ISm("QUEUE_IS(" #q "," #exp ")", q, exp)
+#define QUEUE_ISm(msg, q, exp) ASSERTm(msg, (exp == QUEUE(q)))
+#define QUEUE_CONTAINS(q, n) QUEUE_CONTAINSm("QUEUE_CONTAINS(" #q "," #n ")", q, n)
+#define QUEUE_CONTAINSm(msg, q, n) ASSERTm(msg, g_queue_find(QUEUE(q), n))
+
+#define NOT_LAST(n) do {ASSERT_EQm("Notification " #n " should have been deleted.", 1, notification_refcount_get(n)); g_clear_pointer(&n, notification_unref); } while(0)
+
+static inline struct notification *test_notification(const char *name, gint64 timeout)
```
tsipinakis: What's the point of the `timeout` parameter? In the test cases you change the field you are testing on the next instruction, why have a parameter that's `-1` most of the time?

src/dunst.c
```diff
@@ -24,6 +25,34 @@
 
 GMainLoop *mainloop = NULL;
 
+static struct dunst_status status;
+
+/* see dunst.h */
+void dunst_status(const enum dunst_status_field field,
+                  bool value)
```
tsipinakis: `value` can also be `const`, but is there a point declaring variables on the stack as `const`?

test/queues.h
bebehei: `-1` is the default for the notification's timeout:

https://github.com/dunst-project/dunst/blob/d786381cb0d52788de56732fe0f781e4823b6e99/src/notification.c#L312

A timeout of `-1` doesn't interfere with rules at all.

Also in `notification_init` every timeout below 0 will become the default timeout.

https://github.com/dunst-project/dunst/blob/d786381cb0d52788de56732fe0f781e4823b6e99/src/notification.c#L340-L342

test/queues.h
bebehei: ?

src/dunst.c
bebehei: "Just because I can":tm:

Actually there is no specific reason. I started doing it with strings (where it's a useful thing).

But yes, either both params should be `const` or not.

test/queues.h
tsipinakis: Nevermind, I took a look at the first few test commits and noticed that `timeout` was `-1` for all of them and assumed it was a redundant parameter. A quick `grep` told me it's actually used more often.

test/queues.c
```diff
@@ -166,8 +166,116 @@ TEST test_queue_teardown(void)
...
+
+        struct notification *n = test_notification("n", 10);
+
+        queues_notification_insert(n);
+        ASSERTm("The inserted notification is inside the waiting queue, so it should get ignored.",
+               queues_get_next_datachange(time_monotonic_now()) < S2US(0));
+
+        queues_update(STATUS_NORMAL);
+        ASSERT_IN_RANGEm("The notification has to get closed in less than its timeout",
+               S2US(10), queues_get_next_datachange(time_monotonic_now()), S2US(10));
```
tsipinakis: `ASSERT_IN_RANGE` doesn't make sense here, for one it assumes the value is double which makes the error message confusing by using scientific notation and secondly if `S2US(15)` is returned it still passes as it checks for `+/- S2US(10)`.

test/queues.c
```diff
@@ -166,8 +166,116 @@ TEST test_queue_teardown(void)
...
+        queues_update(STATUS_NORMAL);
+        ASSERT_IN_RANGEm("The notification has to get closed in less than its timeout.",
+               S2US(n->timeout), queues_get_next_datachange(time_monotonic_now()), S2US(n->timeout));
+
+        n = test_notification("n2", 10);
+
+        queues_notification_insert(n);
+        queues_update(STATUS_NORMAL);
+        ASSERT_IN_RANGEm("The timeout of the second notification has to get used as sleep time now.",
+               S2US(n->timeout), queues_get_next_datachange(time_monotonic_now()), S2US(n->timeout));
```
tsipinakis: Same thing in this test.

test/queues.c
```diff
@@ -2,6 +2,36 @@
...
+        queues_notification_close(n, REASON_UNDEF);
+
+        n = test_notification("n2", 0);
+        queues_notification_insert(n);
+        queues_update(STATUS_NORMAL);
+
+        n = test_notification("n3", 0);
+        queues_notification_insert(n);
+
+        QUEUE_LEN_ALL(1,1,1);
```
tsipinakis: It'd be a good idea to also check as you're adding the notifications to ensure that they're being put in the correct queue rather than only at the end.

test/queues.c
```diff
@@ -2,6 +2,36 @@
...
+
+        ASSERT_EQm("Queue waiting has to contain an element",
+                   1, queues_length_waiting());
+        ASSERT_EQm("Queue displayed has to contain an element",
+                   1, queues_length_displayed());
+        ASSERT_EQm("Queue history has to contain an element",
+                   1, queues_length_history());
```
tsipinakis: These are redundant: if `QUEUE_LEN_ALL` fails above they won't get run either way, if it succeeds they will pass.

bebehei: I already wanted to rebase and remove it. The test seems duplicated.

Bit it isn't: The test wants to test `queues_length_*`, so we have those `ASSERT_EQm`.

test/queues.c
bebehei: It's not important which notification is where. It's just an assertion, that every queue will contain an element.

And it seems to be redundant, but think about it like an assertion. If the `queues_notification_insert` and `queues_update` will fail, the test doesn't step into the problem, that `queues_length_*` reports the wrong size.

test/queues.c
bebehei: There's something worse. Do you spot the problem?

test/queues.c
bebehei: Oh, well it **does** make sense. Most importantly it has to be less than the actual timeout. But also it should be non-negative (this would be the representation for not sleeping).

test/queues.c
tsipinakis: Oh, calling `S2US` on `timeout` nice catch!

tsipinakis:
Heads up: I pushed a fixup for a small bug in my id assignment commit.

bebehei:
> Heads up: I pushed a fixup for a small bug in my id assignment commit.

Thoughts in order:

1. Yay, thanks! 
1. Wait, all your comments which got automatically marked outdated, aren't outdated anymore!?
1. Surprise surprise, all my changes are gone! :tada: 

I'm glad that I'm a GitHub beta member (you can click on the images to see the diff):

[Force Push IMG](https://github.com/dunst-project/dunst/compare/c646aeced84a14e016a92b3e562214fe96ea362a..69aa7c4fa942adba807cdb4ea5423f2966804297)

[Force push tsipinakis](https://github.com/dunst-project/dunst/compare/69aa7c4fa942adba807cdb4ea5423f2966804297..811090d77a99efbd62acb1b19e12fee602c5306e)

Please fetch the upstream pull request the next time first. If you need to fetch them very easily, there's a simple refspec available to fetch the GitHub PRs automatically: https://bebehei.de/github-prs/

tsipinakis:
> Surprise surprise, all my changes are gone! tada

Oh crap, sorry! totally forgot you had pushed more changes in the meantime!

bebehei:
Well, this happens.

Would you mind giving a final review?

test/queues.c
```diff
@@ -364,6 +364,65 @@ TEST test_datachange_ttl(void)
...
+TEST test_queue_stacktag(void)
+{
+        const char *stacktag = "THIS IS A SUPER WIERD STACK TAG";
+        struct notification *n1, *n2, *n3;
+
+        queues_init();
+
+        n1 = test_notification("n1", 1);
+        n2 = test_notification("n2", 1);
+        n3 = test_notification("n3", 1);
```
tsipinakis: Looking at this again the timeout parameter _does_ seem redundant. What's the point in setting it to `1` here? Any value wouldn't make a difference as we aren't advancing time.

But anyway, that's a complete nitpick everything else looks good.

bebehei: I guess this is a typo introduced via vim block mode during refactoring.

PR #554, opened 2018-10-28
Add error message for failed regex initialization

anbenson:
Also changes a type and variable name or two.

I noticed the codebase inconsistently uses bool / gboolean but since bool is used for is_initialized, I stuck to that.

### DIFF
--- a/src/menu.c
+++ b/src/menu.c
@@ -21,7 +21,7 @@
 #include "utils.h"
 
 static bool is_initialized = false;
-static regex_t cregex;
+static regex_t url_regex;
 
 struct notification_lock {
         struct notification *n;
@@ -29,29 +29,36 @@ struct notification_lock {
 };
 static gpointer context_menu_thread(gpointer data);
 
-static int regex_init(void)
+/**
+ * Initializes regexes needed for matching.
+ *
+ * @return true if initialization succeeded
+ */
+static bool regex_init(void)
 {
         if (is_initialized)
-                return 1;
+                return true;
 
         char *regex =
             "\\b(https?://|ftps?://|news://|mailto:|file://|www\\.)"
             "[-[:alnum:]_\\@;/?:&=%$.+!*\x27,~#]*"
             "(\\([-[:alnum:]_\\@;/?:&=%$.+!*\x27,~#]*\\)|[-[:alnum:]_\\@;/?:&=%$+*~])+";
-        int ret = regcomp(&cregex, regex, REG_EXTENDED | REG_ICASE);
-        if (ret != 0) {
-                LOG_W("Failed to compile regex.");
-                return 0;
+        int code = regcomp(&url_regex, regex, REG_EXTENDED | REG_ICASE);
+        if (code != 0) {
+                char error_buf[120];
+                regerror(code, &url_regex, error_buf, sizeof(error_buf));
+                LOG_W("Failed to compile URL-matching regex: %s", error_buf);
+                return false;
         } else {
                 is_initialized = true;
-                return 1;
+                return true;
         }
 }
 
 void regex_teardown(void)
 {
         if (is_initialized) {
-                regfree(&cregex);
+                regfree(&url_regex);
                 is_initialized = false;
         }
 }
@@ -73,7 +80,7 @@ char *extract_urls(const char *to_match)
         regmatch_t m;
 
         while (1) {
-                int nomatch = regexec(&cregex, p, 1, &m, 0);
+                int nomatch = regexec(&url_regex, p, 1, &m, 0);
                 if (nomatch) {
                         return urls;
                 }


src/menu.c
```diff
@@ -20,31 +20,41 @@
 #include "settings.h"
 #include "utils.h"
 
+#define ERROR_BUF_LEN 120
+
 static bool is_initialized = false;
 static regex_t cregex;
+static char ERROR_BUF[ERROR_BUF_LEN];
```
bebehei: I'd suggest using a local variable for this and without the `#define`.

```
char error_buf[120];
```

src/menu.c
```diff
@@ -20,31 +20,41 @@
...
             "[-[:alnum:]_\\@;/?:&=%$.+!*\x27,~#]*"
             "(\\([-[:alnum:]_\\@;/?:&=%$.+!*\x27,~#]*\\)|[-[:alnum:]_\\@;/?:&=%$+*~])+";
-        int ret = regcomp(&cregex, regex, REG_EXTENDED | REG_ICASE);
-        if (ret != 0) {
-                LOG_W("Failed to compile regex.");
-                return 0;
+        int code = regcomp(&cregex, regex, REG_EXTENDED | REG_ICASE);
+        if (code != 0) {
+                regerror(code, &cregex, ERROR_BUF, ERROR_BUF_LEN);
+                LOG_W("Failed to compile regex: %s", ERROR_BUF);
```
bebehei: While reading this code, isn't here an additional `is_initialized = false;` missing?

src/menu.c
```diff
@@ -20,31 +20,41 @@
...
+static char ERROR_BUF[ERROR_BUF_LEN];
 
 struct notification_lock {
         struct notification *n;
         gint64 timeout;
 };
 static gpointer context_menu_thread(gpointer data);
 
-static int regex_init(void)
+/*
```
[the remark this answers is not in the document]
anbenson: Done. Happy to use Doxygen, but I couldn't find other examples of it in the codebase to use as reference (though it was a very quick search). Let me know if something's off with the formatting.

src/menu.c
anbenson: Done.

src/menu.c
anbenson: I don't think so - `is_initialized` should already be false, do you want me to do it again to be explicit?

anbenson:
To answer your general question - this isn't strictly necessary, as in if we never change the regex / regex library, this can't cause an error, but I thought this would be helpful in the future if we change the regex to do more than just URL detection in the notification text. (Also we're already logging in case of failure, so we might as well get a human-readable error too.)

src/menu.c
bebehei: Sorry for this pointless suggestion. :see_no_evil: of course we bail out right at the beginning of the method if it's already true, so it has to be false already.

src/menu.c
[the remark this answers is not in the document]
bebehei: Search for examples with this: `git grep '\\\*\*'`

And you can use `make doc-doxygen` to generate the doxygen docs in `./doc/internal/doxygen`.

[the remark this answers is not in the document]
anbenson: Thanks for the suggestions! (Though I think you mean git grep '\/\*\*' :P )

Looks like I was just unlucky with the couple files I looked at. I think my doxygen should be okay?

src/menu.c
```diff
@@ -29,22 +29,30 @@ struct notification_lock {
...
-        int ret = regcomp(&cregex, regex, REG_EXTENDED | REG_ICASE);
-        if (ret != 0) {
-                LOG_W("Failed to compile regex.");
-                return 0;
+        int code = regcomp(&cregex, regex, REG_EXTENDED | REG_ICASE);
+        if (code != 0) {
+                size_t error_buf_len = 120;
+                char error_buf[error_buf_len];
+                regerror(code, &cregex, error_buf, error_buf_len);
+                LOG_W("Failed to compile regex: %s", error_buf);
```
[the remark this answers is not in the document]
anbenson: Done.

bebehei:
Me idiot. :see_no_evil: What about this? :joy: 

```c
                char error_buf[120];
                regerror(code, &cregex, error_buf, sizeof(error_buf));
```

> Also, I've been force-pushing amended commits to update this PR. Is there a preferred way to do this?

Well, it's actually the preferred way!

anbenson:
Nah, you're the one coming up with good suggestions :) Updated!

bebehei:
Sorry, I cannot review this PR with my mobile, but it's "approved" from my side. Love to see another PR for Hacktoberfest here!


> Nah, you're the one coming up with good suggestions

Well, I just recently had the exact same problem while firing a PR. It's a shame that I couldn't see and remember the pattern. :see_no_evil:

---

@tsipinakis Could you give me please some more permissions? IIUC, I can't generate an auth token, which can access the dunst-project orga (which is necessary to review, but not to make public comments).

anbenson:
Is the issue that you can't merge, or that you can't push to prod? Unsure why auth tokens are needed to merge this PR.

I also have a [not on Github yet] PR ready to add a couple tests, but in the interest of keeping a prettier commit graph I was waiting until this was merged to make the PR.

bebehei:
Well, I do have permissions to merge. It's just some weird corner case with my mobile app.

> I also have a [not on Github yet] PR ready to add a couple tests, 

Go on :+1:

tsipinakis:
@bebehei Re: the permissions issue, try again now it _should_ work.

bebehei:
> @bebehei Re: the permissions issue, try again now it should work.

Thanks it works. For reference, what did you change?

tsipinakis:
I approved the app. Apparently by default github doesn't allow apps to access organizations without being granted access first.

Issue #538, opened 2018-08-27
Memory leak


Strubbl:
I had run my PC incl. dunst for about a weekend. After the weekend it was using about 17 GB of memory. Unfortunately the PC is currently not running anymore. 

How can i help to find the memory leak?

### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon 1.3.2 (2018-05-06)` and also until at least 896c008def638d151517e02240f01d1198df09db
- Install type: `pacman -S dunst`
- Distro and version: `Arch Linux`

Strubbl:
History length seems to be ok:
```
$ grep history_length ~/.config/dunst/dunstrc 
    history_length = 50
```

I kept valgrind running for about 8 hours and it reached over 20 GB again. Here is the valgrind log of that session:
https://gist.github.com/Strubbl/13f1d3f4799fde853b62eb8d7746c421

I guess the process is getting this large memory footprint when i am away from the computer, which means i am not reading the dunst messages and they are kept.

bebehei:
@Strubbl Thanks for the output. Is this really everything?!? The valgrind log looks like it got truncated.

WhyNotHugo:
@Strubbl There's a link on top of the gist to the non-truncated version.

bebehei:
Oh sorry. Github gists don't display the whole file :see_no_evil: 

Could you please paste your configuration, too?

Strubbl:
yes, my config is this one: https://gist.github.com/Strubbl/375e2551c1d82dcdba659995997dcf5d

tsipinakis:
Ha! That's something I wouldn't have seen for a while since debian is still on 2.40.20.

I assume it's `librsvg`'s issue if downgrading fixes it. Best thing we could do now is try to make a test case and see how it leaks. @bebehei  since you have the problematic versions in  your package tracker (and compiling libraries is a pain) can you try out variations of the following program and see if it starts leaking?

```
```

bebehei:
Created an [upstream bugreport](https://gitlab.gnome.org/GNOME/librsvg/issues/325) and also [a reproducer](https://gitlab.gnome.org/bebehei/memleak-testproj).

Strubbl:
Thank you very much for your investigation.

tsipinakis:
This should be solved with librsvg 2.44.1.

Strubbl:
the upstream bug was reopened and it should really be solved with 2.44.2 now

PR #555, opened 2018-10-31
Add tests for extract_urls

anbenson:


anbenson:
I appreciate any comments you have. In particular, I'm wondering about the following things, if you'd be willing to take the time to answer:
- My first attempt at this PR resulted in memory leaks. Ok, sure, I didn't think about freeing memory in my tests. But when I ran valgrind locally, it pointed out memory leaks in suite_icon, which TravisCI didn't seem to catch. Do you know anything about that?
- I used g_free to free the strings (which originate from a g_strconcat call, so it seems appropriate). Is there a difference from regular free()?
- Should I not be freeing the pointers which I know to be NULL in the first two tests? It'd be safe if it were normal free(), but interested in your opinion.

tsipinakis:
Hi! Thanks for firing yet another PR. To answer your questions quickly:

> My first attempt at this PR resulted in memory leaks. Ok, sure, I didn't think about freeing memory in my tests. But when I ran valgrind locally, it pointed out memory leaks in suite_icon, which TravisCI didn't seem to catch. Do you know anything about that?

I'm not aware of that (if I was I can't seem to remember about it). Can you post the stack trace?

> I used g_free to free the strings (which originate from a g_strconcat call, so it seems appropriate). Is there a difference from regular free()?

The [glibs docs](https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html) say that `g_*` functions that allocate memory should be matched with `g_free`. A while ago we were using both `free`s interchangeably and hadn't ran into any issues but still, it's better safe than sorry as they say.

> Should I not be freeing the pointers which I know to be NULL in the first two tests? It'd be safe if it were normal free(), but interested in your opinion.

I'm not sure about that, I can come up with good arguments for both sides so I'll settle for "do as you see fit".

bebehei:
> My first attempt at this PR resulted in memory leaks. Ok, sure, I didn't think about freeing memory in my tests. But when I ran valgrind locally, it pointed out memory leaks in suite_icon, which TravisCI didn't seem to catch. Do you know anything about that?

Well, in recent times the authors of librsvg heavily change librsvg's backend. This includes also some new memory leaks (it's heaviest incarnation #538).

I see them now for about 3 months when running the testsuite locally. As they don't appear on with Travis' old librsvg version, I just ignore them.

 We actually should fix them upstream.

bebehei:
I had stitching eyes while reading the extract_urls function.

anbenson:
I certainly can, but I'll have to let this PR sit for a little while as I won't have time for a bit (maybe on the weekend?)

bebehei:
> We actually should fix them upstream.

https://gitlab.gnome.org/GNOME/librsvg/merge_requests/147

bebehei:
> https://gitlab.gnome.org/GNOME/librsvg/merge_requests/147

And it got merged, released with a bugfix release and included in ArchLinux mirrors already :heart_eyes:

bebehei:
Rebased, missing tests written. Reviewed. @tsipinakis it's your turn :wink:

PR #561, opened 2018-11-18
Drop libdbus dependency

tsipinakis:
libdbus was originally dropped with
820cfe73158b875ff6c0a00d646c7952c4cdb623 but it was never removed from
the dependency list.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -21,8 +21,7 @@ CPPFLAGS_DEBUG := -DDEBUG_BUILD
 CFLAGS_DEBUG   := -O0
 LDFLAGS_DEBUG  :=
 
-pkg_config_packs := dbus-1 \
-                    gio-2.0 \
+pkg_config_packs := gio-2.0 \
                     gdk-pixbuf-2.0 \
                     "glib-2.0 >= 2.36" \
                     pangocairo \


tsipinakis:
Eh, apparently this causes issues with querying for the services directory. Will either have to find another way to get that, or just leave it as-is.

bebehei:
Well, we have to query the dbus-1 package to actually get the dbus service path: 

https://github.com/dunst-project/dunst/blob/0410a9369bda04ab09048c464a149938f6299421/Makefile#L22-L26

bebehei:
Sorry, I misread your comment. :see_no_evil:

The services directory is specified during the dbus daemon installation. So the only place, where the true value is stored, is the package config of the dbus daemon package (-> `dbus-1`).

In my perception, the dbus daemon is a dependency (and therefore a also a build dep anyway). It's necessary to get even the most simplistic feature of dunst. It's not like our systemd support, where we want to have it optionally.

But, removing it from the `pkg_config_packs` is a wise action! We don't need it in code and therefore we don't need the flags for it!

bebehei:
@tsipinakis So, what are going to do here?

tsipinakis:
I was looking into if there was a way to avoid failing the build at least until `install` was run so that dunst could be built without dbus but it's probably too much trouble for little reward. I'll settle for removing it from `pkg_config_packs`

PR #567, opened 2018-11-27
CircleCI

### LINKED ISSUE #565: Move CI to CircleCI
I'd love to move to CircleCI.

- It offers docker container support, so there is no installation process necessary during build.
- You can test indefinitely many distros.
- You can create workflows, with dependencies.
- The current workflow in its current incarnation, is as fast as Travis, **but runs on much more Distros**.
- It's much faster giving negative results (because it tests on a single distro first).
- It's not limited to Ubuntu 14.04 (there is a **beta** for 16.04 in Travis available).
- The test environment are also available locally.
    - Container information is here: https://github.com/bebehei/dunst-docker

A first PR is already done on: bebehei/dunst#1

bebehei:
- [x] Add CircleCI to the dunst project

Fixes #565

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,8 +51,8 @@ workflows:
             - misc-doxygen
             - Alpine
       - compileandtest:
-          name: Fedora 27
-          distro: fedora27
+          name: Fedora 30
+          distro: fedora30
           requires:
             - misc-doxygen
             - Alpine
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ jobs:
       - checkout
       - run: make -j all dunstify
       - run: make -j test-valgrind
-      - run: make -j clean
+      - run: make    clean
       - run: make -j test-coverage
 
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,73 @@
+version: 2.1
+jobs:
+  misc-doxygen:
+    docker:
+      - image: bebehei/dunst:misc-doxygen-dev
+    environment:
+      SYSTEMD: 0
+      SERVICEDIR_DBUS: /tmp/none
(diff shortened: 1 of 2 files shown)


bebehei:
Afaik, i don't know. But it worked OOB on my fork.

bebehei:
Ah, I see. You have to add a deploy key for the repo.

Sorry, I can't do this. Could you please make me an org admin, too?

tsipinakis:
I did that already and removed it intentionally to see if it'd work without it. It kinda bothers me that it needs write access to the repo when we don't use any deployment functionality. But I guess everything has its downsides.

tsipinakis:
Looking at the circleCI job details for this PR it shows `bebehei/dunst` rather than `dunst-project/dunst`, it looks like it's building because you have it installed in your fork rather than actually building the PR.

Can you remove the app from your fork and do a force-push to see if it triggers? Just to make sure this works.

.circleci/config.yml
```diff
@@ -0,0 +1,73 @@
...
+      - misc-doxygen
+      - compileandtest:
+          name: Alpine
+          distro: alpine
+      - compileandtest:
+          name: Debian Stretch
+          distro: debian-stretch
+          requires:
+            - misc-doxygen
+            - Alpine
```
tsipinakis: Why test on Alpine first and then on everything else? I'd say it's more logical to do the initial tests on a distro that uses libc (debian or ubuntu preferably) and have alpine as one of the secondary builds after.

bebehei: Because it's the smallest image. Our build times are mainly controlled by the size of the docker image.

So you'll get the fastest response.

And if you wouldn't choose Alpine first, the build would fail too, but just in a later stage.

bebehei:
> Could you please make me an org admin, too?

Please!

tsipinakis:
@bebehei It's come up enough to be useful. Done! :)

PR #569, opened 2018-11-29
[WIP] Draw tests

tsipinakis:
(The start of) An attempt at implementing tests for `draw.c`. I have no clue if this is going to work at all but creating a PR anyway since I'm trying to decide how to move forward and would love some feedback.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -648,6 +648,7 @@ void draw(void)
 
 void draw_deinit(void)
 {
+        pango_font_description_free(pango_fdesc);
         output->win_destroy(win);
         output->deinit();
 }
--- a/src/draw.c
+++ b/src/draw.c
@@ -29,7 +29,7 @@ struct colored_layout {
         const struct notification *n;
 };
 
-struct window_x11 *win;
+window win;
 
 PangoFontDescription *pango_fdesc;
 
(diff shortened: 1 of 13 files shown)


bebehei:
Question: Have you evaluated Xephyr?

bebehei:
Yes, that's right. Xephyr was the wrong hint. But I think xvfb might be the right choice. I've read through the wikipedia page. Looks pormising. Although wikipedia states, that it got "slated" in 2016 (which I can't find any info on the referred citation).

Nevertheless, there's https://xpra.org/trac/wiki/Xdummy, which claims to get an X server as a non-root user with this command: `Xorg -noreset +extension GLX +extension RANDR +extension RENDER -logfile ./10.log -config ./xorg.conf :10`.

Just by the command invocation, I imply that it may even have randr support, so we could our screens section with this, too :grin: 

However, I haven't got enough knowledge and details to decide on spawning an X testserver vs writing a dummy output.

bebehei:
Any progress on this?

tsipinakis:
No, dind't get time to look enough into it but I did figure out what's the next hurdle: Pango. We depend on a lot of pango calls in draw.c and we need to figure out if the results are deterministic(read "pixel perfect") or can they change slightly. Even 1 pixel off would break the tests the way I envision it.

tsipinakis:
Uh, I accidentally deleted the branch while cleaning up my fork and github closed this automatically. Guess it's for the best since I don't know if this will ever be done.

PR #576, opened 2018-12-18
Settings (part 1/x)

bebehei:
Ok, this is the first part for #412 

It introduces string parsers with a uniform signature with a `bool` return value. This `bool` return value enhances the functions to state a fail. And enables the program to get rid of the "default" value in the parser.

### DIFF
--- a/config.h
+++ b/config.h
@@ -46,8 +46,7 @@ struct settings defaults = {
 .separator_height = 2,       /* height of the separator line between two notifications */
 .padding = 0,
 .h_padding = 0,              /* horizontal padding */
-.sep_color = SEP_AUTO,       /* SEP_AUTO, SEP_FOREGROUND, SEP_FRAME, SEP_CUSTOM */
-.sep_custom_color_str = NULL,/* custom color if sep_color is set to CUSTOM */
+.sep_color = {SEP_AUTO},     /* SEP_AUTO, SEP_FOREGROUND, SEP_FRAME, SEP_CUSTOM */
 
 .frame_width = 0,
 .frame_color = "#888888",
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -463,8 +463,8 @@ static bool dbus_get_fdn_daemon_info(GDBusConnection  *connection,
                                      char   **name,
                                      char   **vendor)
 {
-        g_return_val_if_fail(pid, false);
-        g_return_val_if_fail(connection, false);
+        ASSERT_OR_RET(pid, false);
+        ASSERT_OR_RET(connection, false);
 
         char *owner = NULL;
         GError *error = NULL;
--- a/src/draw.c
+++ b/src/draw.c
@@ -93,14 +93,14 @@ static struct color calculate_foreground_color(struct color bg)
 static struct color layout_get_sepcolor(struct colored_layout *cl,
                                         struct colored_layout *cl_next)
 {
-        switch (settings.sep_color) {
+        switch (settings.sep_color.type) {
         case SEP_FRAME:
                 if (cl_next->n->urgency > cl->n->urgency)
                         return cl_next->frame;
                 else
                         return cl->frame;
         case SEP_CUSTOM:
-                return string_to_color(settings.sep_custom_color_str);
+                return string_to_color(settings.sep_color.sep_color);
         case SEP_FOREGROUND:
                 return cl->fg;
         case SEP_AUTO:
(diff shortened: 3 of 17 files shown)


bebehei:
We often use

```
if (!pointer_param)
        return |NULL|false;
```

in the code. Small assertions, which silently return a special value on error.

My thought was that `g_return_val_if_fail` would do exactly this. But I missed, that it's also logging. You can't stop the logging. If you want to stop the logging, you have to stop the assertion check. And then the code below is broken.

The actual benefit, why I introduced `g_return_val_if_fail`: You've got it on one line and you can write visually a whole table with all exit conditions.


```c
         if(!s)
                 return false;
         if(!ret)
                 return false;
```

```c
         g_return_val_if_fail(s,   false);
         g_return_val_if_fail(ret, false);
```

And that's the reason, why I introduced `ASS_OR_RET`.

tsipinakis:
My point is that for a lot of the sanity checks at the beginning of the file `g_return_val_if_fail` fits perfectly.

I'd also spell it out completely to avoid confusing others reading the code `assert_or_return` or `assert_or_ret` would work.
(Capitalization isn't necessary if it behaves like a function)

bebehei:
> My point is that for a lot of the sanity checks at the beginning of the file g_return_val_if_fail fits perfectly.

Why?

>  (Capitalization isn't necessary if it behaves like a function)

It doesn't do at all.

tsipinakis:
Ok, I finally got time to read a bit more in this PR and justify what I'm saying:

Where you used `g_return_val_if_fail` before was a perfectly valid use IMO. Take `dbus_get_fdn_daemon_info` if `connection` or `pid` is NULL something is definitely wrong.

You were talking about the rest of the changes where you basically replaced all `if (x) return` statements with a macro and I partly disagree with that, albeit not that strongly: You're adding abstraction to save 2 lines of code in a few places. Before when someone reading the code would glance over and understand what was happening without needing to know the macro, but now they will need to look into yet another call. That isn't worth the 1 line of code we're saving.

There are a lot of places where you added `ASS_OR_RET` calls that should be converted to `g_return_val_if_fail` for most cases when you're doing standard sanity checks, but otherwise I'd opt to leave the if(...) returns as-is.

bebehei:
During the last week, I started three times writing an answer. And I changed three times my mind, whether I'd prefer `g_val_return_if_fail` or a custom macro like `ASS_OR_RET`. Spoiler: Both of them have their pros and cons.

On the one hand side, `g_return_val_if_fail` might be the correct solution. Printing out a warning is often a good idea. And mostly having a failing assertion is a bad thing. So in combination with some env variables, you can abort the program with a coredump and debug it right in place. This would be especially useful in our testsuite. But it's absolutely contradictory. We suppress all our warnings during the test.

And in a release mode, printing a warning when failing the `g_return_val_if_fail` condition is useless. But also you can't just disable the assertion logging. You only can stop the whole assertion (which we heavily rely on).

Using `ASS_OR_RET` would have the improvement that it's checking much faster and would fail silently. It wouldn't bug the user and would use the default failover (which is IMO absolutely ok).

But actually, why I can't decide about any style is this example code:

```c
bool string_parse_urgency(const char *s, enum urgency *ret)
{
        ASS_OR_RET(s, false);
        ASS_OR_RET(ret, false);
```

We should actually distinguish the imporance of the `s` param and the `ret` param. Usually `s` is a parsed string of the configuration file (where `NULL` is explicitly an allowed value) and `ret` points to a local stack entry. In case of the `NULL` value, this would be absolutely fine. In case of the `ret` value, this would _indicate_ a mistake.

So, the important question is: Why should be the following example our preferred statement?

```c
bool string_parse_urgency(const char *s, enum urgency *ret)
{
        if (!s)
                return false;
        g_return_val_if_fail(ret, false);
```

It prints out a warning for messages, which may indicate an error and still returns silently on values of the string. It works perfectly

But I think the code is actually very ugly this way. Especially the reasoning won't be clear in the future.

tsipinakis:
There's some conflicts here, care to rebase before I review @bebehei ?

bebehei:
> There's some conflicts here, care to rebase before I review @bebehei ?

done.

tsipinakis:
> But I think the code is actually very ugly this way. Especially the reasoning won't be clear in the future.

When you put it that way I agree with your points. One thing I would change is spell out the macro entirely as `ASSERT_OR_RET`. It's still shorter than `g_return_val_if_fail` and gets the point across much better IMO.

bebehei:
@tsipinakis Valid point. Update done.

PR #585, opened 2019-01-05
Manually convert GdkPixbuf to cairo surfaces

bebehei:
As already noted in the code comment, we all knew that the in memory
conversion to a PNG stream is a cumbersome technique. But it was worth
to remove the GTK dependency.

After finding out, that gdk_pixbuf_new_from_data can actually use a
GVariant and share its memory without allocating new memory, I poked
the cairo docs for a similar technique.

There is cairo_image_surface_create_for_data, which would achieve the
same, but cairo uses another color byte order. So we have to manually
copy and reassemble the bytes.

It may look like a big rise in complexity, but actually, it's straight
forward.

And the gdk_pixbuf_to_cairo_surface is 33x faster.

### DIFF
--- a/src/icon.c
+++ b/src/icon.c
@@ -1,5 +1,6 @@
 #include "icon.h"
 
+#include <assert.h>
 #include <cairo.h>
 #include <gdk-pixbuf/gdk-pixbuf.h>
 #include <stdbool.h>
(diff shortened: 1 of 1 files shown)


bebehei:
@tsipinakis I guess, you forgot this one in the flood of all PRs :wink: Especially in this PR, I'd like to have a review. :grin:

src/icon.c
```diff
@@ -15,38 +16,96 @@ static bool is_readable_file(const char *filename)
...
+                                iter_p += n_channels;
+                        }
+                }
+        } else {
+                for (int h = 0; h < height; h++) {
+                              unsigned char *iter_c = pixels_c + h * rowstride_c;
+                        const unsigned char *iter_p = pixels_p + h * rowstride_p;
+                        for (int w = 0; w < width; w++) {
+                                double alpha_factor = iter_p[3] / (double)0xff;
+                                iter_c[CAIRO_R] = (unsigned char)(iter_p[0] * alpha_factor + .5);
```
tsipinakis: Can you explain `alpha_factor` and why you add `.5` here?

bebehei: Docs are here:

https://cairographics.org/manual/cairo-Image-Surfaces.html#cairo-format-t

I've actually taken [this gist](https://gist.github.com/bert/985903/1f7675464104c70fb65ed93328b7538f459775dc#file-draw_in_a_gdkpixbuf_with_cairo_and_pango-c-L104-L210) in my first iteration.

PR #586, opened 2019-01-07
Checksummed icons

### LINKED ISSUE #571: Duplicate notifications from kdeconnect not merged
I'm using kdeconnect to get notifications from my phone to my desktop. For some reason, only the notifications for calls are no deduped by dunst. Some debugging data:
- [Output of `dunst -p`](https://gist.github.com/infokiller/407c6a24c4ca5657c9f5b69a596ae3bb)
- [Output of `dbus-monitor path=/org/freedesktop/Notifications`](https://gist.github.com/infokiller/407c6a24c4ca5657c9f5b69a596ae3bb)- note that I removed the image data which is very big.

When I'm removing the image data (by removing it from the calling contact), deduping works.
Is deduping not applied when there is image data?

### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon 1.3.2 (2018-05-06)`
- Install type: Official Arch Linux package
- Distro and version: `Arch Linux`

bebehei:
Fixes #571

### DIFF
--- a/src/icon.c
+++ b/src/icon.c
@@ -222,9 +222,4 @@ GdkPixbuf *get_pixbuf_from_raw_image(const struct raw_image *raw_image)
         return pixbuf;
 }
 
-cairo_surface_t *icon_get_for_notification(const struct notification *n)
-{
-        return gdk_pixbuf_to_cairo_surface(n->icon);
-}
-
 /* vim: set tabstop=8 shiftwidth=8 expandtab textwidth=0: */
--- a/src/icon.c
+++ b/src/icon.c
@@ -228,8 +228,8 @@ cairo_surface_t *icon_get_for_notification(const struct notification *n)
 
         if (n->raw_icon)
                 pixbuf = get_pixbuf_from_raw_image(n->raw_icon);
-        else if (n->icon)
-                pixbuf = get_pixbuf_from_icon(n->icon);
+        else if (n->iconname)
+                pixbuf = get_pixbuf_from_icon(n->iconname);
         else
                 return NULL;
 
--- a/src/icon.c
+++ b/src/icon.c
@@ -110,6 +110,33 @@ cairo_surface_t *gdk_pixbuf_to_cairo_surface(GdkPixbuf *pixbuf)
         return icon_surface;
 }
 
+GdkPixbuf *icon_pixbuf_scale(GdkPixbuf *pixbuf)
+{
+        ASSERT_OR_RET(pixbuf, NULL);
+
+        int w = gdk_pixbuf_get_width(pixbuf);
+        int h = gdk_pixbuf_get_height(pixbuf);
+        int larger = w > h ? w : h;
+        if (settings.max_icon_size && larger > settings.max_icon_size) {
+                int scaled_w = settings.max_icon_size;
+                int scaled_h = settings.max_icon_size;
+                if (w >= h)
+                        scaled_h = (settings.max_icon_size * h) / w;
+                else
+                        scaled_w = (settings.max_icon_size * w) / h;
+
+                GdkPixbuf *scaled = gdk_pixbuf_scale_simple(
+                                pixbuf,
+                                scaled_w,
+                                scaled_h,
+                                GDK_INTERP_BILINEAR);
+                g_object_unref(pixbuf);
+                pixbuf = scaled;
+        }
+
+        return pixbuf;
+}
+
 GdkPixbuf *get_pixbuf_from_file(const char *filename)
 {
         char *path = string_to_path(g_strdup(filename));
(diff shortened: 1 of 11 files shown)


src/icon.c
```diff
@@ -205,19 +205,120 @@ GdkPixbuf *get_pixbuf_from_icon(const char *iconname)
...
+        unsigned char *data_chk = g_malloc(data_chk_len);
+        size_t rowstride_short = pixelstride * width;
+
+        for (int i = 0; i < height; i++) {
+                memcpy(data_chk + (i*rowstride_short),
+                       data_pb  + (i*rowstride),
+                       rowstride_short);
+        }
+
+        *id = g_compute_checksum_for_data(G_CHECKSUM_MD5, data_chk, data_chk_len);
```
tsipinakis: How about truncating the hash and storing it in a 64 bit int? Faster comparison, slightly bigger chance of collision.
We'll have to also hash the icons that come from the filesystem though.

src/notification.c
```diff
@@ -365,25 +381,10 @@ void notification_init(struct notification *n)
...
-        g_return_if_fail(n);
-
-        g_clear_object(&n->icon);
-
-        if (n->raw_icon)
-                n->icon = get_pixbuf_from_raw_image(n->raw_icon);
-        else if (n->iconname)
-                n->icon = get_pixbuf_from_icon(n->iconname);
-
-        n->icon = icon_pixbuf_scale(n->icon);
```
tsipinakis: Uhm, where did the icon scaling logic go?

src/icon.c
bebehei: > For that we cache the pixel buffer and introduce a field, which saves
the current icon's id. The icon_id may be a path or a hash.
So you can compare two notifications by their icon_id field regardless
of their icon type by their icon_id field.


https://github.com/dunst-project/dunst/blob/0886f926cc151937c125ddcd31adef15959d6a32/src/notification.h#L50-L54

src/notification.c
bebehei: Good point. I deleted this right before opening this PR by accident. :see_no_evil:

src/icon.c
tsipinakis: Yes I got that, I proposed changing `icon_id` to a 64 bit int and hashing all icons regardless. The gains would be minimal thinking more about it so not sure how well it'd work out.

bebehei: > How about truncating the hash and storing it in a 64 bit int? Faster comparison

But initially you have to hash the whole image. For making a comparison faster, whcih would occur maybe, this would be a no brainer.

Look, we have to have `appname`, `body` and `summary` the same, before we test for icon equality.

https://github.com/dunst-project/dunst/blob/0886f926cc151937c125ddcd31adef15959d6a32/src/notification.c#L178-L185

 And the `STR_EQ(a->icon_id, b->icon_id)` will occur roughly a single time for a notification.

src/icon.c
```diff
@@ -205,19 +205,120 @@ GdkPixbuf *get_pixbuf_from_icon(const char *iconname)
...
+         * row n:   |   data for row n    |
+         */
+
         GdkPixbuf *pixbuf = NULL;
+        GVariant *data_variant = NULL;
+        unsigned char *data_pb;
+
+        gsize len_expected;
+        gsize len_actual;
+        gsize pixelstride;
```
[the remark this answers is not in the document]
bebehei: Actually both should be a `gsize` or `size_t`. But the gdk pixbuf requires an int and the `GVariant` stores an int for the rowstride.

src/notification.c
```diff
@@ -328,10 +346,12 @@ void notification_init(struct notification *n)
...
 
         /* Icon handling */
-        if (STR_EMPTY(n->icon))
-                g_clear_pointer(&n->icon, g_free);
-        if (!n->raw_icon && !n->icon)
-                n->icon = g_strdup(settings.icons[n->urgency]);
+        if (STR_EMPTY(n->iconname))
+                g_clear_pointer(&n->iconname, g_free);
+        if (!n->icon && n->iconname)
+                n->icon = icon_get_for_name(n->iconname, &n->icon_id);
```
tsipinakis: Missed one here, this call doesn't scale the icon.

bebehei: OMG. Yes.

bebehei: > EDIT: I think this is a hint to maybe put all the icon parsing in one place to be able to manage them.

Valid point. This already exists. [I just had to use the right function](https://github.com/dunst-project/dunst/compare/5748c41124a9523f26414720da3df601d52d8611..6f8b53c4e8e0edc45758316dc3ccf006f931f86b).

PR #588, opened 2019-01-13
Add a tip to the docs about how to skip initial display

mgsloan:
My usecase is that I don't want to get notified for every new song
that plays (it's distracting), but I do want to bring them up via the
notification history.

### DIFF
--- a/docs/dunst.pod
+++ b/docs/dunst.pod
@@ -645,6 +645,10 @@ together. The default stack_stag value is set from the string hints
 "synchronous", "private-synchronous", "x-canonical-private-synchronous", and
 "x-dunst-stack-tag".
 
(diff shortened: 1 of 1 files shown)


mgsloan:
I considered adding an option for this, but this hack seems sufficient for my purposes.  The main case where a more explicit option might be helpful is if `sticky_history` is set to `false`.  In that case, history for the item will be relatively useless assuming it uses the same timeout.

tsipinakis:
The docs aren't getting much love so any addition is very welcome.

> In that case, history for the item will be relatively useless assuming it uses the same timeout.

This can probably be solved by implicitly enabling sticky history for all recalled notifications with a very short timeout (<500ms?). But as it is I wouldn't bother with it unless it comes up.

PR #590, opened 2019-01-16
Skip display rule option

mgsloan:
The hack described in https://github.com/dunst-project/dunst/pull/588 - setting `timeout = "1ms"` in a rule to skip display - doesn't always work.  Sometimes I see the notification for a brief flash.

There could potentially be a different workaround for this, say, skipping enqueue to the waiting list if the timeout is sufficiently low.

I've gone ahead and implemented an explicit option.  This will also make things work better in the case that `sticky_history` is set to `false`.  It also means that if `DUNST_COMMAND_PAUSE` has been used, when it gets unpaused it won't need to work its way through a queue of waiting notifications with tiny timeout.

### DIFF
--- a/src/queues.c
+++ b/src/queues.c
@@ -133,6 +133,9 @@ static bool queues_notification_is_finished(struct notification *n, struct dunst
 {
         assert(n);
 
+        if (n->skip_display && !n->redisplayed)
+                return true;
+
         if (n->timeout == 0) // sticky
                 return false;
 
@@ -428,7 +431,12 @@ void queues_update(struct dunst_status status)
                 notification_run_script(n);
 
                 g_queue_delete_link(waiting, iter);
-                g_queue_insert_sorted(displayed, n, notification_cmp_data, NULL);
+
+                if (n->skip_display && !n->redisplayed) {
+                        queues_history_push(n);
+                } else {
+                        g_queue_insert_sorted(displayed, n, notification_cmp_data, NULL);
+                }
 
                 iter = nextiter;
         }
--- a/src/queues.c
+++ b/src/queues.c
@@ -430,11 +430,10 @@ void queues_update(struct dunst_status status)
                 n->start = time_monotonic_now();
                 notification_run_script(n);
 
-                g_queue_delete_link(waiting, iter);
-
                 if (n->skip_display && !n->redisplayed) {
-                        queues_history_push(n);
+                        queues_notification_close(n, REASON_USER);
                 } else {
+                        g_queue_delete_link(waiting, iter);
                         g_queue_insert_sorted(displayed, n, notification_cmp_data, NULL);
                 }
 
--- a/test/queues.c
+++ b/test/queues.c
@@ -190,6 +190,50 @@ TEST test_queue_notification_close_histignore(void)
         PASS();
 }
 
+TEST test_queue_notification_skip_display(void)
+{
+        struct notification *n;
+
+        // Test skipping display
+        n = test_notification("n", -1);
+        n->skip_display = true;
+
+        queues_init();
+        queues_notification_insert(n);
+        QUEUE_LEN_ALL(1, 0, 0);
+        queues_update(STATUS_NORMAL);
+        QUEUE_LEN_ALL(0, 0, 1);
+        queues_teardown();
+
+        PASS();
+}
+
+TEST test_queue_notification_skip_display_redisplayed(void)
+{
+        struct notification *n;
+
+        // Test skipping display
+        n = test_notification("n", -1);
+        n->skip_display = true;
+
+        queues_init();
+        queues_notification_insert(n);
+        QUEUE_LEN_ALL(1, 0, 0);
+        queues_update(STATUS_NORMAL);
+        QUEUE_LEN_ALL(0, 0, 1);
+
+        queues_history_pop();
+        QUEUE_LEN_ALL(1, 0, 0);
+        queues_update(STATUS_NORMAL);
+        QUEUE_CONTAINSm("A skip display notification should stay in displayed "
+                        "queue when it got pulled out of history queue",
+                        DISP, n);
+
+        queues_teardown();
+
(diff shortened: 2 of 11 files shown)


bebehei:
On my mobile, but some thoughts:

- There's no `NotificationClosed` signal sent for notifications with `skip_display` set.
- The queue length of `history` isn't respected.

But I definitele appreciate your new PR!

tsipinakis:
:+1: As I mentioned in the previous issue, perhaps making this happen implicitly if the timeout is really short could save us adding another knob. @bebehei what do you think?

mgsloan:
Yeah I'd be fine with `timeout == 1` implying skipping display.  It's a bit funky, would be better if it was `timeout == 0`, but that already has a special meaning.

Certainly can understand the desire to reduce the number of knobs.  On the other hand, that would increase the number of particularly distinguished values on one of the knobs - maybe that's preferred?

mgsloan:
@bebehei Thanks for reviewing.  I believe I've addressed your feedback in subsequent commits.

I think it's easier to review with incremental commits, but let me know if I should squash it into one (though I think it's best to keep the revert separate).

test/queues.c
```diff
@@ -190,6 +190,25 @@ TEST test_queue_notification_close_histignore(void)
...
+{
+        struct notification *n;
+
+        // Test skipping display
+        n = test_notification("n", -1);
+        n->skip_display = true;
+
+        queues_init();
+        queues_notification_insert(n);
+        QUEUE_LEN_ALL(0, 0, 1);
```
bebehei: Honestly, I'm missing another check for the signal.

mgsloan: How do we check for the signal?  Note that this code is exercising the queues, not end-to-end dbus functionality (the signal is done in `dbus.c` not `queues.c`).  I don't see any end-to-end tests that also involve rules.

mgsloan:
> What shall we do with notifications having a stack_tag, or qualify for stack_duplicates or have a replacement id. Skip_display behaves inconsistent.

Good question.  I considered this, but at the time thought it was reasonable that if a notification is being displayed, then stacking a skip_display notification atop it should continue being displayed.

However, I can see it being confusing if a notification that has a skip_display rule ends up being displayed in this case.  I don't think there is necessarily a right answer here, but I also don't think it's worth making the knob more complicated to allow both semantics.

I can certainly try adding a commit that implements latter semantics, but I don't see a way to implement it without making the code a fair bit more complicated.

mgsloan:
@bebehei Ping!  As far as I can tell the next steps here are:

1. Determine how `skip_display` and notification stacking should interact

2. How to test signaling of issue close, particularly when the test is unit testing the queues which is not responsible for sending the close signal.  At least, this is how I'm interpreting your comment on the test code.

bebehei:
> Determine how skip_display and notification stacking should interact

This would be my turn https://github.com/dunst-project/dunst/compare/master...bebehei:pr590. You can take the code fully.

Every replacement case is covered by: Just replace it as usual, but vanish the notification right with the next update. see the docs from `queues_notification_insert()`:

https://github.com/dunst-project/dunst/blob/711b11c92f3d304fb21c77b882adac45d90528e1/src/queues.h#L126-L127

> How to test signaling of issue close, 

Sorry, I've sent you into the wrong direction. The basics are actually testable. But in conjunction together, it's too tricky to get tested right now.

mgsloan:
@bebehei LGTM, thanks for making those changes.  They seem to work well.  I've rebased atop master and pushed to my branch.

I think this is ready for merge, right?

mgsloan:
@tsipinakis Ping!  Looks like @bebehei is requesting that you review this.  Thanks!

tsipinakis:
Opps, I completely missed the notification for this, sorry for the delay!

I did a review then realised that when pulling @bebehei 's tests you accidentally rolled back the commit fixing the initial mistakes.
e.g. There's still no `NotificationClosed` signal.

bebehei:
>  e.g. There's still no `NotificationClosed` signal.

The signal is there. But there are no tests for it. Right now the signals aren't feasible to test.

tsipinakis:
With 977a7a7ea47f2d86ea3168e609576ebeb0ffa719 I don't see a NotificationClosed signal when `skip_display` is used. Am I missing something?

bebehei:
Ouch, `queues_history_push` is the wrong method. I should have proposed `queues_notification_close` :see_no_evil:

mgsloan:
@tsipinakis @bebehei Ah, good catch!  I've pushed a commit that fixes the problem, and have verified via `dbus-monitor` that the notification close is sent.  I also moved an invocation of `g_queue_delete_link`, because `queues_notification_close` only works if the notification is still on the waiting queue.

src/queues.c
```diff
@@ -133,6 +133,9 @@ static bool queues_notification_is_finished(struct notification *n, struct dunst
 {
         assert(n);
 
+        if (n->skip_display && !n->redisplayed)
+                return true;
```
tsipinakis: This hunk seems redundant, the notifications never see the displayed queue but skip directly from waiting to history. Was there a reason for adding this?

mgsloan:
@tsipinakis Ping!  (No rush, just wanted to check if this had fallen off your radar)

tsipinakis:
It hasn't, was just waiting for @bebehei to respond but it looks like he's busy. I'll go ahead and merge this as it's not an actual issue and as you said some redundancy is a good thing.

mgsloan:
Great, thanks a bunch for reviewing, y'all!

src/queues.c
bebehei: It may seem redundant, but actually it isn't. For the case, that an already displayed notification gets replaced, this would trigger the immediate removal.

bebehei:
> It hasn't, was just waiting for @bebehei to respond but it looks like he's busy

Yeah, I had been a few days off. Thanks for merging and caring here.

PR #592, opened 2019-01-23
Desktop entry filters.

### LINKED ISSUE #470: Add support for desktop-entry hint filter
GApplications set the `desktop-entry` hint in meaning of a real UID. The appname of every GApplication is set to an empty string. This makes sense, as the appname sent is  depending on the current locale and may get translated.

See https://bugzilla.gnome.org/show_bug.cgi?id=766882

bebehei:
Adds another filter type. Especially this filter isn't localised. So it's much better than the `appname` filter.

Fixes #470


src/settings.c
```diff
@@ -724,6 +724,7 @@ void load_settings(char *cmdline_config_path)
                 r->history_ignore = ini_get_bool(cur_section, "history_ignore", r->history_ignore);
                 r->match_transient = ini_get_bool(cur_section, "match_transient", r->match_transient);
                 r->set_transient = ini_get_bool(cur_section, "set_transient", r->set_transient);
+                r->desktop_entry = ini_get_string(cur_section, "desktop-entry", r->desktop_entry);
```
tsipinakis: We're using underscores for separating words in all other rules. Maybe it's best to keep it consistent and do the same here.

bebehei:
All typos fixed.

PR #597, opened 2019-01-29
[WIP] Add initial doc

bebehei:
Read the docs here: https://dunsttestproject.readthedocs.io/en/docs/

As there are some distros out, which will have dunst 1.3 for a longer time, I'm currently starting to write the docs for our current minor release 1.3. The overhead of master and 1.3 is actually quite small, because the hardest work is actually creating the docs. But in favor of this, we have always a good working 1.3 docs copy.

So, I've written a few TODO-Marks into the rst files. If there's something "inconsistent" below the mark, just ignore it. Thanks

- [ ] go over the text with `aspell` or something similar
- [ ] convert, split and rewrite the manpages and generate them via `sphinx`
- [ ] document the `dunstrc` on the rtd
    - [ ] add a mechanism to mark config options old/deprec and "available since"
- [ ] Rewrite the project's readme
    - [ ] Remove the compilation instructions
    - [ ] Convert to rst and include in `docs/index.rst`
    - [ ] Add some examples and make it actually a "showcase" page
- [ ] Add a support page
- [ ] evaluate, wheter there's compatibility with doxygen

One **big** point is actually rewriting the readme. I would like to convert it to something user-friendly. Currently it's dev-only friendly. But we both don't need this information. The readme is the landing page of our project and I think we should provide there information covering those questions: "What can dunst do?" "How can it look like?" "Why do I (not) need this?"

----

If you've got some ideas. If you want to write docs, too, you can push it here (but please notify me in a short comment).

I would love any input.


docs/getting_started.rst
```diff
@@ -0,0 +1,45 @@
...
+Open your terminal and start the dunst **server** for validation::
+
+    dunst -startup_notification
+
+A small window should pop up saying ``dunst is up and running`` for at least a few seconds.
+
+Relogin to your session
+-----------------------
+
+Now, after you've installed and made sure that dunst is able to start, please log out and log in again. Dunst ist now running and receives notifications.
```
tsipinakis: > Now, after you've installed and made sure that dunst is able to start, please log out and log in again.

Why? Dunst is running, there is nothing to reset here.

docs/packages.rst
```diff
@@ -0,0 +1,21 @@
...
+You can install dunst easily just by installing the ``dunst`` package. Here are some distros with prepackaged dunst:
+
+.. keep this list in alphabetical order
+
+============ ======================== =============================================================================
+Distribution Command                  Source
+============ ======================== =============================================================================
+Arch Linux   ``pacman -S dunst``      `[community] <https://www.archlinux.org/packages/community/x86_64/dunst/>`_
+Arch Linux   ``yay -S aur/dunst-git`` `AUR <https://aur.archlinux.org/packages/dunst-git>`_
+Debian       ``apt install dunst``    `Debian packages <https://packages.debian.org/sid/dunst>`_
```
tsipinakis:
```suggestion
Debian       ``apt install dunst``    `Debian packages <https://tracker.debian.org/pkg/dunst>`_
```

docs/getting_started.rst
bebehei: My point was, that there are between the installation of the package `dunst` and a working dunst instance many factors.

And one common factor, is that there's either no DBus at all. (wrong initialised session, etc.) or there's another notification daemon already running.

And at this point, we assured that both problematic conditions are solved (I see, I haven't added some docs how to handle a competing notification daemon). So we can give DBus the responsibility to start the service. Because if DBus can't or doesn't start our daemon, we do not have good logs or commands to debug it.

It's to assure that it can work. That's it.

----

But yeah actually that's a one reason, why I wanted to have a review. If that's counterintuitive for you. I'd definitely like to improve it. Getting dunst to run is the hardest part and we only can solve it by writing good docs.

I also appreciate radically different approaches!

tsipinakis:
@bebehei What's the status of this? I'd like to have 1.4 released within the next few days so either we marathon this one or skip it for now. Since this is mostly for readthedocs which is online we can simply add the docs later and point the users there through the website after the fact.

tsipinakis:
ping @bebehei ?

bebehei:
Sorry for the huge delay in answering.

 I'd love to finish the docs **before** release. Can we set march 1st for release? I'm sure to have them finished until then.

tsipinakis:
Sounds good to me, bumped the milestone deadline to March 1st.

tsipinakis:
>  document the dunstrc on the rtd 

Having both an example `dunstrc` and setting documentation on rtd(and manpage) seems counter-intuitive to me.

We have to manually keep them in sync and it can cause confusion when one setting is explained in more detail in one place in the docs vs dunstrc and vice-versa. Take the `geometry` setting for example, much more detailed in the docs and just a basic outline in the dunstrc.


Additionally, we recommend copying the example dunstrc and using that as a base to customize the settings which makes things complicated when new settings are added between releases since we obviously cannot updated existing dunstrcs and users will have to seek the new settings addition in the docs to find them.


I'd suggest we simply remove the example dunstrc and replace it with a bare minimum working configuration file.
In it's place we should add detailed explanation of every available setting in the documentation along with use cases and working examples and let users figure out the config from that.

tsipinakis:
@bebehei Can't seem to push to the docs branch, have you enabled maintainer access?

Anyway, I transferred the `configuration` section of the man-page see https://github.com/dunst-project/dunst/commit/76a502114e63f348755641791969154e3b73d9b0

tsipinakis:
I took a look into using sphinx to build the manpages but spotted a few hurdles:

* Currently a big part of the docs is how to obtain and install dunst. In the manpage we obviously need to somehow cut that part out as those who are reading it already have dunst installed. 
Perhaps there is a way to specify a different index file?

* I have added a quick table of contents in the `configuration.rst` file so that when someone looks at the docs online they can quickly jump between settings but since manpages don't allow for such links we need a way to cut that out only for the man format. I'm not sure if that's possible with sphinx.

tsipinakis:
@bebehei ping?

tsipinakis:
This needs some more discussion/work and @bebehei seems inactive (ping again?).

I intend to release 1.4 this weekend, this is the last blocker so I'll bump it to 1.5.

fwSmit:
It seems like there is quite some documentation here. Any plans for merging or reusing this any time soon?

bebehei:
Actually this work is about 2 years old. I haven't touched since.

I appreciate any plans in resuing/merging/updating this PR.

fwSmit:
Don't think this will be merged any time soon. Most of this is already present in other parts of the documentation. Feel free to reopen if you want to continue this

Issue #349, opened 2017-08-04
Transitioning rules to a more consistent naming scheme


tsipinakis:

bebehei:
Are there any words which represent `match` and `set` and have an equal length? For the readability in your dunstrc this might be a very good choice.

bebehei:
> Are there any words which represent match and set and have an equal length?

I would propose `has_` and `set_`:

```ini
[critical]
    has_body = "backup failed"
    has_summary = "backup"
    
    has_urgency = "normal"
    set_urgency = "critical"
```

hobarrera:
I've been thinking rules could have a much more english-like format, something like:

```
match message "backup failed" set urgency "critical"
match message "backup successful" set urgency "normal"
match message "backup something else" set transient on
```

The language for this is basically:

```
match [condition value]+ set [attribute value]+
```

This is really based on pf and opensmtpd's syntax format, which are really readable.

The upsides are:
* One line/statement per rule.
* Much more readable.
* Extendable
* Large amount of rules are easier to handle [when maintaining a conf file[.

Downsides:
* Not an ini-like format.
* Parser may be slightly more non-trivial

tsipinakis:
If we're going to go this far with the matching logic, we might as well implement a full on filter language like [newsbeuter](http://newsbeuter.org/doc/newsbeuter.html#_filter_language).

> One line/statement per rule.

I'm not sure this is an upside, having an option for a rule to be a single line would be good, but for a more complicated rule with multiple matches and assignments it can be much better to split it over multiple lines.

> Not an ini-like format.

This is a big downside, it breaks the current configuration format meaning that the config parser will have to be reworked and that it breaks the ability for external tools to parse the config file if needed (for example I was just looking into using pythons ini capability to attempt to write a script to generate test cases).

One workaround to this (albeit a bit ugly) might be to split rules into a different file(e.g. `dunst.rules`).

hobarrera:
> I'm not sure this is an upside, having an option for a rule to be a single line would be good, but for a more complicated rule with multiple matches and assignments it can be much better to split it over multiple lines.

Yeah, the main point was once *statement* per rule.

tsipinakis:
There hasn't been any work done for this and haven't really decided yet how to handle it. I'm bumping this to 1.5.

fedang:
i will close this in favour of #1333

PR #598, opened 2019-01-31
Allow rules to override icon size via set_max_icon_size

mgsloan:
This PR isn't ready yet, it has the following deficiencies:

- [ ] No documentation changes
- [ ] No tests
- [ ] Initialization of icon pixbuf is ugly and inefficient because I think it will now get done twice - once in `notification_init` and again after the call to `notification_init` in `dbus_message_to_notification`.  The reason the involved code moved is so that the call to `notification_icon_replace_data` would run after the rules, and so see the overridden max icon size.

I'm putting it up to get feedback as to whether this should be added to dunst, and possibly if a different name should be used (just `max_icon_size`?)

Motivation for this change in my case is that I'd like most icons to be 64 pixels, but prefer 256 pixel icons for spotify, since it uses the album art for the icon.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -238,15 +238,6 @@ static struct notification *dbus_message_to_notification(const gchar *sender, GV
                 g_variant_unref(dict_value);
         }
 
-        dict_value = g_variant_lookup_value(hints, "image-data", G_VARIANT_TYPE("(iiibiiay)"));
-        if (!dict_value)
-                dict_value = g_variant_lookup_value(hints, "image_data", G_VARIANT_TYPE("(iiibiiay)"));
-        if (!dict_value)
-                dict_value = g_variant_lookup_value(hints, "icon_data", G_VARIANT_TYPE("(iiibiiay)"));
-        if (dict_value) {
-                notification_icon_replace_data(n, dict_value);
-                g_variant_unref(dict_value);
-        }
 
         /* Check for transient hints
          *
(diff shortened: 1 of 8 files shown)


mgsloan:
Cool, makes sense!  Do you think it would be good to have the rule setting have the same name as the global config?  Or perhaps it is nice that `set_...` implies overriding.

- [ ] Use n->max_icon_size = -1 as a special value.  (I agree with that decision)

- [ ] Make sure the icon code behaves reasonably for negative or small values

> But actually, I'd love to see PRs, which build. I'd recommend the target set of make clean all test-valgrind doc-doxygen to check, whether it builds your local machine.

Ah, good point!

- [ ] Fix doxygen build

bebehei:
> Cool, makes sense! Do you think it would be good to have the rule setting have the same name as the global config? Or perhaps it is nice that set_... implies overriding.

Well, actually there is #349, but nobody had any comment or "final" resolution on it. Do you have any ideas? Yet, there's no necessity to stick with `set_XY` or whatsoever.

mgsloan:
Status update for this PR:  I ended up procrastinating following up on this quite a bit.  At the moment I don't feel like finishing it up.  If someone else wants to take it past the end 

@bebehei RE #349 sticking with `set_XY` makes sense to me.  It would be nice to have consistency and uniformity.  As for having a more sophisticated predicate language, I'd almost go as far as suggesting that the config should be a C file - hah!  That wouldn't be very friendly to non-developers though

fwSmit:
Closing this, since it's very old and likely not gonna be finished

PR #601, opened 2019-02-03
Remove trailing spaces from MANPREFIX path

rhssk:
Variables with trailing comments include the spaces leading up to the
comment (https://www.gnu.org/software/make/manual/make.html#Flavors)
so the corresponding install command looked like
"install -Dm644 docs/dunst.1 $(DESTDIR)/usr/share/man /man1/dunst.1"

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -2,7 +2,8 @@
 PREFIX ?= /usr/local
 BINDIR ?= ${PREFIX}/bin
 DATADIR ?= ${PREFIX}/share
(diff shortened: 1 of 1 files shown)


tsipinakis:
Yikes, so much for my review in #600. Thanks for the quick find & fix!

bebehei:
OMG. Thank you for reporting. Such a stupid mistake, I made. :see_no_evil:

PR #603, opened 2019-02-03
Test installation

tsipinakis:
Ensure that dunst is installed correctly to avoid bugs like #601.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,6 +24,7 @@ jobs:
       - checkout
       - run: make CC=clang -j all dunstify test/test
       - run: make CC=clang -j test-valgrind
+      - run: ./test/test-install.sh
       - run: make             clean
       - run: make CC=gcc   -j all dunstify test/test
       - run: make CC=gcc   -j test-valgrind
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ before_install:
 
 script:
   - CFLAGS="-Werror" make all dunstify test-valgrind doc-doxygen
+  - ./test/test-install.sh
   - CFLAGS="-Werror" make clean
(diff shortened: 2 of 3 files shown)


tsipinakis:
> /bin/sh: ./test/test-install.sh: not found
Exited with code 127

I assume this is because it can't find /bin/bash. And since I'm using the `<()` bash-ism I don't see how to translate that to sh.
@bebehei Can you add bash to the docker images?

bebehei:
So, while I updated the docker images, I also have incorporated some changes to the file. Two mistakes, I found:

- `SYSTEMD` wasn't set.
- The script couldn't get executed in from random locations.

```bash
#!/usr/bin/env bash

set -euo pipefail

BASE="$(dirname "$(dirname "$(readlink -f "$0")")")"
PREFIX="${BASE}/install"

make -C "${BASE}" SYSTEMD=1 SERVICEDIR_SYSTEMD="${PREFIX}/systemd" SERVICEDIR_DBUS="${PREFIX}/dbus" PREFIX="${PREFIX}" install

diff -u <(find "${PREFIX}" -type f -printf "%P\n" | sort) - <<EOF
bin/dunst
dbus/org.knopwob.dunst.service
share/dunst/dunstrc
share/man/man1/dunst.1
systemd/dunst.service
EOF

make -C "${BASE}" SYSTEMD=1 SERVICEDIR_SYSTEMD="${PREFIX}/systemd" SERVICEDIR_DBUS="${PREFIX}/dbus" PREFIX="${PREFIX}" uninstall

if ! [ -z "$(find "${PREFIX}" -type f)" ]; then
        echo "Uninstall failed, following files weren't removed"
        find "${PREFIX}" -type f
        exit 1
fi
```

tsipinakis:
Updated

PR #605, opened 2019-02-07
Changes DBus name owner flag from None to Replace

petrblaho:
As some desktop environments do not support swapping out
their notification daemon
(https://gitlab.gnome.org/GNOME/gnome-shell/issues/99)
we need to use G_BUS_NAME_OWNER_FLAGS_REPLACE flag that
allows Dunst to take org.freedesktop.Notifications name
for itself.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -591,7 +591,7 @@ int dbus_init(void)
 
         owner_id = g_bus_own_name(G_BUS_TYPE_SESSION,
                                   FDN_NAME,
-                                  G_BUS_NAME_OWNER_FLAGS_NONE,
+                                  G_BUS_NAME_OWNER_FLAGS_REPLACE,
                                   dbus_cb_bus_acquired,
                                   dbus_cb_name_acquired,
                                   dbus_cb_name_lost,


petrblaho:
This solved mentioned issue for me and now I use dunst running as systemd user service as my notification daemon with GNOME.

bebehei:
I'm not sure if a hardcoded replace flag is a good thing.

petrblaho:
> I'm not sure if a hardcoded replace flag is a good thing.

Yep, I am not sure about that too... but I am not versed in either DBus or daemons development.

Maybe some flag or ENV var to read? No idea how to do that though as I do not use C at all :-)

bebehei:
I would actually favor `dunst -replace` in this case. As a short reminder: Backwards compatibility :wink:

I'm suspicious about this change. I'm pretty sure we hit a corner case and will shoot our foot with it.

tsipinakis:
In this case, we'd also need to have a similarly named dunstrc option.
`dunst -replace` sounds good in theory but then we'll also have `replace = {true | false }` in the dunstrc that can be confused with stacking or similar features. I'd propose `dbus_replace`.

tsipinakis:
Closing this as it's stale. Please open an issue or a new PR if this is still needed.

Issue #508, opened 2018-04-02
Some notifications are not saved in history


Sweenu:
For some reason, my notifications coming from kdeconnect don't get saved in history even though I have no special rule set.

Any idea where this behavior could come from and how I could solve that?

### Installation info

- Version: `1.3.0-5-gf7cf5b6`
- Install type: `pacman`
- Distro and version: `4.15.14-1-ARCH`

tsipinakis:
That's probably from #343, there's an internal rule that notifications marked as `transient` by the client are not saved in history. This is a bit confusing as I've mentioned [before](https://github.com/dunst-project/dunst/issues/310#issuecomment-358126293). I've now removed that rule from master.

Sorry for the confusion.

For anyone running the latest release you can add the following in your dunstrc to get around this:

```
[keep-transient]
match_transient = true
history_ignore = false
```

Sweenu:
Thank you! That did it for me.

PR #606, opened 2019-02-23
Decouple transient hint from the behaviour

tsipinakis:
This occurred to me while writing the changelog for 1.4.

As we learned in #508 adding implicit behaviour to a not-so-well-defined hint can cause confusion. Instead, lets decouple it from any action entirely and let the user decide what they want to do with it.


tsipinakis:
@bebehei Any clues what's up with the ubuntu image? It seems to [fail intermittently](https://circleci.com/gh/dunst-project/dunst/525)

tsipinakis:
So yesterday when I was writing this I thought `*_transient` hadn't been in a release yet so I just changed it directly. Apparently it was in 1.3.0 so backwards compatibility is also missing here.

bebehei:
> @bebehei Any clues what's up with the ubuntu image? It seems to [fail intermittently](https://circleci.com/gh/dunst-project/dunst/525)

It's a super rare race condition in the dbus tests. It's about the threading necessary to create a main loop. Idk why this happens. I suspect it inside GLib.

bebehei:
@tsipinakis Sorry, I forgot to answer about the real code here.

What's your point on this PR? Would you like to move `idle_threshold` from settings to the notification?

tsipinakis:
> What's your point on this PR? Would you like to move `idle_threshold` from settings to the notification?

No, my point is that the `transient` hint is not well defined and many programs set it without much consideration.
On the users perspective now some notifications behave one way and others differently without any clear indication as to why. It's the problem we ran in with #508 ( had linked the wrong issue in the original PR, oops!).

The main point of this PR is to make the transient hint do nothing by default. An addition is that it also exposes a way to bypass `idle_threshold` for some notifications so the same behaviour can be brought back as rules if needed.

bebehei:
Yeah, but #508 is about the counterintuitive behavior, I introduced. You fixed it in f8538c9.

> Instead, lets decouple it from any action entirely and let the user decide what they want to do with it.

I don't get what the point of this change should be. It's all just about the default? So that if an app sends the `transient` hint, it should get ignored by default?

That's easily done with just a simple rule, if you want to ignore the hint:

```
[ignore_transient_hints]
    set_transient = false
```

Introducing a new field `idle_threshold`, which is already available in the `[global]` section, might be counterintuitive, too. Especially if it's using complete different semantics.

----

But I definitely see at least one problem: I haven't documented it and I should do that at least.

tsipinakis:
You're right, lets leave it as it is then aim to improve documentation.

PR #608, opened 2019-02-25
XRandR DPI Online Changes

### LINKED ISSUE #382: Re-read DPI before displaying new notification
Traveling laptop use case.
DPI can change on the fly, but running dunst will continue using outdated value that was in effect back when it was started.

bebehei:
Changes DPI values online when receiving a corresponding XEvent.

To verify:

```shell
xrandr --dpi "${DPI}"
```

See the `Query the X11 screen's DPI instead of monitor` commit's description for further info.

Fixes #382

### DIFF
--- a/src/x11/screen.c
+++ b/src/x11/screen.c
@@ -37,31 +37,42 @@ static int x_follow_tear_down_error_handler(void);
 static int FollowXErrorHandler(Display *display, XErrorEvent *e);
 static Window get_focused_window(void);
 
+
+/**
+ * A cache variable to cache the Xft.dpi xrdb values.
+ * We do not expect to change xrdb often, but there's much
+ * overhead to query it once.
+ *
+ * @retval -DBL_MAX: uncached
+ * @retval     <=0: Invalid and unusable value
+ * @retval      >0: valid
+ */
+double screen_dpi_xft_cache = -DBL_MAX;
+
+void screen_dpi_xft_cache_purge(void)
+{
+        screen_dpi_xft_cache = -DBL_MAX;
+}
+
 static double screen_dpi_get_from_xft(void)
 {
-        static double dpi = -1;
-        //Only run this once, we don't expect dpi changes during runtime
-        if (dpi <= -1) {
+        if (screen_dpi_xft_cache == -DBL_MAX) {
+                screen_dpi_xft_cache = 0;
+
                 XrmInitialize();
                 char *xRMS = XResourceManagerString(xctx.dpy);
 
-                if (!xRMS) {
-                        dpi = 0;
-                        return 0;
-                }
+                ASSERT_OR_RET(xRMS, screen_dpi_xft_cache);
 
                 XrmDatabase xDB = XrmGetStringDatabase(xRMS);
                 char *xrmType;
                 XrmValue xrmValue;
 
-                if (XrmGetResource(xDB, "Xft.dpi", "Xft.dpi", &xrmType, &xrmValue)) {
-                        dpi = strtod(xrmValue.addr, NULL);
-                } else {
-                        dpi = 0;
-                }
+                if (XrmGetResource(xDB, "Xft.dpi", "Xft.dpi", &xrmType, &xrmValue))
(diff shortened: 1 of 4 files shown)


src/x11/screen.c
```diff
@@ -37,31 +37,37 @@ static int x_follow_tear_down_error_handler(void);
...
+/**
+ * A cache variable to cache the Xft.dpi xrdb values.
+ * We do not expect to change xrdb often, but there's much
+ * overhead to query it once.
+ *
+ * @retval -DBL_MAX: uncached
+ * @retval      <=0: Invalid and unusable value
+ * @retval       >0: valid
+ */
+double screen_dpi_xft_cache = -DBL_MAX;
```
tsipinakis: `DBL_MAX` should probably be `DBL_MIN`

src/x11/screen.c
```diff
@@ -64,6 +64,27 @@ static double get_xft_dpi_value(void)
...
+                return screen_dpi_get_from_monitor(scr);
+
+        if (screen_dpi_get_from_xft() > 0)
+                return screen_dpi_get_from_xft();
+
+        /* Calculate the DPI on the overall screen size.
+         * xrandr --dpi <DPI> does only change the overall screen's millimeters,
+         * but not the physical screen's sizes. */
+        return ((((double)DisplayWidth  (xctx.dpy, 0)) * 25.4) /
+                 ((double)DisplayWidthMM(xctx.dpy, 0)));
```
tsipinakis: What happens with multi-monitor setups here? Perhaps `0` should be replaced with `scr->id`?

bebehei: It's explicitly a `0`. Because we're accessing the (virtual?) x11 screen and not a screen linked to a physical monitor. Run

```
> xrandr -q | head -n1
Screen 0: minimum 8 x 8, current 3840 x 1080, maximum 16384 x 16384
```

And this `0` given there in the code "hardcodes" it to that `Screen 0` mentioned by xrandr. While `scr->id` would reference the `nth` active physical monitor.

It "hardcodes" it to the `Screen 0` in this code. Yes, this code doesn't scale for "Screen 1`, this would be a problem. I don't know how to create multiple xrandr screens either. It covers 99.9% of all users DPI problems when you choose to change the DPI with xrandr.

The definitions of the terms of monitor, display, screen and output are a complete mess. Our "screen" is something completely different to the term screen used in xrandr.

src/x11/screen.c
bebehei: IEEE754 numbers are symmetrical. This works perfectly fine.

I've probably overseen the `DBL_MIN` definition.

src/x11/screen.c
bebehei: But yeah, It warrants a comment. I've added it.

bebehei:
Rebased and all points of your review covered.

tsipinakis:
Running this under a debugger looks like even after re-querying the Xresources db it still returns the old value and isn't updated when doing `xrdb -merge`. Does it work for you?

This is what I did to reproduce:
```bash
./dunst &
notify-send Test
echo "Xft.dpi: 120" | xrdb -merge
xrandr --dpi 120
notify-send test
```

I verified that `xrdb -query` shows the new dpi value and that Xresources is re-queried via gdb.

src/x11/screen.c
bebehei: `0` might not be portable, but `XDefaultScreen` will always report the correct value. So I changed it with this.

src/x11/screen.c
```diff
@@ -37,31 +37,42 @@ static int x_follow_tear_down_error_handler(void);
...
+/**
+ * A cache variable to cache the Xft.dpi xrdb values.
+ * We do not expect to change xrdb often, but there's much
+ * overhead to query it once.
+ *
+ * @retval DBL_MIN: uncached
+ * @retval     <=0: Invalid and unusable value
+ * @retval      >0: valid
+ */
+double screen_dpi_xft_cache = DBL_MIN;
```
tsipinakis: Sorry for the misleading review earlier `DBL_MIN` is the smallest _positive_ value, should've double checked that one :/

PR #610, opened 2019-03-01
Use @retval in codedocs

bebehei:
I've found the `@retval` directive in the doxygen doc. Nice thing

### DIFF
--- a/src/queues.h
+++ b/src/queues.h
@@ -29,7 +29,8 @@ const GList *queues_get_displayed(void);
 /**
  * Get the highest notification in line
  *
- * @return a notification or NULL, if waiting is empty
+ * @returns the first notification in waiting
+ * @retval NULL: waiting is empty
  */
 const struct notification *queues_get_head_waiting(void);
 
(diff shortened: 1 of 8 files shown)


src/queues.h
```diff
@@ -73,8 +74,8 @@ int queues_notification_insert(struct notification *n);
  *
  * @param new replacement for the old notification
  *
- * @return true, if a matching notification has been found and is replaced
- * @return false, else
+ * @return true: a matching notification has been found and is replaced
+ * @return false: else
```
tsipinakis:
```suggestion
 * @return false: otherwise
```
It was there before this PR, but lets keep things consistent.

---

Here you're using two `@return`'s for a boolean return value but in `window_is_fullscreen`, `dbus_get_fdn_daemon_info` etc you're using two `@retvals`, I assume `@retval` would be the correct one for this as well.

bebehei: > Here you're using two @return's for a boolean [...]

Yeah, that was a mistake. Thanks for the info.

bebehei: > Edit: I'm not sure who has the @ RetVal handle but you just pinged them with this commit message! :)

I'm not sure about that. It got in my mind, too. But it was right after the push. But I thought it might be fun :grin:

PR #647, opened 2019-08-09
Don't overflow when passing INT_MAX on DBus wire

### LINKED ISSUE #646: The timeout value in dunst wraps around
dunst internally stores timeouts in microseconds, in a signed 32 bit integer, which wraps around at 2147484, which corresponds to ~40 minutes.  Having timeouts longer than that is not unreasonable, so the wrapping should be fixed.

notify-send -t 2147483 foo
will stay on the screen for 35 minutes, while
notify-send -t 2147484 foo
will use the default (for me, 10s)

### Installation info

- Version: Dunst - A customizable and lightweight notification-daemon 1.3.2
- Install type: package
- Distro and version: Fedora 30

bebehei:
Fixes #646

### DIFF
--- a/test/dbus.c
+++ b/test/dbus.c
@@ -660,6 +660,34 @@ TEST test_hint_raw_image(void)
         PASS();
 }
 
+/* We didn't process the timeout parameter via DBus correctly
+ * and it got limited to an int instead of a long int
+ * See: Issue #646 (The timeout value in dunst wraps around) */
+TEST test_timeout_overflow(void)
+{
+        struct notification *n;
+        struct dbus_notification *n_dbus;
+
+        n_dbus = dbus_notification_new();
+        n_dbus->app_name = "dunstteststack";
+        n_dbus->app_icon = "NONE";
+        n_dbus->summary = "test_hint_urgency";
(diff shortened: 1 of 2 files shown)


test/dbus.c
```diff
@@ -660,6 +660,34 @@ TEST test_hint_raw_image(void)
...
+        struct notification *n;
+        struct dbus_notification *n_dbus;
+
+        n_dbus = dbus_notification_new();
+        n_dbus->app_name = "dunstteststack";
+        n_dbus->app_icon = "NONE";
+        n_dbus->summary = "test_hint_urgency";
+        n_dbus->body = "Summary of it";
+        n_dbus->expire_timeout = 2147484;
+        long int expected_timeout = 2147484000ll;
```
tsipinakis: Nitpick since it doesn't really affect anything in this case but: You're assigning a `long long int` to a `long int`.

bebehei: So, I've used the GLib way.

Idk why GLib is only using `long` as their 64 bit specifier. According to the german wikipedia-article this is not true in all cases, but it seems to work and the GLib guys have a reason to use only one `l`.

PR #656, opened 2019-08-22
Fix dead link to pango reference in dunstrc

rootkiwi:


tsipinakis:
Just as I clicked merge I realised this is incomplete: There's another reference to that url in the manpage. Also fixed with 9e29406bb53e115d5f024e9052bdb916825a7991.

PR #659, opened 2019-08-31
Enable extended patterns in fnmatch

### LINKED ISSUE #645: There is a way to enable extglob in dunstrc file
Hi, this is not a bug, just a question to know if is there a way to enable extglob in dunstrc file, I really would to know  how enable, I put an spefic sound for all notifications, but some apps have your own notification sound, and  when is a notification is fired for these apps play two songs at same time.

I trying to make of following way:

```
[playsound]
    icon = "dialog-information"
    desktop_entry = "!(org.gnome.clocks)"
    script = /home/rodrigo/.local/bin/notification-sound # custom script path
```
and do not works.

or who knows made a blacklist for spefic apps, I know this could be hard, so would be nice if extglob works :)

### LINKED ISSUE #658: applying `new_icon` to one rules changes for all
I was looking to disable icon for all application, except one. I was looking to set `icon_position` via rule but its not available (relevent issue: https://github.com/dunst-project/dunst/issues/638).  I tried to get workaround the issue by setting `icon_position = left` in `global` and use `new_icon` in rules. Then I caught this issue.

My `dunstrc` is something like this. 

```cfg
[global]

icon_position = left

[my-app]
	appname = myapp
	summary = "*"
	timeout = 0

[fix-icon]
	summary = "*"
	new_icon = ""
```

## Expected Behaviour 

Only notifications from `myapp` should display the icon and other icons should not show icon at all.

## Actual Behaviour

No icon in all notifications. If `new_icon = ""` rule is removed, then icon in all notifications is shown.

___

It looks like setting `new_icon = ""` in one rule does not respect icon supplied to a notification in all notifications and sets them all to `null`.

tsipinakis:
I wanted to add a compilation warning but apparently `#warning` is a gcc specific extension. Not sure how else we could indicate if this is enabled or not.

Fixes #645
Fixes #658

### DIFF
--- a/docs/dunst.pod
+++ b/docs/dunst.pod
@@ -678,7 +678,9 @@ Matches only messages that were send via notify-send. If multiple filter
 expressions are present, all of them have to match for the rule to be applied
 (logical AND).
 
-Shell-like globing is supported.
+Shell-like globing is supported. For more information see fnmatch(3). Support
+for extended patterns is enabled (EXTMATCH) but I<they are only available on GNU
+libc>
 
 =item B<modifying>
 
--- a/src/rules.c
+++ b/src/rules.c
@@ -1,11 +1,20 @@
 /* copyright 2013 Sascha Kruse and contributors (see LICENSE for licensing information) */
 
+// Needed for FNM_EXTMATCH flag
+#define _GNU_SOURCE
+
 #include "rules.h"
 
 #include <fnmatch.h>
 #include <glib.h>
 
 #include "dunst.h"
+
+#ifdef FNM_EXTMATCH
+#define FNMATCH_FLAGS FNM_EXTMATCH
+#else
+#define FNMATCH_FLAGS 0
+#endif
 
 GSList *rules = NULL;
 
@@ -84,7 +93,7 @@ struct rule *rule_new(void)
 
 static inline bool rule_field_matches_string(const char *value, const char *pattern)
 {
-        return !pattern || (value && !fnmatch(pattern, value, 0));
+        return !pattern || (value && !fnmatch(pattern, value, FNMATCH_FLAGS));
 }
 
 /*


tsipinakis:
> I've got absolutely no knowledge, how GLib and other libs are working when compiling multiple units with different sources and linking them.

Theoretically this should not play a role at all. Either the GNU extensions exist in libc or they don't our definitions do not affect that. All it does is control if they are exported in the header to allow us to use them.

See [`fnmatch.h`](https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/fnmatch.h;h=fa4d56aeb7b39af7370ca15ff3f79208c331c1fc;hb=HEAD), [`fnmatch.c`](https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/fnmatch.c;h=a58e1743ce14f1276587c427d9f5a9dbdb8a05d2;hb=HEAD).

tsipinakis:
> But I've been getting the suspicious if this is a good solution to solve our problem

I've been getting suspicious too but for an entirely different reason - I'm unsure if `fnmatch` is the best fit here. As the manpage mentions it's intended for paths and while it does what we need I feel like regex would be a better fit.

bebehei:
To achieve negative matching, I guess we either have to implement such "quirk" and document it right or allow regexes.

TBH, I'd prefer the fnmatch thing ... if it wouldn't be such a weird mess.

---

Just to evaluate a possible regex path: How would we implement it? While thinking about it, I guess,. you had some ideas. _We can't have it as a drop-in replacement._ Chars like `.?\()[]` would have a complete different meaning now.

tsipinakis:
> We can't have it as a drop-in replacement. Chars like .?\()[] would have a complete different meaning now.

I was thinking using something like a tilde as a prefix for it, so `appname = "firefox"` uses fnmatch and `appname = "~firefox" uses regex.

bebehei:
> I was thinking using something like a tilde as a prefix for it, so `appname = "firefox"` uses fnmatch and `appname = "~firefox" uses regex.

What about a slash? I guess it's more intuitive.

tsipinakis:
I'll close this as, as said above it's better to add an option to use regex rather than enable extended patterns.

PR #666, opened 2019-10-06
fix icon positioning and add image_padding variable

joesitton:


tsipinakis:
Ah, changes to the render code. That's a dangerous area to play in as it's completely untested.

I don't have time to review right now and next week gets busy, I'll try to give this a thorough review in the next few days. Apologies for any delays. I vaguely remember an issue requesting something similar to this even though I can't find it right now with a quick search.

---

What's up with the commit history, it appears you added {x,y}offset support to dunstify but removed them after?

joesitton:
I think I had x/y offsets and was attempting to do that but couldn't be bothered after trying and just ended up with this

tsipinakis:
I found some issues when testing this,

Using `image_padding` seems to add some sort of padding to the right of the notification as well when the icon is on the left.
`image_padding = 20`:
2019-10-11T19-45-27
padding disabled:
2019-10-11T19-45-40

Additionally when `icon_position = right` the padding seems to be broken on the right side. You probably need to update `calculate_dimensions` to include that and update the window side accordingly.

2019-10-11T19-46-04

Issue #392, opened 2017-09-27
 Set minimum image/icon size 


xkr47:
On HiDPI (280dpi) displays icons embedded in notifications are sometimes too small to see, so an option to set the minimum image/icon size (e.g. upscale image if too small) would be appreciated.

tsipinakis:
There is an option to *downscale* an icon but upscaling would have some serious quality issues. Is there a reason why you're not using a higher resolution icon pack?

xkr47:
Yes I am entirely aware of this :)
But I'm talking about icons embedded in notifications originating from web sites which I am not in control of. The images look fine on 100dpi displays. Upscaling won't make them look much worse than that I suppose. Now it's hard to recognize the images (people's faces mostly) when they are only ~2mm (0.1in) big on the display, so just seeing them bigger even if they would be a bit blurry would actually help.
The default could be set to 0 so upscaling would occur unless the user really wants it.
I could even give a shot at implementing this if you don't mind?

xkr47:
It ended up being quite simple; see PR #393. Works for me :)

bebehei:
I'm not a user of a HiDPI screen. Can somebody explain me this situation?

I see and understand, why the image gets too small, when using higher DPIs and your software is not DPI-aware. But if I understand correctly, dunst should be DPI-aware (#240). So, why does the icon get displayed too small? Shouldn't be the best fix to scale the icon DPI-aware, too?

xkr47:
hmm.. I'm actually using an older snapshot from this summer because the xrandr library in devuan is a bit old. I failed to realize there was dpi stuff taking care of this in latest version so I took my own approach instead. So maybe we can close this for now and reopen later if there is still need. Sorry for the disturbance.

bebehei:
@xkr47 Just because I say dunst **should** be DPI-aware, doesn't make dunst automatically fully DPI-aware.

There may be some bugs. Are you sure, that dunst scales now the icon correct?

nikarh:
It would still be nice to have min_icon_size for icon themes consisting solely of SVG images (such as https://github.com/daniruiz/flat-remix).

ShawnMcCool:
This would be a fantastic feature for SVG as so many laptops have higher resolutions now. Dunst renders svg icons very small and I can't find a way to make them larger.

xkr47:
[...]

PR #674, opened 2019-11-29
Implement #392, take 2

xkr47:
New attempt at #392, with [great ideas from @tsipinakis](https://github.com/dunst-project/dunst/issues/392#issuecomment-558519183) taken into account. Have still not attempted to test/take into account any DPI-related functionality, so cannot say yet how/if it will affect the outcome. 

Do you want more tests for the scaling functionality? Current tests pass.

Implementation details: I moved the scaling responsibility to the lowest functions in the call hierarchy involved in loading the images - both when loading from disk and from the notification message so responsibilities are clearer code-wise. Made some functions static that are no longer needed elsewhere.

### DIFF
--- a/src/icon.c
+++ b/src/icon.c
@@ -177,7 +177,6 @@ GdkPixbuf *get_pixbuf_from_file(const char *filename)
         if (!gdk_pixbuf_get_file_info (path, &w, &h)) {
                 LOG_W("Failed to load image info for %s", filename);
                 g_free(path);
-                g_error_free(error);
                 return NULL;
         }
         icon_size_clamp(&w, &h);
--- a/src/icon.c
+++ b/src/icon.c
@@ -175,6 +175,7 @@ GdkPixbuf *get_pixbuf_from_file(const char *filename)
 
         if (!gdk_pixbuf_get_file_info (path, &w, &h)) {
                 LOG_W("Failed to load image info for %s", filename);
+                g_free(path);
                 g_error_free(error);
                 return NULL;
         }
--- a/src/icon.c
+++ b/src/icon.c
@@ -115,6 +115,7 @@ cairo_surface_t *gdk_pixbuf_to_cairo_surface(GdkPixbuf *pixbuf)
  *
  * @param w a pointer to the image width, to be modified in-place
  * @param h a pointer to the image height, to be modified in-place
+ * @return TRUE if the dimensions were updated, FALSE if they were left unchanged
  */
 static bool icon_size_clamp(int *w, int *h) {
         int _w = *w, _h = *h;
@@ -155,11 +156,11 @@ static GdkPixbuf *icon_pixbuf_scale(GdkPixbuf *pixbuf)
         int h = gdk_pixbuf_get_height(pixbuf);
 
         if (icon_size_clamp(&w, &h)) {
-                GdkPixbuf *scaled;
-                scaled = gdk_pixbuf_scale_simple(pixbuf,
-                                                 w,
-                                                 h,
-                                                 GDK_INTERP_BILINEAR);
+                GdkPixbuf *scaled = gdk_pixbuf_scale_simple(
+                                pixbuf,
+                                w,
+                                h,
+                                GDK_INTERP_BILINEAR);
                 g_object_unref(pixbuf);
                 pixbuf = scaled;
         }
--- a/src/icon.c
+++ b/src/icon.c
@@ -110,16 +110,18 @@ cairo_surface_t *gdk_pixbuf_to_cairo_surface(GdkPixbuf *pixbuf)
         return icon_surface;
 }
 
-GdkPixbuf *icon_pixbuf_scale(GdkPixbuf *pixbuf)
-{
-        ASSERT_OR_RET(pixbuf, NULL);
-
-        int w = gdk_pixbuf_get_width(pixbuf);
-        int h = gdk_pixbuf_get_height(pixbuf);
-        int landscape = w > h;
-        int orig_larger = landscape ? w : h;
+/**
+ * Scales the given image dimensions if necessary according to the settings.
+ *
+ * @param w a pointer to the image width, to be modified in-place
+ * @param h a pointer to the image height, to be modified in-place
+ */
+static bool icon_size_clamp(int *w, int *h) {
+        int _w = *w, _h = *h;
+        int landscape = _w > _h;
+        int orig_larger = landscape ? _w : _h;
         double larger = orig_larger;
-        double smaller = landscape ? h : w;
+        double smaller = landscape ? _h : _w;
         if (settings.min_icon_size && smaller < settings.min_icon_size) {
                 larger = larger / smaller * settings.min_icon_size;
                 smaller = settings.min_icon_size;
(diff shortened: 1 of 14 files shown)


xkr47:
Will check those CI errors and try to fix them..

xkr47:
Apparently the function `gdk_pixbuf_get_file_info()` which I used to be able to get the default image size has a memory leak that is triggered when loading e.g. `test/data/icons /path/invalid/icon1.svg`. I currently have 2.36.11-1ubuntu0.1 and at least the version in the CI build also has the same issue.

@tsipinakis I could check with the latest release and with the latest development branch version of gdk-pixbuf to see if it already fixed or try to get it fixed if not. But nevertheless I guess you'd want me not to use this function as not to cause memory leaks when loading broken images? As an alternative, I could just load the image using the normal `gdk_pixbuf_new_from_file()`, check its size and in case a different size is wanted, reload it using the `gdk_pixbuf_new_from_file_at_scale()`. What do you think?

tsipinakis:
Weirdly enough I _can't_ reproduce this with Debian stable or sid, which leads me to believe that this might've already been fixed and our CI images are simply out of date. It's also not clear if the leak is in gdk-pixbuf or librsvg, my bet is on the latter.

> But nevertheless I guess you'd want me not to use this function as not to cause memory leaks when loading broken images?

Eh, that's a bug that'll even if it's not fixed yet it'll be fixed by the time this is released and reaches the distros. There is no need to hold off to using that function IMO.

---

On another note, some small TODOs for adding a new setting:
- [ ] Add `min_icon_size` to the `dunstrc` with a small explanation similar to `max_icon_size`
- [ ] Similarly, document its function in `docs/dunst.pod`.

src/icon.c
```diff
@@ -141,8 +171,19 @@ GdkPixbuf *get_pixbuf_from_file(const char *filename)
 {
         char *path = string_to_path(g_strdup(filename));
         GError *error = NULL;
+        gint w, h;
 
-        GdkPixbuf *pixbuf = gdk_pixbuf_new_from_file(path, &error);
+        if (!gdk_pixbuf_get_file_info (path, &w, &h)) {
+                LOG_W("Failed to load image info for %s", filename);
+                g_error_free(error);
+                return NULL;
```
tsipinakis: The last leak from valgrind is actually ours, because of the early return here `path` isn't freed.

tsipinakis:
> Weirdly enough I can't reproduce this with Debian stable or sid, which leads me to believe that this might've already been fixed and our CI images are simply out of date. It's also not clear if the leak is in gdk-pixbuf or librsvg, my bet is on the latter.

Oh! Duh! This is probably [rsvg/147](https://gitlab.gnome.org/GNOME/librsvg/merge_requests/147), our suppression file just didn't catch it because the stack trace changed with this PR. In this case you can just update `.valgrind.supressions` and ignore it.

xkr47:
Thanks for your feedback & review!

xkr47:
Do you have any preferred default value for min_icon_size? Should I set it to 0 to avoid changing current operation (except for downscaling of svgs which will nevertheless

xkr47:
Are these out of sync?
* `max_icon_size = 0` in `config.h`
* `max_icon_size` defaults to `0` in `dunst.pod`
* `max_icon_size = 32` in `dunstrc`

xkr47:
I want to add that this solution is not taking DPI into account as @bebehei [requested](https://github.com/dunst-project/dunst/issues/392#issuecomment-333327441). If we want it to scale icons based on DPI as well, I guess we would need to determine a base DPI like 96 or something, and then scale images by `current_dpi / base_dpi`. Are images reloaded when notification moves between montiors? I.e. do we have an opportunity to reload & rescale them to the correct (dpi-adjusted) size every time?

xkr47:
If we go down the path of dpi-awareness I guess at least the following settings should be dpi-aware too:
* geometry
* line_height
* notification_height
* separator_height
* padding
* horizontal_padding

tsipinakis:
> Do you have any preferred default value for min_icon_size? Should I set it to 0 to avoid changing current operation (except for downscaling of svgs which will unconditionally be improved slightly smile)?

Leaving it at 0 is fine, I don't think there's any real reason to have it enabled by default.

> Also what is the purpose of the "default" values set in config.h?

The default values in `config.h` is what will be used if a setting isn't present in the dunstrc. These weren't in sync when I took on maintaining dunst and since then I've never really looked more into fixing that because I don't want to break existing setups without a proper reason. (I've seen many people define only the settings they want in `dunstrc` and leave everything else use `config.h` values).

> Are images reloaded when notification moves between montiors? I.e. do we have an opportunity to reload & rescale them to the correct (dpi-adjusted) size every time?

Yes, images are reloaded on every redraw (perhaps wrongly) - So it can be done and there's also opportunity for optimization to avoid rescaling every time.

> I want to add that this solution is not taking DPI into account as @bebehei requested

IMO this is out of the scope of this PR, let's not delay merging this.
But if you're up to implementing proper DPI support don't let me discourage you by all means I'd love to see such a PR :) (Especially if you have the screen/setup to properly test it!)

---

Either way I'll do the final review either tonight or tomorrow, thanks for implementing this!

xkr47:
Yeah the dpi scaling is apparently experimental and not even documented in the pod.. I almost started writing a script already that ensures everything is documented but got distracted..

src/icon.c
xkr47: Should be fixed now

src/icon.c
```diff
@@ -141,8 +172,20 @@ GdkPixbuf *get_pixbuf_from_file(const char *filename)
 {
         char *path = string_to_path(g_strdup(filename));
         GError *error = NULL;
+        gint w, h;
 
-        GdkPixbuf *pixbuf = gdk_pixbuf_new_from_file(path, &error);
+        if (!gdk_pixbuf_get_file_info (path, &w, &h)) {
+                LOG_W("Failed to load image info for %s", filename);
+                g_free(path);
+                g_error_free(error);
```
tsipinakis: Whoop, caught a bug! `error == NULL` always here, there is no need to free. (thankfully glib catches it and doesn't crash)

xkr47:
Do you want more tests for the scaling functionality? Current tests pass.

src/icon.c
xkr47: :sweat_smile:

tsipinakis:
> Do you want more tests for the scaling functionality?

More tests are always better. I didn't mention anything about tests earlier because it didn't feel right asking for them when there aren't any for `max_icon_size` either.
If you have the time to add them please go ahead!

xkr47:
I will try next week. If you are in a hurry I can create a separate PR later.

tsipinakis:
@xkr47 Ping? No pressure, just wanted to make sure you haven't forgotten about this.

xkr47:
Thanks! Yeah bit busy but not forgotten :)

xkr47:
There we go, got it done thanks to some well-timed motivation speech :) I split most of `dbus_notification_set_raw_image()` test helper to a separate function in a separate file (see `test/helpers.[ch]`) so I could do better unit testing of images embedded in notifications.

Let me know of any improvements/corrections/etc you'd wish to see.

(The travis builds did not get very far and the failures feel totally unrelated to my code.)

tsipinakis:
LGTM, merged! Kudos for taking care of the tests as well, it's the part most people dread.

xkr47:
Excellent news! It was a pleasure really, the codebase was easy to follow and the existing test helpers did most of the dreaded hard work already :)

Issue #486, opened 2018-01-18
Add option to align icon/text vertically


chronus7:
In it's current state (v1.3.0), dunst seems to vertically center the icon and text. I personally find this irritating and would prefer to align them to the top of the notification.

The following graphics demonstrate/clarify the idea with a) an icon smaller than the text and b) a much larger icon than the text.

```
Center alignment (current behaviour)
┌───────────────────┐  |  ┌───────────────────┐
│      first line   │  |  │ ####              │
│  #   second       │  |  │ ####  one line    │
│      third line   │  |  │ ####              │
└───────────────────┘  |  └───────────────────┘

Top alignment
┌───────────────────┐  |  ┌───────────────────┐
│  #   first line   │  |  │ ####  one line    │
│      second       │  |  │ ####              │
│      third line   │  |  │ ####              │
└───────────────────┘  |  └───────────────────┘

Bottom alignment
┌───────────────────┐  |  ┌───────────────────┐
│      first line   │  |  │ ####              │
│      second       │  |  │ ####              │
│  #   third line   │  |  │ ####  one line    │
└───────────────────┘  |  └───────────────────┘
```

Would it be possible to implement such alignment? I would much appreciate it.

FeintGS:
Agreed, I can totally understand why people would be driven nuts by a limitation in the current limitation like this. For future readers who are looking for a solution to this problem, here's a workaround hack you can use. Basically, what you do is trick dunst into preserving the leading/trailing whitespace in your string by using a "pseudo" space character. This way, you can vertically top-align or bottom-align your notification text.



**Before the hack:**
```
notify-send "🎵 ~Now playing~ 🎵" "$SONG_TITLE\n$ALBUM_NAME"
```
![before](https://user-images.githubusercontent.com/40724862/56878532-9273b680-6a22-11e9-9e16-11df0a5c0c25.png)

**After the hack:**
```
# NOTE!!! The "space" at the end is actually the Unicode symbol '\u0020'
notify-send "🎵 ~Now playing~ 🎵" "$SONG_TITLE\n$ALBUM_NAME\n\n\n "
```
![after](https://user-images.githubusercontent.com/40724862/56878537-97386a80-6a22-11e9-8830-a9a82dfbb8f3.png)

Good luck! [Here's a video demo of it in action](https://twitter.com/FeintGS/status/1122726445552300033)

kushraj:
Hi, is there any possibility this will be taken up any time soon ?
[...]

Issue #336, opened 2017-07-15
Redesigning the window geometry/dimension settings


tsipinakis:
As mentioned in a [comment](https://github.com/dunst-project/dunst/pull/335#issuecomment-315520790) in #335, a lot of the current display settings are built on top of each-other in a confusing way e.g. the height part of the geometry is actually number of notifications to be displayed, `notification_height` is the pixel height, or the confusion between width, dynamic width and `shrink` which takes some close reading of the documentation to understand the difference.

We should
- [ ] Compile a list of use cases we should account for (feel free to post yours if it's not that well supported, or not supported at all)
- [ ] Create a list of changes so that the end result is an intuitive configuration experience while not sacrificing any of the currently available functionality

---

As it stands, geometry is probably the most complicated part of the config since there are a lot of corner cases one might not predict. For example, if the width is omitted the message expands to cover the whole screen but if it's 0 it's considered dynamic and expands to the longest message.

After looking at the current implementation I came up with the following proposal:

Replace `geometry` `shrink`, `notification_height` with the following
* `minimum_width`: The minimum pixel width for the notification as described in #335 
* `minimum_height`: The equivalent of `notification_height`
If any of the minimum_* settings is zero, it is considered dynamic and will be shortened as needed.
* `maximum_width`: The maximum width in pixels for the notification, when that limit is reached it will either be wrapped, or ellipsized according to `word_wrap`
* `maximum_height`: The maximum height in pixels for the notification, after that limit is reached any further text is cut and ellipsized.

* `origin`: A setting that can be set to `topleft`, `bottomleft`, `middle`, `topright` and `bottomright` corresponding to the location that the notification will be drawn at. `middle` would also account for the notification window length to account for proper centering as described in #233 
* `offset_horizontal`: The offset in pixels from the origin in the horizontal axis, can be positive (to the left) or negative (to the right
* `offset_vertical`: As with `offset_horizontal` but in the vertical axis.
* `notification_limit`: The maximum number of notification that will be displayed at any one time (equivalent to geometry's height setting)

Another important point to consider to account for backwards compatibility with the previous settings if possible to provide the best experience for downstream users.

This is a very rough draft at this point and unclear whether any of this is going to move forward. I would love to hear some opinions on this, and more suggestions if you think there is a better way or I missed something.

hobarrera:
I'd rather have a single `width` which takes up to two values:
```
width = 50, 100 #  min, max, respectively
[...]

PR #684, opened 2020-01-25
Adding vertical content alignment control.

chronus7:
This adds the option `content_alignment`, which allows the user to set the vertical alignment of the notification's content (i.e. icon and
text) to either:

- top (`CONTENT_TOP`),
- center (`CONTENT_CENTER`, default), or
- bottom (`CONTENT_BOT`).

The default preserves current behaviour, while the other options fulfill #486.

I am open to any suggestions or comments regarding naming and/or code-style.

### DIFF
--- a/dunstrc
+++ b/dunstrc
@@ -134,7 +134,7 @@
 
     # Vertical alignment of message text and icon.
     # Possible values are "top", "center" and "bottom".
-    content_alignment = center
+    vertical_alignment = center
 
     # Show age of message if message is older than show_age_threshold
     # seconds.
--- a/dunstrc
+++ b/dunstrc
@@ -132,6 +132,10 @@
     # Possible values are "left", "center" and "right".
     alignment = left
 
+    # Vertical alignment of message text and icon.
+    # Possible values are "top", "center" and "bottom".
+    content_alignment = center
+
     # Show age of message if message is older than show_age_threshold
     # seconds.
     # Set to -1 to disable.
--- a/src/draw.c
+++ b/src/draw.c
@@ -558,7 +558,7 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width)
                                 break;
                         case CONTENT_BOTTOM:
                                 image_y = h + settings.padding - image_height;
-                                if (image_y < 0 || image_y > h) image_y = settings.padding;
+                                if (image_y < settings.padding || image_y > h) image_y = settings.padding;
                                 break;
                         default:    // CONTENT_CENTER
                                 break;
--- a/src/draw.c
+++ b/src/draw.c
@@ -511,32 +511,65 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width)
         int h_text;
         pango_layout_get_pixel_size(cl->l, NULL, &h_text);
 
-        if (cl->icon && settings.icon_position == ICON_LEFT) {
-                cairo_move_to(c, cairo_image_surface_get_width(cl->icon) + 2 * settings.h_padding,
-                                 settings.padding + h/2 - h_text/2);
-        } else if (cl->icon && settings.icon_position == ICON_RIGHT) {
-                cairo_move_to(c, settings.h_padding, settings.padding + h/2 - h_text/2);
-        } else {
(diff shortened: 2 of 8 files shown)


tsipinakis:
Thanks for adding this! I'm a bit busy at the moment, you can expect a review around Wednesday.

From a very quick glance you also need to add the new setting to the `dunstrc` and do `docs/dunst.pod`.

chronus7:
> Thanks for adding this! I'm a bit busy at the moment, you can expect a review around Wednesday.

Sure, no stress. :)

> From a very quick glance you also need to add the new setting to the `dunstrc` and to `docs/dunst.pod`.

Ah, makes sense. I will look into it and update accordingly.

dunstrc
```diff
@@ -132,6 +132,10 @@
     # Possible values are "left", "center" and "right".
     alignment = left
 
+    # Vertical alignment of message text and icon.
+    # Possible values are "top", "center" and "bottom".
+    content_alignment = center
```
tsipinakis: Lets call it `vertical_alignment`, as IMO, `content_alignment` is a bit confusing when there's already a setting called `alignment`.

src/settings.c
```diff
@@ -423,6 +423,21 @@ void load_settings(char *cmdline_config_path)
...
+        {
+                char *c = option_get_string(
+                        "global",
+                        "content_alignment", "-content_alignment", "center",
+                        "Align icon and text top/center/bottom"
+                );
+                if (!string_parse_content_alignment(c, &settings.content_alignment)) {
+                        if (c)
+                                LOG_W("Unknown content alignment: '%s'", c);
+                        settings.content_alignment = defaults.content_alignment;
```
tsipinakis: You forgot to initialize `defaults.content_alignment` (`config.h` in root).

src/draw.c
```diff
@@ -511,32 +511,65 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width)
...
+                                break;
+                        case CONTENT_BOTTOM:
+                                text_y = h + settings.padding - h_text;
+                                if (text_y < 0) text_y = settings.padding;
+                                break;
+                        default:    // CONTENT_CENTER
+                                break;
+                }
+                // icon position
+                switch (settings.icon_position) {
```
tsipinakis: [nitpick] I'd prefer else-if chains rather than switches - it saves an indentation level and is generally more readable IMO.

src/draw.c
```diff
@@ -511,32 +511,65 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width)
...
+        // text positioning
+        if (cl->icon) {
+                // vertical alignment
+                switch (settings.content_alignment) {
+                        case CONTENT_TOP:
+                                text_y = settings.padding;
+                                break;
+                        case CONTENT_BOTTOM:
+                                text_y = h + settings.padding - h_text;
+                                if (text_y < 0) text_y = settings.padding;
```
tsipinakis: [nitpick] Please add a newline after if statements, it's hard to see the assignment there.

dunstrc
chronus7: I thought about it and came to the conclusion, that one could use `vertical_alignment` for alignment of the complete message window. And while `alignment` describes the alignment of the text, `content_alignment` then describes alignment of the message's content. But I concur, that it is not really obvious.

If you prefer, I will change it. I am not insisting on `content_alignment` in any way. :)

src/settings.c
chronus7: Ha. I looked for it, but didn't find it. Will do so.

src/draw.c
chronus7: Hm. Sure, no problem.

src/draw.c
chronus7: I kinda expected this. ^^ That was also not meant to make it into the final version originally.

dunstrc
tsipinakis: > I thought about it and came to the conclusion, that one could use vertical_alignment for alignment of the complete message window

In #336 we've decided that we'll call the new setting defining the actual window position `origin`, so there's no issue with using 'vertical_alignment' here.

Issue #669, opened 2019-10-25
Custom global actions


rosshadden:
Is there a way to add a custom global action that would be present in all notifications? I would like to add a `copy` action which would copy the text of the notification to my clipboard. For notifications created by me I would be able to specify a script and add this action, but most notifications on my system are not initiated by me.

tsipinakis:
Hi, this is not possible as actions are something the client (read: the program that sent the notification) does, the logic isn't being run on dunst, we only send back the action that was chosen.

However it might be a good idea to consider custom "actions" which in reality are scripts which are called in similar way to the `script=` directive in dunstrc. I'll leave this open for discussion as a feature request.

infokiller:
I'm looking for the same functionality as well. Use case:
I'd like to add a notification action to copy OTP codes from SMS messages. Given the notification text, a simple regex to match consecutive digits will probably work fine for me. So I could add a custom action to process the notification content and copy OTP codes to the clipboard.

hicklemon:
Being able to middle click the popup to copy the latest notification to the clipboard would be excellent.

AkashKrDutta:
+1 for this feature. It would be extremely helpful for me, to point out those notifications that I have missed from dunst (timed-out notifications). I could add a left-click/middle-click action to run a script to increase a shared counter variable. This counter can then be subtracted from `dunstctl history count` output to get count of those notifications missed from timeouts.

MultiCoreNop:
Another nice way would be an environment variable DUNST_ADD_ACTIONS (maybe a BASH array) that allows adding (command line, action name) pairs. This could then be filled from within a "script=".

yutkat:
I recently found the workaround to define custom actions and I will share it with you.

First, I call the shell script using `script=` in the condition that I want to call.
https://github.com/yutkat/dotfiles/blob/a6a483ded2912b9bda8ebaf790f59f5484a803c1/.config/dunst/dunstrc#L456-L459

Then, I could override the notification by adding `$DUNST_ID` to the `--replace-id` of `notify-send`.
Next, I can switch by action.
https://github.com/yutkat/dotfiles/blob/a6a483ded2912b9bda8ebaf790f59f5484a803c1/.config/dunst/scripts/libreoffice.sh#L9

bynect:
maybe something like `default_action` or `default_script` would do

PR #689, opened 2020-02-23
Added shortcut to copy formatted notification text via xclip

Jaywalker:
Provides the `copy` shortcut, which sends the notification text formatted as specified by `clipboard_format` to the program specified in `xclip`. 

Example of added config:
```
[global]
        clipboard_format = "%s - %b"
        xclip = /usr/bin/xclip -selection clipboard

[shortcuts]
        copy = ctrl+c

[copy-format-override]
        appname = Mumble
        clipboard_format = "%b"
```

### DIFF
--- a/config.h
+++ b/config.h
@@ -103,6 +103,10 @@ struct settings defaults = {
         .code = 0,.sym = NoSymbol,.is_valid = false
 },                              /* ignore this */
 
+.copy_ks = {.str = "none",
+        .code = 0,.sym = NoSymbol,.is_valid = false
+},                              /* ignore this */
+
 .mouse_left_click = MOUSE_CLOSE_CURRENT,
 
 .mouse_middle_click = MOUSE_DO_ACTION,
--- a/src/clipboard.c
+++ b/src/clipboard.c
@@ -1,0 +1,120 @@
+/* copyright 2020 Sascha Kruse and contributors (see LICENSE for licensing information) */
+
+#include "clipboard.h"
+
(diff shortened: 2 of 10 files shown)


tsipinakis:
I'm hesitant to merge this as-is as it's clearly designed to satisfy a single use-case, and introduces two new setting knobs in the process in an already complicated configuration file.

I'd much rather this be transformed into something like what's described in #669. Specifically, create a way to define custom actions for notifications that call a script when activated. With that, this usecase can be satisfied by using a script that calls xclip while additionally allowing many more similar hacks with notifications.

Now, that's obviously a much bigger chunk of work but it's the Right Thing To Do™ if we want to avoid future configuration restructures and compatibility breakages.
Would you be willing to work on something like that to get it merged?

infokiller:
@tsipinakis I'm curious why you think your suggestion is much more work. My thinking is that work required is adding configuration syntax to add custom actions, which basically call a script with the notification data (for example first arg is title, second arg is content, etc).
Am I missing something? Thanks!

tsipinakis:
No, that's pretty much it, but it's harder as you have to introduce some new concepts in the configuration and notification code, i.e. a list of all actions that can be run, a way to tell client actions and dunst actions apart etc.

If you're thinking of working on it don't be discouraged though, it's definitely doable and you can feel free to ping me if you need help.

infokiller:
Gotcha, thanks for the explanation. I'll see if I can carve out some time to work on this.

PR #692, opened 2020-03-01
Added override_redirect option

### LINKED ISSUE #627: Notification window hidden behind other windows in StumpWM when switching groups
Version: 1.3.2
Install type: AUR
Distro and version: Artix Linux
Window manager: StumpWM (master branch)
When you switch groups (i.e. workspaces) in StumpWM, the dunst notification window, if it's open, gets buried behind the other windows in the group, and will not raise back above them until the existing window is closed by dismissing all notifications. This seems like it may be tangentially related to #160 (or at least to some of the comments on it), and the corresponding StumpWM issue is https://github.com/stumpwm/stumpwm/issues/469.

### LINKED ISSUE #678: Dunst disappears when I move/resize a floating window in XMonad
I'm using XMonad as my WM.  Repro steps:

1. Start XMonad.

2. Open two windows: one a terminal, and the other Chromium.  Make sure they are both tiled.

3. Click `[Mod] + [Left mouse]` to convert Chromium into a floating window.  The terminal should stay tiled and should now occupy the whole screen, but partially obscured by Chromium.

4. Send a non-expiring notification via `notify-send -t 0 hi`.  The notification should show up above everything.

5. Now move the floating Chromium window in XMonad via `[Mod] + [Left mouse]`.  **The notification vanishes.** (Unexpected behavior.)  (If you move the terminal around, you'll see it's hiding under the terminal.)

6. In Chromium, open some non-empty web page, select some text, and then drag it.  The notification reappears in front of everything.

Not sure if this is an XMonad bug or Dunst one...

### Installation info

- Version: Dunst - A customizable and lightweight notification-daemon 1.4.1 (2019-07-03)
- Install type: Arch official repos
- Distro and version: Linux 5.4.6-arch3-1 #1 SMP PREEMPT Tue, 24 Dec 2019 04:36:53 +0000 x86_64 GNU/Linux

TAAPArthur:
To get notifications to reliably be on top of other windows, dunst
should ask the window manager. However, the override_redirect explicitly
tells the window manager to no manage notifications. An option has been
added to turn this behaviour off and is now the default in the sample
dunstrc.

We now properly rely on the window manager to keep our notifications on top which should close #678, close #627.

### DIFF
--- a/config.h
+++ b/config.h
@@ -25,6 +25,7 @@ struct settings defaults = {
               .negative_width = 0,
               .width_set = 0
             },
+.override_redirect = false,  /* prevent window manager from managing this window */
 .title = "Dunst",            /* the title of dunst notification windows */
 .class = "Dunst",            /* the class of dunst notification windows */
 .shrink = false,             /* shrinking */
--- a/config.h
+++ b/config.h
@@ -25,7 +25,7 @@ struct settings defaults = {
               .negative_width = 0,
               .width_set = 0
             },
-.override_redirect = false,  /* prevent window manager from managing this window */
+.override_redirect = true,  /* prevent window manager from managing this window */
 .title = "Dunst",            /* the title of dunst notification windows */
 .class = "Dunst",            /* the class of dunst notification windows */
 .shrink = false,             /* shrinking */
--- a/docs/dunst.pod
+++ b/docs/dunst.pod
@@ -142,9 +142,9 @@ the notification on the left border of the screen while a horizontal offset of
 
 =back
 
-=item B<override_redirect> (values: [true/false], default: false)
+=item B<override_redirect> (values: [true/false], default: true)
 
-Dicates whether or not the window managers manages notifications. This should be
+Dictates whether or not the window managers manages notifications. This should be
 false if you want notifications to always be on top.
 
 =item B<indicate_hidden> (values: [true/false], default: true)
--- a/docs/dunst.pod
+++ b/docs/dunst.pod
@@ -142,6 +142,11 @@ the notification on the left border of the screen while a horizontal offset of
 
 =back
 
+=item B<override_redirect> (values: [true/false], default: false)
+
+Dicates whether or not the window managers manages notifications. This should be
+false if you want notifications to always be on top.
+
 =item B<indicate_hidden> (values: [true/false], default: true)
 
 If this is set to true, a notification indicating how many notifications are
--- a/dunstrc
+++ b/dunstrc
@@ -31,6 +31,9 @@
     # screen width minus the width defined in within the geometry option.
     geometry = "300x5-30+20"
 
+    # When true the window manager won't manage this window
+    override_redirect = false
+
     # Show how many messages are currently hidden (because of geometry).
     indicate_hidden = yes
 
--- a/dunstrc
+++ b/dunstrc
@@ -32,7 +32,7 @@
     geometry = "300x5-30+20"
 
     # When true the window manager won't manage this window
-    override_redirect = false
+    override_redirect = yes
 
     # Show how many messages are currently hidden (because of geometry).
     indicate_hidden = yes
--- a/src/settings.c
+++ b/src/settings.c
@@ -256,9 +256,9 @@ void load_settings(char *cmdline_config_path)
 
         }
 
-        settings.override_redirect = !option_get_bool(
+        settings.override_redirect = option_get_bool(
                 "global",
-                "override_redirect", "-no-override_redirect", false,
+                "override_redirect", NULL, false,
                 "Determines if notifications are managed by the window manager."
         );
 
--- a/src/settings.c
+++ b/src/settings.c
@@ -256,9 +256,9 @@ void load_settings(char *cmdline_config_path)
 
         }
 
-        settings.override_redirect = option_get_bool(
+        settings.override_redirect = !option_get_bool(
                 "global",
-                "override_redirect", "-or/-override_redirect", defaults.override_redirect,
+                "override_redirect", "-no-override_redirect", false,
                 "Determines if notifications are managed by the window manager."
         );
 
--- a/src/settings.c
+++ b/src/settings.c
@@ -256,6 +256,12 @@ void load_settings(char *cmdline_config_path)
 
         }
 
+        settings.override_redirect = option_get_bool(
+                "global",
+                "override_redirect", "-or/-override_redirect", defaults.override_redirect,
+                "Determines if notifications are managed by the window manager."
+        );
+
         settings.shrink = option_get_bool(
                 "global",
                 "shrink", "-shrink", defaults.shrink,
--- a/src/settings.h
+++ b/src/settings.h
@@ -47,6 +47,7 @@ struct settings {
         char *icons[3];
         unsigned int transparency;
         struct geometry geometry;
+        bool override_redirect;
         char *title;
         char *class;
         int shrink;
--- a/src/x11/x.c
+++ b/src/x11/x.c
@@ -601,9 +601,10 @@ static void x_set_wm(Window win)
                 XInternAtom(xctx.dpy, "_NET_WM_STATE", false);
 
         data[0] = XInternAtom(xctx.dpy, "_NET_WM_STATE_ABOVE", false);
+        data[1] = XInternAtom(xctx.dpy, "_NET_WM_STATE_STICKY", false);
 
         XChangeProperty(xctx.dpy, win, net_wm_state, XA_ATOM, 32,
-                PropModeReplace, (unsigned char *) data, 1L);
+                PropModeReplace, (unsigned char *) data, 2);
 }
 
 GSource* x_win_reg_source(struct window_x11 *win)
--- a/src/x11/x.c
+++ b/src/x11/x.c
@@ -751,7 +751,7 @@ void x_win_hide(struct window_x11 *win)
         x_shortcut_ungrab(&settings.context_ks);
 
         XUngrabButton(xctx.dpy, AnyButton, AnyModifier, win->xwin);
-        XUnmapWindow(xctx.dpy, win->xwin);
+        XWithdrawWindow(xctx.dpy, win->xwin, XDefaultScreen(xctx.dpy));
         XFlush(xctx.dpy);
         win->visible = false;
 }
--- a/src/x11/x.c
+++ b/src/x11/x.c
@@ -643,7 +643,7 @@ struct window_x11 *x_win_create(void)
 
         root = RootWindow(xctx.dpy, DefaultScreen(xctx.dpy));
 
-        wa.override_redirect = true;
+        wa.override_redirect = settings.override_redirect;
         wa.background_pixmap = ParentRelative;
         wa.event_mask =
             ExposureMask | KeyPressMask | VisibilityChangeMask |


tsipinakis:
I'm against adding an explicit _setting_ for this as it's an intricate implementation detail of dunst and it doesn't make sense to be exposed in the config. Ideally, _it should just work_ :/. However, if it's the case that half of WMs handle these types of windows one way and the rest the others, it might be a necessary evil.

On the correct value of `override_redirect`, specifically: Does this fix the mentioned issues for you without any other side-effects (such as being able to move the window when you shouldn't)? Which of the mentioned WMs are you using/have tested?

`Override-redirect` basically tells X11 that this isn't an ordinary window and shouldn't be treated as such, with it disabled, i3 for example, (and other WMs) assume it's a simple application window and  add their own decoration which leads me to believe that disabling it is a step backwards.
2020-03-01T10-18-38

I'll take a deeper look into how these WMs handle window events and report back if I find anything.

tsipinakis:
I looked into the xmonad case specifically, see my [comment](https://github.com/xmonad/xmonad/issues/208#issuecomment-593082897) for more details

TAAPArthur:
To my understanding there are 2 trains of thought on window management: let the window manager manager windows or let the windows manage themselves. The EWMH/ICCCM seem to lead towards the former. But the "everything should just work" models and setting override-redirect seems to match the latter. If we wanted dunst notifications to actually always be on top and not rely on the window manager, we could set the override-redirect flag, listen for all configure/visibility events and manually raise windows. This should just "work" on all setups out of the box, but this would likely cause problems with other programs that always want to be on top and cause them to fight for supremacy. Also makes it harder for the user to handle their notifications from the WM (if they want to). These are some reasons why I support delegating this action to the WM, since most have this functionality built in already.

I'm using a custom window manager, but it behaves like Xmonad for this issue. When a window is clicked (maybe only for floating windows in Xmonad), it will be raised to the top of the stack, which may hide the notification windows. Any window manager that ever tries to raise a window to the top of the stack will run into this problem unless they explicitly handle these special cases.

I understand the concern for disabling. WMs may add decorations. For mine/Xmonad (which expect heavy initial customization), you can tell it to not decorate/ignore notification windows. I assume i3 and other WMs have similar abilities, but it may break some setups.  I'm not sure how you/the community feels about breaking existing setups vs having incorrect default behavior (from my perspective). I can live with having the default be the current behavior, but I strongly feel this setting should be exposed.

From [ICCCM spec](https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.10)

>If the window will be visible for a very short time and should not be decorated at all, the client can set override-redirect on the window. In general, this should be done only if the pointer is grabbed while the window is mapped. The window manager will never interfere with these windows, which should be used with caution. An example of an appropriate use is a pop-up menu. 
Advice to Implementors
    The user will not be able to move, resize, restack, or transfer the input focus to override-redirect windows, since the window manager is not managing them. If it is necessary for a client to receive keystrokes on an override-redirect window, either the client must grab the keyboard, or the client must have another top-level window that is not override-redirect and that has selected the Locally Active or Globally Active focus model. The client may set the focus to the override-redirect window when the other window receives a WM_TAKE_FOCUS message or one of the events listed in section 4.1.7 in the description of the Globally Active focus model.

tsipinakis:
> To my understanding there are 2 trains of thought on window management: let the window manager manager windows or let the windows manage themselves. The EWMH/ICCCM seem to lead towards the former.

Indeed I was aware of this, but if we're going to add a setting to disable `override_redirect` this, to me, means also implementing proper EMWH support. By 'proper' I mean that dunst should, in fully EMWH compliant WMs, work the same way as it did before the change without any manual user configuration. It makes no sense for 'install and use' WMs like i3 to require manual configuration for a notification window to be displayed correctly.

Another requirement would be that if EMWH is enabled by default dunst should be able to figure out if a WM supports EMWH to a level that would work for it and if not automatically switch back to `override_redirect`.

---

I'll focus on the `xmonad` case here as it's one of the issues you linked as fixed with this PR, and given its customizability it's a good example to take of a WM where managing the windows via the WM could be useful.

> When a window is clicked (maybe only for floating windows in Xmonad), it will be raised to the top of the stack, which may hide the notification windows. Any window manager that ever tries to raise a window to the top of the stack will run into this problem unless they explicitly handle these special cases.

Indeed but that's the problem here, `xmonad`, and apparently your WM as well, calls `XRaiseWindow` on focus/move, meaning it pulls the window to the top disregarding the other windows in the stack, so in this case what benefit does disabling `override_redirect` give if the WM itself doesn't consider the hints when managing the stacking order?
If this was done similar to how i3 does it, where they have an internal representation of the stack and update it only when necessary this wouldn't be an issue.

I have not used `xmonad` but I'm speculating that there's not really a way to fix this without source modifications.

---

On the topic of implementing EMWH support.
I read [the spec](https://developer.gnome.org/wm-spec/) more carefully and it doesn't look like it would support non-override-redirect notifications.

Specific points:
* Dunst (obviously) sets `TYPE_NOTIFICATION` as its window type, according to the spec
    > This property is typically used on override-redirect windows. 

    So EMWH itself is nudging us into using `override_redirect`.
* There is no way, according to the spec, to force WMs to not add decorations to dunst.
   I found this [thread from 2005](https://mail.gnome.org/archives/wm-spec-list/2005-June/msg00002.html) about adding a hint to disable decorations, and long story short the 2 solutions mentioned in that thread are:
    * Use `TYPE_DOCK`, not useful for dunst as docks have very different semantics and the window isn't placed correctly (in i3 for example, it reserves a space for the window at the top).
    * Use MOTIF hints, this would be the ideal solution but:
        * They aren't included in the EMWH spec, what's the point of switching if we have to use hacks to make it work?
        * It isn't exposed in `_NET_SUPPORTED` so dunst can't correctly decide whether to use plai n EMWH or `override_redirect`.

The spec also mentions that `_NET_WM_WINDOW_TYPE` is supposed to replace the MOTIF hints, but from what I have seen no WM has coded in the case to handle `TYPE_NOTIFICATION` windows (read: Disable decorations).

TAAPArthur:
Thanks for shedding light on how i3 works. I can see how it could minimize the problem, but I'd still assume this would be an issue in i3. It has to update the stack at some point which would obscure notifications; perhaps this isn't a concern because of the probably small window where this could happen. I'd guess one could test this by sending a _NET_ACTIVE_WINDOW client message while a notification is visible and see if it gets hidden...

What mine (and I think Xmonad) expect is for windows that want to be on top to set _NET_WM_STATE_ABOVE (which I believe we do currently). 

> _NET_WM_STATE_ABOVE indicates that the window should be on top of most windows (see the section called “Stacking order” for details). 

If Xmonad is different, then one could explicitly tell it to treat windows of a particular type (or name/class etc) to be above others with one line in the config file.

I'll concede the point about TYPE_NOTIFICATION windows having the OR bit set. I missed that. However, I still think adding this flag would be easier/simpler than changing the source of Xmonad and friends to handle notifications differently.

How about I change the default to keep the current behavior, and users could deliberately change it if they need to?

P.S. This fix would also a similar issue with StumpWM; I mis-formatted the link in the description

TAAPArthur:
I double checked the code. in src/x11/x.c we do set STATE_ABOVE property, but we don't set the STATE_STICKY also. I fixed this is the update in addition to changing the default value of override-redirect

tsipinakis:
Sorry for the delay in responding, I need to think a bit about this and look into it in more detail:

> I can see how it could minimize the problem, but I'd still assume this would be an issue in i3. It has to update the stack at some point which would obscure notifications; 

It's not in this case because IIUC they use the restack call rather than raise each window, so non-i3 managed windows remain unaffected.

> I still think adding this flag would be easier/simpler than changing the source of Xmonad and friends to handle notifications differently.
> How about I change the default to keep the current behavior, and users could deliberately change it if they need to?

I guess this is more of a philosophical question on how WMs are supposed to interact with override-redirect windows. Before I make a decision here I'd like to read the various specs more carefully to see if anything specific is mentioned for this case or if it's a general wild-west.

Another thing that I'd like to check is if, as you mentioned, the config languages of xmonad and StumpWM are powerful enough to actually fix this issue, because if it's impossible without source modifications what's the point of adding this anyway?

My general philosophy on adding new settings to the dunstrc is that it has to be general enough to allow for many different usecases and not be intended for just one very specific use. This is not out spite or anything, but the dunstrc and settings code is already _very_ bloated with too many settings and because of that I'm trying to limit to only those that are actually going to be used.

TAAPArthur:
I took a crack at implementating a solution with my WM that wouldn't require windows to be raised to the top of the global stack. I just have two guard windows and I raise/lower everything relative to them. So now when dunst appears as a OR window, it will stay on top. This seems similar to XRestackWindow, which doesn't seem to be available for xcb. Things can break when things are restarting and WMs like Xmonad need to restart (and recompile) to see changes to their config/settings. But I don't honestly think this will be a problem in practice.

That said I could live with the current behavior, if and only if dunst didn't try to keep itself on top. As you mention in [208](https://github.com/xmonad/xmonad/issues/208), if dunst and another window (who has OR set and wants to be on top) overlap, the user's gonna have a bad day.


> Another thing that I'd like to check is if, as you mentioned, the config languages of xmonad and StumpWM are powerful enough to actually fix this issue, because if it's impossible without source modifications what's the point of adding this anyway?

XMonad's config file is an actual haskell file. It is common for users to want to "float" windows by name/class which will prevent them from being tiled (not what the default behavior would be for no OR notifications). [Xmonad Floating windows](https://wiki.haskell.org/Xmonad/General_xmonad.hs_config_tips#ManageHook_examples). Window "decorations" aren't a thing from what I remember in Xmonad, if that was a concern. I could put together a minimal XMonad config if desired.

I understand your hesitations to add new settings, although I'd prefer too many over too few. I hope the hesitation is limited to the number of settings and not the extent of functionality. I do hope to extend the functionality of a few settings (like this one and idle_timeout) to influence the events dunst listens for. Like if override_redirect is false, then it doesn't need to listen for configure_notify events, but that will be a different merge request. 

Just want to highlight that we've been talking about 2-4 WMs but there are a [bunch](https://wiki.archlinux.org/index.php/Window_manager) out there. Most of them (probably all of the EWMH/ICCCM ones) could get the user's desired behavior if the notification didn't have the OR flag set.

tsipinakis:
> I just have two guard windows and I raise/lower everything relative to them. So now when dunst appears as a OR window, it will stay on top. This seems similar to XRestackWindow, which doesn't seem to be available for xcb

I don't even think the guard windows are necessary, i3 uses xcb as well, you can take a look at how they do it.

>That said I could live with the current behavior, if and only if dunst didn't try to keep itself on top. As you mention in 208, if dunst and another window (who has OR set and wants to be on top) overlap, the user's gonna have a bad day.

That's already the case, dunst will keep itself on top of normal windows but will not raise itself above override redirect windows.

> XMonad's config file is an actual haskell file. It is common for users to want to "float" windows by name/class which will prevent them from being tiled (not what the default behavior would be for no OR notifications). Xmonad Floating windows. Window "decorations" aren't a thing from what I remember in Xmonad, if that was a concern. I could put together a minimal XMonad config if desired.

Obviously all the things you mention are possible from a configuration standpoint, my concern is that the linked issues (e.g. xmonad fighting with dunst) might not be really fixable from our side.

> I understand your hesitations to add new settings, although I'd prefer too many over too few. I hope the hesitation is limited to the number of settings and not the extent of functionality.

Indeed that's the point, to not overstuff the config. Extending existing functionality is perfectly fine.

> Just want to highlight that we've been talking about 2-4 WMs but there are a bunch out there. Most of them (probably all of the EWMH/ICCCM ones) could get the user's desired behavior if the notification didn't have the OR flag set.

I have my doubts on that - as I said above it seems like EWMH isn't really designed for non-override redirect notifications, so if this works in any WMs it's going to require a lot of customization to be usable.

TAAPArthur:
> I don't even think the guard windows are necessary, i3 uses xcb as well, you can take a look at how they do it.
From the quick glance of i3's code I think their model of the window stack differs as in they look to be in sync with the actual hierarchy of windows.

> That's already the case, dunst will keep itself on top of normal windows but will not raise itself above override redirect windows.
That's good to hear. I tried to keep my original comment general, but some other programs are less smart. First things that comes to mine is onboard which isn't an OR window, but wants to keep itself on top. The window would likely be a floating (as in the WM would allow it to raise itself) and would end of fighting with dunst. I think the true problem lies within the other program, but  I'm sure other software does this. I'm not actually saying we should patch dunst just to work around bugs in other code bases (I just patch onboard in my case); just illustrate some annoying cases and why I lean towards not having apps trying to force themselves on top.

> I have my doubts on that - as I said above it seems like EWMH isn't really designed for non-override redirect notifications, so if this works in any WMs it's going to require a lot of customization to be usable.
So if I provide a minimal XMonad config file where everything works, will that seal the deal? I have high confidence it will be easy, but I'll have to brush up on my Haskell. 

I think its been a good discussion (I at least learned something), but I hope we can close this issue soonish.

tsipinakis:
I've been thinking about this a lot whether this actually makes sense to add. I'm now convinced that this isn't really a good idea.
I'll list my thoughts here for transparency:

* As I mentioned above to add new settings in the dunstrc I'm aiming for them to either be very often requested feature, or at least be general enough so that it's not just used once and forgotten.

* From what I've seen, this has _no_ other usecases rather than fixing this weird window stacking quirk, everything else you could do from the WM like move the window with the focus etc can already be done from the dunstrc.

* Dunst isn't the only application using override redirect, it's used widely and for many different purposes, does it really make sense to add a switch to disable it in just one program rather than deal with the issue at the core (i.e. WM or X11 level)?

And additionally there's all the arguments mentioned above:

* The EWMH hints is pushing us to have the notifications as OR anyway
* There is no way to implement full EWMH support so that when the OR switch is flipped dunst, consistently works the same way in all WMs, from that point on it just becomes a configuration mess.



So I'll close this for now, obviously I am still open to discussion of the points mentioned above.

PR #695, opened 2020-03-20
Added all shortcuts as commands.

Czaplicki:
For external bindings and scripting.
(Example : sxhkd)


tsipinakis:
Sorry, this has been proposed before, and given that there's #445, I don't see a reason to extend a hack that is going to be deprecated/removed after.

PR #699, opened 2020-03-30
Invalidate actions after notification is closed

tsipinakis:


bebehei:
Can we kick out travis CI? One shouldn't care in a CI, why the status isn't reported. (And maybe kick out CircleCI, not that it's a bad platform, but it can get easliy replaced with GitHub Actions, which also has a native integration into GitHub)

tsipinakis:
> There is already a dbus_valid field, which has now a feature overlap.

I had forgot about this, this made me realise that this is call is probably best inside the `signal_notification_closed` function, since that's where the actual invalidating happens.

tsipinakis:
> Can we kick out travis CI?

I just removed it from the org, I haven't looked into github actions yet but if they work just as well I don't see a reason not to remove CircleCI as we..

bebehei:
> I just removed it from the org, I haven't looked into github actions yet but if they work just as well I don't see a reason not to remove CircleCI as we..

Well I'm currently already on it. The stuff with the docker images is straight forward. The only tricky part is the coveralls integraiton.

bebehei:
The latest set of docker images will get built now. It'll finish in about 2 hours. Dockerhub is slow.

Afterwards, I'll open a PR [from my `ci` branch.](https://github.com/dunst-project/dunst/compare/master...bebehei:ci).

PR #651, opened 2019-08-11
Implement a command line control (dunstctl)

bebehei:
Work in progress, I expect to be finish & polish it during the next week.

## Fixes:

- [ ] #308 
- [ ] #307 
- [ ] #216
- [ ] #171 
- [ ] #103 
- [ ] #112 
- [ ] #138 
- [ ] #284 
- [ ] #241 
- [ ] #445

## TODO:

- [ ] Rewrite the commit messages
- [ ] test in dunstctl, if we're actually speaking with a compatible dunst instance
- [ ] add a debug subcommand to `dunstctl` to support debugging later.

PR #701, opened 2020-04-03
Install dunstify

braunbearded:
Put dunstify in $PATH, so you can run it easily without manually moving files, like dunst itself

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,7 +155,7 @@ clean-coverage-run:
         install-service install-service-dbus install-service-systemd \
         uninstall \
         uninstall-service uninstall-service-dbus uninstall-service-systemd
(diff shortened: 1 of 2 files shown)


bebehei:
I guess you have to fix the tests :wink: You have to make changes here: https://github.com/dunst-project/dunst/blob/master/test/test-install.sh

braunbearded:
Fixed: Dont know how I missed that

tsipinakis:
I have rejected similar PRs before on the notion that `dunstctl` aka #651 would be feature-equivalent with dunstify and one could use that instead.
This doesn't look to be the case looking at the current implementation in that PR, so I agree here, let's add dunstify officially in.

PR #702, opened 2020-04-03
Fix dunstify installation

braunbearded:
As it turns out, not every distribution link (lib)notify by default, so I didnt catch that earlier. I separate install and install-dunstify again because autoconf checks if string "dunstify" is applied as a make argument and then add libnotify to pkg_config_pack.

It would be awesome if someone could update the wiki page on this topic.

I am kinda new to C "building"/autoconf so I didnt fully understand respectively underestimated my change, sorry for that.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -48,6 +48,4 @@ $(warning STATIC_CONFIG is deprecated behavior. It will get removed in future re
 endif
 
 # dunstify also needs libnotify
-ifneq (,$(findstring dunstify,${MAKECMDGOALS}))
-	pkg_config_packs += libnotify
-endif
+pkg_config_packs += libnotify


test/test-install.sh
```diff
@@ -23,3 +22,36 @@ if ! [ -z "$(find "${PREFIX}" -type f)" ]; then
         find "${PREFIX}" -type f
         exit 1
 fi
+
```
bebehei, line 25: This patch doesn't look right.

braunbearded: Maybe it is a little bit over the top but it tests "every" make scenario for that change

braunbearded: Ahh I see, do you mean the single dunstify test with the systemd options and stuff?

bebehei: Well, I do not read code in a review, which is obviously copied multiple times. How am I supposed to find the bits, which are different? How am I supposed understand the differences in the future? Would a comment help? (spoiler: no) ...

Keep your IT DRY. [Here's a perfect article to summarize it.](https://michael.lustfield.net/pages/keeping-it-dry)

tsipinakis: I'm not sure I get what's the point of testing dunstify installation separately, it doesn't have any special cases to test for?

The mistake here was probably not modifying `config.mk` to always include libnotify in the dependencies.

tsipinakis:
> As it turns out, not every distribution link (lib)notify by default, so I didnt catch that earlier. 

Obviously, as we essentially added a new build-dependency to dunst by compiling dunstify by default, this means downstream maintainers will have to add libnotify to the dependency lists for dunst to build correctly.

I think that's a non-issue generally because if you're using dunst it's pretty much guaranteed that something in the system is going to be using libnotify.

As I mentioned in the review comments, all you need to do is explicitly specify that `libnotify` is now a base dependency with the patch given.

> I am kinda new to C "building"/autoconf so I didnt fully understand respectively underestimated my change, sorry for that.

And don't apologize I'm the one who merged it :)

test/test-install.sh
braunbearded: > How am I supposed to find the bits, which are different? How am I supposed understand the differences in the future? Would a comment help? (spoiler: no) ...
> 

You are completely right, I didn't think about that. Thanks for the article!

braunbearded:
> Obviously, as we essentially added a new build-dependency to dunst by compiling dunstify by default, this means downstream maintainers will have to add libnotify to the dependency lists for dunst to build correctly.

Exactly what I thought but I wasn't sure about it.

PR #705, opened 2020-04-05
Allow for multiple actions per mouse event

### LINKED ISSUE #690: Allow for multiple actions per mouse event
I'd like to request a feature to allow for users to specify a list of actions to perform for each mouse event (left, middle, and right click). To fit with INI formatting, it would be a comma-separated list, and the actions would be executed in-order left to right as listed. 

My use case for this is that I want my left click to perform an action/bring up the context menu, and then also close the current notification. As I have it configured right now, left clicking on a notification performs the action and then leaves it on the screen, which to me feels unnatural.

mkrasnitski:
This PR introduces functions for parsing configuration options provided in `dunstrc` or on the command line as comma-separated lists of one or more values. Elements are separated by single commas, and trailing whitespace around commas is trimmed, although whitespace in the middle of elements is preserved. Additionally, it makes use of this new parsing by changing the `mouse_left_click`, `mouse_middle_click`, and `mouse_right_click` options to be lists instead of just strings. Current config files are not broken by this. Some tests are provided, though any suggestions or improvements to better the style and code coverage of said tests is appreciated.

Implements #690.

### DIFF
--- a/src/option_parser.c
+++ b/src/option_parser.c
@@ -141,6 +141,27 @@ bool string_parse_mouse_action(const char *s, enum mouse_action *ret)
         return false;
 }
 
+bool string_parse_mouse_action_list(char **s, enum mouse_action **ret)
+{
+        ASSERT_OR_RET(s, false);
+        ASSERT_OR_RET(ret, false);
+
+        int len = 0;
+        while (s[len])
+                len++;
+
+        *ret = g_malloc((len + 1) * sizeof(enum mouse_action));
+        for (int i = 0; i < len; i++) {
+                if (!string_parse_mouse_action(s[i], *ret + i)) {
+                        LOG_W("Unknown mouse action value: '%s'", s[i]);
+                        g_free(*ret);
+                        return false;
+                }
+        }
+        (*ret)[len] = -1; // sentinel end value
+        return true;
+}
+
 bool string_parse_sepcolor(const char *s, struct separator_color_data *ret)
 {
         ASSERT_OR_RET(STR_FULL(s), false);
@@ -484,7 +505,8 @@ char *cmdline_get_path(const char *key, const char *def, const char *description
                 return string_to_path(g_strdup(def));
 }
 
-char **cmdline_get_list(const char *key, const char *def, const char *description){
+char **cmdline_get_list(const char *key, const char *def, const char *description)
+{
         cmdline_usage_append(key, "list", description);
         const char *str = cmdline_get_value(key);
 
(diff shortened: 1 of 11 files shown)


mkrasnitski:
I would also like suggestions on changing the description of these options in the default `dunstrc`. Would it be prudent to make one of the options a multi-element list by default, or should the comment describing them be edited?

src/option_parser.c
```diff
@@ -141,6 +141,27 @@ bool string_parse_mouse_action(const char *s, enum mouse_action *ret)
...
+bool string_parse_mouse_action_list(char **s, enum mouse_action **ret)
+{
+        ASSERT_OR_RET(s, false);
+        ASSERT_OR_RET(ret, false);
+
+        int len = 0;
+        while (s[len])
+                len++;
+
+        *ret = g_malloc((len + 1) * sizeof(enum mouse_action));
```
bebehei, line 153: Please use `g_n_malloc()` it avoids buffer overflows.

src/utils.c
```diff
@@ -135,6 +146,27 @@ void string_strip_delimited(char *str, char a, char b)
         str[iwrite] = 0;
 }
 
+/* see utils.h */
+char **string_to_array(const char *string)
```
bebehei, line 150: I guess you implemented `g_strsplit()` here, right?

bebehei (dismissed):

src/option_parser.c
bebehei: Sorry, it's called `g_malloc_n()`. See here: https://developer.gnome.org/glib/stable/glib-Memory-Allocation.html#g-malloc-n

bebehei:
But anyways, all in all the code looks nice and polished. I've seldomly seen a PR written with perfect style! Keep up the good work :+1:

src/option_parser.c
mkrasnitski: Thanks, will do.

src/utils.c
mkrasnitski: Wow, I'm a dummy, haha. It was right there in front of me! Somehow I didn't see that function, since glib documentation doesn't come up too much when googling.

mkrasnitski:
Thanks for the encouragement! Fixes should be up. I decided to keep `string_to_array` as a wrapper function, since `g_strsplit` doesn't trim whitespace from tokens.

tsipinakis (approved):
As @bebehei said, it's rare to see a perfect PR from the get-go. Keep up the good work!

tsipinakis:
> Would it be prudent to make one of the options a multi-element list by default, or should the comment describing them be edited?

I think that's fine as you've put it. In my case all programs I've come across close the notification by themselves when an action is triggered, so I'd consider this the 'default behavour'.

PR #708, opened 2020-04-27
follow=keyboard: Fall back to follow=mouse instead of XDefaultScreen()

cdown:
In dwm and similar window managers, it's common to often have empty
tags, and navigate to them with the intent to create clients there. If I
navigate to one of those empty tags with a dunst notification visible
and follow=keyboard set, the notification warps over to the default
screen.  If I then open a client, it then warps back, which is pretty
jarring.

This is mostly an artefact of the implementation of follow=keyboard --
if we fail to get a focused window, we use the default screen. However
this case isn't necessarily really a "failure" on window managers like
dwm where it's a common occurrence to end up with no clients on the
screen, whereas that would be significantly rarer on (say) GNOME or KDE.

A guess that's more likely to fit user expectations is falling back to
where the mouse pointer currently is, since this indicates the currently
focused monitor that the window manager would create a client on. This
avoids warping back to that monitor again when a client is created.

### DIFF
--- a/src/x11/screen.c
+++ b/src/x11/screen.c
@@ -303,6 +303,7 @@ bool window_is_fullscreen(Window window)
 struct screen_info *get_active_screen(void)
 {
         int ret = 0;
+        bool force_follow_mouse = false;
         if (settings.monitor > 0 && settings.monitor < screens_len) {
                 ret = settings.monitor;
                 goto sc_cleanup;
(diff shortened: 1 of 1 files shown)


cdown:
Video demonstration of the problem and solution, with dunst HEAD first, and dunst with this patch after:

https://www.youtube.com/watch?v=STXILe-udt0

cdown:
Thanks for the quick review!

Issue #709, opened 2020-05-01
Allow toggle paused status via dunstctl


mrossinek:
### Feature Idea

I just started taking a look at `dunstctl` from #651 since I have been waiting for the PR to be merged.
I would like to be able to simply toggle the paused state of dunst with a single command. Since I think this would also be useful to other users I think it would be an easy addition to the `dunstctl` shell script.

If you like, I am happy to implement this myself and open a PR. If so, which approach would you like:
- add another argument option to `set-running` e.g. `toggle`
- set this as the default when calling `dunstctl set-running` without additional arguments
- add a new command entirely

If you do not agree that this should be integrated into `dunstctl` directly I will opt to write a wrapper myself, I just wanted to put this out there.

tsipinakis:
I agree this would be useful (and been requested before as an addition to the `DUNST_COMMAND_...` hacks if I remember right). 

> If you like, I am happy to implement this myself and open a PR.

If you have the time to work on it, please do!

> If so, which approach would you like:

I'd prefer `dunstctl set-running toggle`, it's best to be explicit here in my opinion.

mrossinek:
@tsipinakis Okay will do. I will fix #710 too, while I'm at it.

tsipinakis:
Thank you! If you have any questions or get stuck somewhere feel free to ask.

PR #711, opened 2020-05-01
Allow toggling pause state via dunstctl

### LINKED ISSUE #710: dunstctl: Rename set-running to set-paused
After looking at #709  the name `set-running` sounds a bit odd to me. It was named like this as this is our internal status name. But for the user facing side `set-paused` is probably more clear.

Bit unfortunate that I noticed this after merging, but thankfully it hasn't been officially released yet so no harm in directly changing it in master.

mrossinek:
Implements the feature proposed in #709 to allow toggling the pause state via `dunstctl`.
This also fixes #710 to improve user intuition.

### DIFF
--- a/dunstctl
+++ b/dunstctl
@@ -45,11 +45,11 @@ property_set() {
 }
 
 invert_boolean() {
-        if [ "${1}" = "true" ]; then
-            printf "%s\n" "false"
-        elif [ "${1}" = "false" ]; then
-            printf "%s\n" "true"
-        fi
+	if [ "${1}" = "true" ]; then
+		printf "%s\n" "false"
+	elif [ "${1}" = "false" ]; then
+		printf "%s\n" "true"
+	fi
 }
 
 command -v dbus-send >/dev/null 2>/dev/null || \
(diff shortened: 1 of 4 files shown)


dunstctl
```diff
@@ -64,15 +72,29 @@ case "${1:-}" in
...
+                    running=$(property_get running | ( read -r _ _ running; printf "%s\n" "${running}"; ))
+                    if [ "${running}" = "true" ]; then
+                        property_set running variant:boolean:false
+                    else
+                        property_set running variant:boolean:true
+                    fi
+                else
+                    # invert boolean to indiciate running status rather than pause one
+                    running=$(invert_boolean "${2}")
+		    property_set running variant:boolean:"${running}"
```
tsipinakis, line 96: What's up with the indentation here?

Also: dunstctl uses tabs for indentation, not spaces. It looks like you forgot to match your editor settings.

mrossinek:
@tsipinakis Will fix the code style in a second. Could you take a look at my other comments, if you have a sec?

tsipinakis:
What other comments do you mean?

dunstctl
mrossinek: Done.

dunstctl
```diff
@@ -64,15 +72,29 @@ case "${1:-}" in
 	"history-pop")
 		method_call "${DBUS_IFAC_DUNST}.NotificationShow" >/dev/null
 		;;
-	"running")
-		property_get running | ( read -r _ _ paused; printf "%s\n" "${paused}"; )
+	"is-paused")
+                running=$(property_get running | ( read -r _ _ running; printf "%s\n" "${running}"; ))
```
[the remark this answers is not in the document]
tsipinakis: I'd say the cleanest solution here is to invert the boolean server side. No need to brainstorm too much since we're the only ones using this interface.

Just do the same running->paused change in `src/dbus.c` and invert the return values of the Set/Get functions.

What do you think?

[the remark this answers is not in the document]
mrossinek: Ah I didn't know that these property getter and setter functions were only added together with dunstctl. I will change it on the server side then and also rename those properties from running to paused if you agree.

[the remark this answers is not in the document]
tsipinakis: It's cleaner that way so go ahead.

src/dbus.c
```diff
@@ -588,8 +588,8 @@ GVariant *dbus_cb_dunst_Properties_Get(GDBusConnection *connection,
 {
         struct dunst_status status = dunst_status_get();
 
-        if (STR_EQ(property_name, "running")) {
-                return g_variant_new_boolean(status.running);
+        if (STR_EQ(property_name, "paused")) {
+                return !g_variant_new_boolean(status.running);
```
tsipinakis, line 592: You're inverting the return pointer, not the boolean value. This should be
```c
                 return g_variant_new_boolean(!status.running); 
```

mrossinek: Sorry my bad. Fixed below

tsipinakis: I see the fix on the XML but not this, perhaps you forgot to commit it?

mrossinek: Yep it was still unstaged..

tsipinakis:
LGTM :+1:

PR #713, opened 2020-05-03
Fix ugly frame

nick87720z:
1. Original shaping mask caused antialiasing pixels to be cut, leading to ugly look. Solved by increasing mask corder radus by half-borderwidth, but not less than 1pix (for zero-width border).
2. And when border width is set to zero, it should not be painted. Current approach caused transparent pixels from AA to sum, producing subtle edge-like artifact at corners.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -388,6 +388,29 @@ static int layout_get_height(struct colored_layout *cl)
         return MAX(h, h_icon);
 }
 
+/* Attempt to make internal radius more organic.
+ * Simple r-w is not enough for too small r/w ratio.
+ * simplifications: r/2 == r - w + w*w / (r * 2) with (w == r)
+ * r, w - corner radius & frame width,
+ * h  - box height
+ */
+static int frame_internal_radius (int r, int w, int h)
+{
+        // Integer precision scaler, using 1/4 of int size
+        const int s = 2 << (8 * sizeof(int) / 4);
+
+        int r1, r2, ret;
+        h *= s;
+        r *= s;
+        w *= s;
+        r1 = r - w + w * w / (r * 2);    // w  <  r
+        r2 = r * h / (h + (w - r) * 2);  // w  >= r
+
+        ret = (r > w) ? r1 : (r / 2 < r2) ? r / 2 : r2;
+
+        return ret / s;
+}
+
 /**
  * Create a path on the given cairo context to draw the background of a notification.
  * The top corners will get rounded by `corner_radius`, if `first` is set.
@@ -485,22 +508,7 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
                 else
                         height -= settings.separator_height;
 
-                /* Attempt to make internal radius more organic.
-                 * Simple r-w is not enough for too small r/w ratio.
-                 * simplifications: r/2 == r - w + w*w / (r * 2) with (w == r)
-                 **/
-                {       const int s = 2 << (sizeof(int) * 8 / 4); // Integer precision scaler
-                        int h, w, r, r1, r2;
-
-                        h = s * height;
-                        r = s * corner_radius;
-                        w = s * settings.frame_width;
-                        r1 = r - w + w * w / (r * 2);    // w < r;
-                        r2 = r * h / (h + (w - r) * 2);  // w >= r
-
-                        radius_int = (r > w) ? r1 : (r / 2 < r2) ? r / 2 : r2;
-                        radius_int /= s;
-                }
+                radius_int = frame_internal_radius (corner_radius, settings.frame_width, height);
         }
 
         cairo_set_source_rgb(c, cl->bg.r, cl->bg.g, cl->bg.b);
--- a/src/draw.c
+++ b/src/draw.c
@@ -464,25 +464,25 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
         else
                 height += settings.separator_height;
 
-	if (settings.frame_width > 0)
+        if (settings.frame_width > 0)
         {
-            cairo_set_source_rgb(c, cl->frame.r, cl->frame.g, cl->frame.b);
-            draw_rounded_rect(c, x, y, width, height, corner_radius, first, last);
-            cairo_fill(c);
+                cairo_set_source_rgb(c, cl->frame.r, cl->frame.g, cl->frame.b);
+                draw_rounded_rect(c, x, y, width, height, corner_radius, first, last);
+                cairo_fill(c);
 
-            /* adding frame */
-            x += settings.frame_width;
-            if (first) {
-                    y += settings.frame_width;
-                    height -= settings.frame_width;
-            }
+                /* adding frame */
+                x += settings.frame_width;
+                if (first) {
+                        y += settings.frame_width;
+                        height -= settings.frame_width;
+                }
 
-            width -= 2 * settings.frame_width;
+                width -= 2 * settings.frame_width;
 
-            if (last)
-                    height -= settings.frame_width;
-            else
-                    height -= settings.separator_height;
+                if (last)
+                        height -= settings.frame_width;
+                else
+                        height -= settings.separator_height;
         }
 
         cairo_set_source_rgb(c, cl->bg.r, cl->bg.g, cl->bg.b);
--- a/src/draw.c
+++ b/src/draw.c
@@ -454,6 +454,7 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
                                           int *ret_width)
 {
         int x = 0;
+        int radius_int = corner_radius;
 
         cairo_t *c = cairo_create(srf);
 
@@ -483,10 +484,27 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
                         height -= settings.frame_width;
                 else
                         height -= settings.separator_height;
+
+                /* Attempt to make internal radius more organic.
+                 * Simple r-w is not enough for too small r/w ratio.
+                 * simplifications: r/2 == r - w + w*w / (r * 2) with (w == r)
+                 **/
+                {       const int s = 2 << (sizeof(int) * 8 / 4); // Integer precision scaler
+                        int h, w, r, r1, r2;
+
+                        h = s * height;
+                        r = s * corner_radius;
+                        w = s * settings.frame_width;
+                        r1 = r - w + w * w / (r * 2);    // w < r;
+                        r2 = r * h / (h + (w - r) * 2);  // w >= r
+
+                        radius_int = (r > w) ? r1 : (r / 2 < r2) ? r / 2 : r2;
+                        radius_int /= s;
+                }
         }
 
         cairo_set_source_rgb(c, cl->bg.r, cl->bg.g, cl->bg.b);
-        draw_rounded_rect(c, x, y, width, height, corner_radius, first, last);
+        draw_rounded_rect(c, x, y, width, height, radius_int, first, last);
         cairo_fill(c);
 
         if (   settings.sep_color.type != SEP_FRAME
--- a/src/draw.c
+++ b/src/draw.c
@@ -416,7 +416,7 @@ static int frame_internal_radius (int r, int w, int h)
  * The top corners will get rounded by `corner_radius`, if `first` is set.
  * Respectably the same for `last` with the bottom corners.
  */
-static void draw_rounded_rect(cairo_t *c, int x, int y, int width, int height, int corner_radius, bool first, bool last)
+void draw_rounded_rect(cairo_t *c, int x, int y, int width, int height, int corner_radius, bool first, bool last)
 {
         const float degrees = M_PI / 180.0;
 
--- a/src/draw.c
+++ b/src/draw.c
@@ -464,23 +464,26 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
         else
                 height += settings.separator_height;
 
-        cairo_set_source_rgb(c, cl->frame.r, cl->frame.g, cl->frame.b);
-        draw_rounded_rect(c, x, y, width, height, corner_radius, first, last);
-        cairo_fill(c);
+	if (settings.frame_width > 0)
+        {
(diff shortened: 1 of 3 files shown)


src/draw.c
```diff
@@ -464,24 +464,27 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
...
-        draw_rounded_rect(c, x, y, width, height, corner_radius, first, last);
-        cairo_fill(c);
-
-        /* adding frame */
-        x += settings.frame_width;
-        if (first) {
-                y += settings.frame_width;
-                height -= settings.frame_width;
+	if (settings.frame_width > 0)
+        {
```
tsipinakis, line 468: [code style] Braces should be in the same line as the statement

src/draw.c
```diff
@@ -464,24 +464,27 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
...
-        cairo_set_source_rgb(c, cl->frame.r, cl->frame.g, cl->frame.b);
-        draw_rounded_rect(c, x, y, width, height, corner_radius, first, last);
-        cairo_fill(c);
-
-        /* adding frame */
-        x += settings.frame_width;
-        if (first) {
-                y += settings.frame_width;
-                height -= settings.frame_width;
+	if (settings.frame_width > 0)
```
tsipinakis, line 467: [code style] I'm placing the comment here but it's about the entire PR: The indentation is inconsistent, we're using 8 spaces per indentation level.

src/x11/x.c
```diff
@@ -110,10 +110,13 @@ static void x_win_round_corners(struct window_x11 *win, const int rad)
 
         /* To mark all pixels, which should get exposed, we
          * use a circle for every corner and two overlapping rectangles */
+	int fw = settings.frame_width / 2 + settings.frame_width % 2;
+	if (fw < 1)
+		fw=1;
```
tsipinakis, line 115: [code style] Please use a space before and after the operator `fw = 1`

tsipinakis:
After a closer look I found a bug. There seems to be some artifacts on all corners except top left.
2020-05-03T20-55-24
2020-05-03T20-57-47

That's obviously not a bug with your code but it exposed an existing issue, but still I'd prefer it to be fixed before merging.

tsipinakis:
On closer look this does look like a bug, there are no rounded corners on the window shape, so most likely some of the math is off.

nick87720z:
One problem, previously invisible - when some notifications are stacked (even if e.g. startup low priority notification disappeared with new appeared before), it looks as if they all are rendered, with same effect at antialiased edge as with 0px edge before second commit. You might want to comment out shaping call to see, how it's rendered when in stack. No even need to trigger notifications, just recall one many times. I guess, it assumes that notification sequence is in vertical row, so when they are stacked, it looks even more strange).

As for mask painting (if you are about x_win_round_corners()) it is rounded, but somewhat ineffective. Two crossing rectangles, yet circles are full 360deg, while 1/4 would be enough. Though could be more effective in terms of xserver calls, I'm not sure).

Yet can't understand, where 0.5,0.5 displacement is set (originall integer coordinates would point to pixel corner instead of its center).

nick87720z:
I casually got to making transparent (argv) override-redirect window, like xfce4-notifyd has. But even better. xfce4-notifyd seems not using xshape, only relying to alpha chan. When some compositing manager (compton) tries to make its shadow, with rounded corners there are just gaps, between square shadow and rounded window edge (although this chan affects to rounded corner detection). XShape(ing) affects shadow geometry as well.

src/x11/x.c
```diff
@@ -110,10 +110,13 @@ static void x_win_round_corners(struct window_x11 *win, const int rad)
...
+        int fw = settings.frame_width / 2 + settings.frame_width % 2;
+        if (fw < 1)
+                fw = 1;
         unsigned const int centercoords[] = {
-                0,               0,
-                width - dia - 1, 0,
-                0,               height - dia - 1,
+                -fw,             -fw,
+                width - dia - 1, -fw,
+                -fw,             height - dia - 1,
```
tsipinakis, line 119: > Yet can't understand, where 0.5,0.5 displacement is set (originall integer coordinates would point to pixel corner instead of its center).

This was bothering me since the beginning and I finally figured out why: According to the [X11 docs](https://tronche.com/gui/x/xlib/graphics/filling-areas/XFillArc.html), the coordinates in XFillArc "specify the upper-left corner of the bounding rectangle", so at (0,0) this includes the frame, this change over-compensates and moves the arcs too much, breaking the round corners if the frame is large enough.

I'd only put 1 or 2 pixels of leeway here. The transparency changes are awesome, but they only work if there's a running compositor. Limiting the number of transparent pixels preserves the AA effect but minimizes the black artifacts in the corners.

nick87720z: It's not center, XDrawArc takes rectangle coords, containing arc. Did not test all frame width, only 1 and 0. Imho not sure, what's better - aliased border or several too dark pixels. Comparing with Xfce4-notifyd it just disabled rounding for no-comp.

nick87720z:
Oh, and with compositing and compton shadow I got white artifacts due to some transparent pixels, above white areas)))
dunst-rounding-comp-shape

nick87720z:
I was slow, thinking cairo had 0.5,0.5 displacement and not noticing, that there are no strokes))).
1pix outside is right enough to contain antialiasing. Btw, bitmap could be filled using cairo with its subpixel coordinate system, to even better tune frame->mask stroke margin (e.g. value 0.9 or 0.8 could drop some pixel with enough transparency).

tsipinakis:
> e.g. value 0.9 or 0.8 could drop some pixel with enough transparency).

Dropping it in cairo wouldn't help us here though, we need a way to remove them from the window/on the X11 side to fix the shadow artifacts.

This seems like that last hurdle to have this working correctly. Do you know if there are any functions in the XShape extension to do that?
If no we may just have to live with these artifacts.

nick87720z:
As far as I understand, xshape is 1-bit alpha channel implementation, available without compositing. Applied above all rgb/rgba content. Initial implementation was fine in no-comp, if only paintings were not accumulating (it seems, that window appeared, its content was just a garbage from what was there before, not a true attempt to use background for transparency).
It's my first attempt to touch lower level windowing toolkit than qt/gtk :)

I tried to get internal radius more natural, but could not get satisfied with simple radius-fw difference.
Still tried to stick with integer arithmetics.

nick87720z:
One problem (was) ugly xlib drawing quality, even at 1-bit cairo surpasses it. On case less to use xlib drawing. I did not even add that extra pixel for antialiased border, just repeated border painting with given radius against bitmap surface.

nick87720z:
While configuring picom I randomly triggered xfce4-notifyd without compositing. I was certainly wrong about it not using xshape, yet its quality was same as on my last shots above.

Just to not mimic xfce policy about config options, there could be option to force xshape in compositing mode if anyone likes shadows under notifications. But I'm not sure it's necessary. Same picom (former compton, so blamed for ugly shadows with rounded argb windows) now has option for custom shader, could be used for custom shado. Yet there is full-shadow toggle for winlist block (though not a place for arbitrary rules like for "shadow-exclude").

tsipinakis:
From my tests this looks perfect now! There are still some artifacts but they're barely visible IMO.

> Just to not mimic xfce policy about config options, there could be option to force xshape in compositing mode if anyone likes shadows under notifications. 

From a quick test with compton, the shadows were rounded and looked like they matched the window. So I'm not sure if there's anything wrong there.
In either cause for this I'll say let's wait until someone complains to add that config option, no reason to over-complicate :).

nick87720z:
While learning code, I noticed, that all drawing is done to image surface. What if it were done to recording surf as some way to draw right to xlib surf? When I tried so for self, I did not notice any visible difference. Of I'm aware that not all cairo ops work via backend, as case with opengl backend (remebering some notes about gradients).

nick87720z:
`*8/4` - to make numbers less magic (I know both formulas are already like magic)))).
Trial and error part was only partial, thanks to gnuplot (last time I simply made surface, involving both radius and border width as x/y parameters). I tried first to make first formula to softly fade to second, but it went up too quickly after (bw > r). It's really hard to imagine them without origin, written with math graphics for ratios, sub/super-scripts, roots :/ (in fact I played with libreoffice math and qalculate, using first to improve qalculate output).

r1 - first supposed to change linearly as well from outer radius to some value for (width == radius), which I indeed got just by taste :) . However, this factor seemed to make less impact with high (radius/width) ratios, so I made it than slightly non linear... more exactly, with linear dependency on another factor.
I wanted to show something I did in lomath to explain it better, when need comes. Got system freeze (can't remember reason), requiring to reboot.

r2 - I made it keeping same ration to outer radius as int height to (outer height <minus> outer radius). But since its bigger than r1 at `radius/width==1`, I involved some coarse connection, implemented with `(r/2 < r2) ?` condition.

Btw I thought even about different drawing way:
- get fillabe stroke area
- subtract it from its origin path to get area inside border
- join them to get xshape mask
At least this way - no need to bother with internal radius, since there is just nothing to do with that :) .

tsipinakis:
> *8/4 - to make numbers less magic (I know both formulas are already like magic)))).

I still don't get what's the purpose, to increase accuracy? Then why not use floating points?

I couldn't find anything wrong functionality wise, so what's left here is:
a) For me to understand what's going on (I can't maintain the code if I don't understand how it came to be :) and b) some minor code style fixes that I'll post right after.

src/draw.c
```diff
@@ -464,26 +465,46 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
...
+                if (last)
+                        height -= settings.frame_width;
+                else
+                        height -= settings.separator_height;
+
+                /* Attempt to make internal radius more organic.
+                 * Simple r-w is not enough for too small r/w ratio.
+                 * simplifications: r/2 == r - w + w*w / (r * 2) with (w == r)
+                 **/
+                {       const int s = 2 << (sizeof(int) * 8 / 4); // Integer precision scaler
```
tsipinakis, line 492: [code style] There should be a newline after the brace
```
{
        // code here
}
```

src/x11/x.c
```diff
@@ -154,15 +135,34 @@ static void x_win_round_corners(struct window_x11 *win, const int rad)
...
+        XFlush(xctx.dpy);
+
         cairo_set_source_surface(win->c_ctx, srf, 0, 0);
         cairo_paint(win->c_ctx);
         cairo_show_page(win->c_ctx);
 
-        if (settings.corner_radius != 0)
+        sprintf(astr, "_NET_WM_CM_S%i", win->cur_screen);
+        cm_sel = XInternAtom(xctx.dpy, astr, true);
+        if (settings.corner_radius != 0 && XGetSelectionOwner(xctx.dpy, cm_sel) == None)
```
tsipinakis, line 153: This should probably be a new function `bool has_compositing()` or something similar, keep things simple

src/x11/x.c
```diff
@@ -641,12 +641,30 @@ struct window_x11 *x_win_create(void)
...
+        int depth;
+        Visual * vis;
+        XVisualInfo vi;
         XSetWindowAttributes wa;
 
-        root = RootWindow(xctx.dpy, DefaultScreen(xctx.dpy));
+        scr_n = DefaultScreen(xctx.dpy);
+        root = RootWindow(xctx.dpy, scr_n);
+        if (XMatchVisualInfo(xctx.dpy, scr_n, 32, TrueColor, &vi))
+        {
```
tsipinakis, line 653: [code style] Braces should be in the same line as the if/else statements

nick87720z:
I decided to go a bit further than has_compositing(). Added few more functions (from my previous changes) and renamed existing one.
```
x.c
renamed: x_win_corners_shape <- x_win_round_corners // Since all its job is to add xshape
new:
x_win_corners_unshape (win) // From else block, to not merge into x_win_corners_shape
bool x_win_composited (win),

draw.c (new): frame_internal_radius ()
```
Yet could be nice to reuse draw_rounded_rect() in xshaping, to not duplicate same func.

nick87720z:
What if move draw_rounded_rect() from degrees to direct M_PI / M_PI_2 use since they are exactly half/quater parts?

nick87720z:
Did not mean to push full ARGB support here, but noticed too late loss for one changed line, which went to style fix (gg, blindly selected entire block).

nick87720z:
Completed ARGB with proper frame stroke limit, so that it doesn't paint behind background.
Color in config may be specified now in '#aarrggbb', '#rrrrggggbbbb' and '#aaaarrrrggggbbbb' formats.

tsipinakis:
Thanks for also working on RGBA support! It's been a widely requested feature. Though can you split it into a new PR? It's easier to review and reference in the future when each PR has a smaller scope.

> Color in config may be specified now in '#aarrggbb', '#rrrrggggbbbb' and '#aaaarrrrggggbbbb' formats.

That's a bit excessive in my opinion, `#rrggbbaa`, and `#rrggbb` for backwards compatibility should be enough to cover all cases, is there any reason to support 4-5 different color formats?
(personally prefer RGBA, find it more readable and widely used)

nick87720z:
Hoped to avoid forced push).

About color coding - I tried to remember, which other apps I ever used need alpha first in config (configured by xresources). Rofi accepts as `#aarrggbb` (from where I took it). But urxvt accepts alpha last, though in specific format.

Yet dunst docs claim format support with single digit per channel `#rgb`, though no attempts were to differentiate them. Yet I expected first, that alpha-first will be useful to accept both argb and rgb int in single function, but it seems not - if alpha=0, it would mean opaque color without alpha.

nick87720z:
Hope PR will not suffer from forced push

tsipinakis:
@nick87720z Feel free to submit the RGBA patches as a new PR whenever you can - looking forward to having these merged as well

tsipinakis:
Found a bug here, in 16e38a9a84c46b0cd4fd7b0f2f88e9d8054ccc0b if the frame width is 0 the separator will not be added at all.
Fixed with 7735c9a4f60ef8cd645a39a98557c8950d0fa9ec.

PR #717, opened 2020-05-23
Rgba support

nick87720z:
Besides core rgba support I tried to implement support for arbitrary hex color string length (though limited by maximum unsigned length for given machine), as well as rgba() and rgb() formats, using scanf.

What's also interesting, adjacent non-overlaping areas are better to be combined in additive mode, so that fading out pixels at bounds mix correctly. I'm not sure though, if cairo uses alpha channel correctly (must be summed too, not like how it does with multiply op), but at least it removes artifacts between frame and background, visible when both are equally contrast with background.

Btw, imho colors are better to be parsed once on theme loading instead of on each draw().

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -59,37 +59,61 @@ static struct color hex_to_color(uintmax_t hexValue, int dpc)
         return ret;
 }
 
+static inline void color_clip(struct color * c)
+{
+        if (c->r > 1.0) c->r = 1.0;
+        if (c->g > 1.0) c->g = 1.0;
+        if (c->b > 1.0) c->b = 1.0;
+        if (c->a > 1.0) c->a = 1.0;
+}
+
 static struct color string_to_color(const char *str)
 {
-        uintmax_t val;
-        unsigned clen;
+        if (str[0] == '#') {
+                uintmax_t val;
+                unsigned clen;
+                {
+                        int cn;
+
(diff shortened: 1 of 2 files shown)


tsipinakis:
The [comment](https://github.com/dunst-project/dunst/pull/713#issuecomment-631981198) I posed in #713 still applies here. I don't see a reason to support so many different formats. #RRGGBBAA is widely used and and I don't think adding provides any sort of useful functionality. In fact, the opposite, it necessarily increases code complexity and adds more things that may break without anyone realising. 'Less is more' as they say.

The docs indeed _did_ claim #RGB support, but given that no-one has complained thus far that it doesn't work, my reaction would be to fix the docs rather than implement it.

nick87720z:
Hm, let assume FP format and hex string generic parsers are too complex.
But as for #RGB, it looks two times more readable than #RRGGBB. At least I started to use it where is possible :] .

I could undo back to `switch(len)` variant. It is still necessary to differ between #RRGGBBAA and #RRGGBB, unless you are going to break compatibility - with original code RRGGBB will be GGBBAA :) . IMHO, there is nothing complex in `hex_to_color()` taking digits-per-channel parameter.

nick87720z:
ping? At least one (my) vote for #RGB/#RGBA hex code support :)

PR #732, opened 2020-07-10
Add ignore_dbusclose settings

### LINKED ISSUE #731: Add a configuration to ignore dbus closeNotification event
In my use case, I'd like to prevent application to close notifications before the timeout configured by dunst.

I propose to add a `ignore_dbusclose` settings that would make dunst ignore such messages.

PR following.

matclab:
closes #731

However I was not able to test it because of X BadAtom error:
```
./dunst -print -verbosity debug
{
	appname: 'dunst'
	summary: 'startup'
	body: 'dunst is up and running'
	icon: 'dialog-information'
	raw_icon set: false
	icon_id: 'dialog-information'
	desktop_entry: ''
	category: 
	timeout: 10000
	urgency: LOW
	transient: 0
	formatted: '<b>startup</b>
dunst is up and running'
	fg: #888888
	bg: #222222
	frame: #aaaaaa
	fullscreen: pushback
	progress: -1
	stack_tag: 
	id: 2
	actions: {}
	script: (null)
}
DEBUG: RUN
DEBUG: XEvent: Ignoring '65'
DEBUG: XEvent: processing 'CreateNotify'
DEBUG: XEvent: Checking for active screen changes
DEBUG: XEvent: Checking for active screen changes
DEBUG: XEvent: Checking for active screen changes
DEBUG: XEvent: Checking for active screen changes
DEBUG: XEvent: Ignoring '65'
DEBUG: XEvent: Checking for active screen changes
DEBUG: XEvent: Checking for active screen changes
DEBUG: XEvent: Ignoring '19'
DEBUG: XEvent: Ignoring '19'
DEBUG: XEvent: Ignoring '15'
DEBUG: XEvent: processing 'Expose'
X Error of failed request:  BadAtom (invalid Atom parameter)
  Major opcode of failed request:  23 (X_GetSelectionOwner)
  Atom id in failed request:  0x0
  Serial number of failed request:  218
  Current serial number in output stream:  218
```

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -494,8 +494,16 @@ static void dbus_cb_CloseNotification(
 {
         guint32 id;
         g_variant_get(parameters, "(u)", &id);
-        if (!settings.ignore_dbusclose) 
+        if (settings.ignore_dbusclose) {
+                // Stay commpliant by lying to the sender,  telling him we closed the notification 
+                if (id > 0) {
+                        struct notification *n = queues_get_by_id(id);
+                        if (n) 
+                                signal_notification_closed(n, REASON_SIG);
+                }
+        } else {
                 queues_notification_close_id(id, REASON_SIG);
+        }
         wake_up();
         g_dbus_method_invocation_return_value(invocation, NULL);
         g_dbus_connection_flush(connection, NULL, NULL, NULL);
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -525,8 +525,6 @@ static void dbus_cb_GetServerInformation(
 void signal_notification_closed(struct notification *n, enum reason reason)
 {
         if (!n->dbus_valid) {
-                LOG_W("Closing notification '%s' not supported. "
-                      "Notification already closed.", n->summary);
                 return;
         }
 
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -494,7 +494,8 @@ static void dbus_cb_CloseNotification(
 {
         guint32 id;
         g_variant_get(parameters, "(u)", &id);
-        queues_notification_close_id(id, REASON_SIG);
+        if (!settings.ignore_dbusclose) 
+                queues_notification_close_id(id, REASON_SIG);
         wake_up();
         g_dbus_method_invocation_return_value(invocation, NULL);
         g_dbus_connection_flush(connection, NULL, NULL, NULL);
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -495,6 +495,7 @@ static void dbus_cb_CloseNotification(
         guint32 id;
         g_variant_get(parameters, "(u)", &id);
         if (settings.ignore_dbusclose) {
+                LOG_D("Ignoring CloseNotification message");
                 // Stay commpliant by lying to the sender,  telling him we closed the notification 
                 if (id > 0) {
                         struct notification *n = queues_get_by_id(id);
--- a/src/queues.c
+++ b/src/queues.c
@@ -506,6 +506,27 @@ gint64 queues_get_next_datachange(gint64 time)
         return sleep != G_MAXINT64 ? sleep : -1;
 }
 
+
+
+
+/* see queues.h */
+struct notification* queues_get_by_id(int id)
+{
+        assert(id > 0);
+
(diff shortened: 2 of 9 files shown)


tsipinakis:
Some points: The notification spec has no concept of disallowing closing notifications, so to make this 'compliant' I'd choose to lie to the clients that the notification is actually closed, but keep it open.
To do this, we need to:
 - [ ] Send a `NotificationClosed` signal back (see `signal_notification_closed`)
 - [ ] Invalidate all client-based actions as they most likely won't work anymore (see `notification_invalidate_action`)

Additionally please update the example `dunstrc` and the man page with some documentation about the new option.

tsipinakis:
Oh, I almost forgot the elephant in the room: 
> However I was not able to test it because of X BadAtom error:

Can you please expand more into this? Does it happen on `master` without this PR? If so, that's an important bug, please file an issue about it!

matclab:
> Oh, I almost forgot the elephant in the room:
> 
> > However I was not able to test it because of X BadAtom error:
> 
> Can you please expand more into this? Does it happen on `master` without this PR? If so, that's an important bug, please file an issue about it!

I just tried with master and cannot reproduce the problem, and switching back on this branch I cannot reproduce it anymore. And then I remember that I had an external screen connected when the problem occurred. Checking it again with an external screen and the problem happens also with master. I'll open a bug report.

matclab:
Note that I didn't call `notification_invalidate_action` which is already call by `signal_notification_closed`.

matclab:
The CI fails on codecov timeout. Nothing I can do on my side.

src/dbus.c
```diff
@@ -494,8 +494,16 @@ static void dbus_cb_CloseNotification(
 {
         guint32 id;
         g_variant_get(parameters, "(u)", &id);
-        if (!settings.ignore_dbusclose) 
+        if (settings.ignore_dbusclose) {
+                // Stay commpliant by lying to the sender,  telling him we closed the notification 
+                struct notification *n = dbus_message_to_notification(sender, parameters);
```
tsipinakis, line 499: That's not the right call here. `dbus_message_to_notification` constructs a new notification based on the contents of the message (used for the initial `Notify` call). So what this does is create a new notification and immediately close it.

There isn't a way to retrieve a notification via id currently so it looks like you're going to have to add it to `queues.c`, a simple linear search through the active queues (`waiting`, `displayed`) should be enough.

matclab: Oh yes… I've jumped on the first matching function without having grok the overall scheme.
Thank you for your insight !

I've reworked the patchset to search the notification in the queues

Note that I'm not so proud of my English and as such, the documentation may also welcome another eye.

tsipinakis:
The changes look good. One thing left is that when a notification is closed, a warning is printed `WARNING: Closing notification 'Test' not supported. Notification already closed.` from `signal_notification_closed`. I guess the easiest way to solve this is to remove that line entirely.

As for the docs, `Useful to enforce _the_ timeout set by dunst configuration`, just missing a `the` but otherwise it looks good to me.

matclab:
commits updated to edit the english according to review and remove the warning.

src/queues.c
```diff
@@ -506,6 +506,27 @@ gint64 queues_get_next_datachange(gint64 time)
         return sleep != G_MAXINT64 ? sleep : -1;
 }
 
+
+
+
+/* see queues.h */
+struct notification* queues_get_received_by_id(int id)
```
tsipinakis, line 513: What does `received` mean in this context?
I'd say `queues_get_by_id` would be a shorter and more descriptive name.

matclab: Hmmm I was trying to say that `id` was not looked for in all the queues… 
`queues_get_by_id`  would be a bit misleading as `history` queues is not used.

The best would probably to name the function `queues_get_by_id`   and search in all queues. In our use case it won't change the looking time as we expect the `id` to be in one of the first two queues.
Would this be ok ?

tsipinakis: That's fine by me. Default history max is ~20 items so performance impact is negligible.

tsipinakis:
Thanks for the PR! And sorry for the delay in merging this.

matclab:
No problem. Thanks for your help.

PR #735, opened 2020-07-12
multiple scripts for notification

Djeeberjr:
I have noticed that when multiple rules match only one script gets called. So i changed `script` to an array and store all the scripts in there where the rule matches.

I don't usually work with c so i am not sure if the use of `g_renew` is correct here.

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -44,8 +44,15 @@ void rule_apply(struct rule *r, struct notification *n)
         }
         if (r->format)
                 n->format = r->format;
-        if (r->script)
-                n->script = r->script;
+        if (r->script){
+                n->scripts = g_renew(const char*,n->scripts,n->script_count + 1);
+                n->scripts[n->script_count] = r->script;
+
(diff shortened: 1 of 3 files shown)


tsipinakis:
Ouch, this one fell through the cracks of the notification system - sorry! I'll review it tomorrow.

src/rules.c
```diff
@@ -44,8 +44,15 @@ void rule_apply(struct rule *r, struct notification *n)
...
         if (r->format)
                 n->format = r->format;
-        if (r->script)
-                n->script = r->script;
+        if (r->script){
+                n->scripts = g_renew(const char*,n->scripts,n->script_count + 1);
+                n->scripts[n->script_count] = r->script;
+
+                n->script_run = g_renew(bool,n->script_run,n->script_count + 1);
+                n->script_run[n->script_count] = false;
```
tsipinakis, line 52: There's no need to have this as an array, scripts are all run at the same time either way, so if one has ran, they all have.

PR #750, opened 2020-08-10
Fix possible division by zero in radius calculation.

### LINKED ISSUE #746: [i3-gaps] Transparency issue with rounded corners?
### Description

When RGBA colors are defined for backgrounds and rounded corners are enabled, rounded corners appear, though the base color (without transparency, i.e. RGB) fills the frame at the corner and makes the frame square.

### Installation info

- Version `1.5.0 (2020-07-23)`
- Install type `manually`
- Distro and version `Ubuntu 20.04 LTS`
- Further information: I am using `i3-gaps` _without_ a desktop environment.

### Workaround

If you are using picom (version `vgit-44b49`, [this fork](https://github.com/ibhagwan/picom)), you can enable `corner-radius = X;` to "fix" this.

MaximeWack:
Check all possible causes of division by zero in the calculation of the radius.

This fixes #746, and correctly checks for division by zero.

Checking `s == 0` too might be a bit overkill though.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -412,7 +412,7 @@ static int layout_get_height(struct colored_layout *cl)
  */
 static int frame_internal_radius (int r, int w, int h)
 {
-        if (r == 0 || w == 0 || h == 0)
+        if (r == 0 || h + (w - r) * 2 == 0)
                 return 0;
 
         // Integer precision scaler, using 1/4 of int size


src/draw.c
```diff
@@ -412,12 +412,12 @@ static int layout_get_height(struct colored_layout *cl)
  */
 static int frame_internal_radius (int r, int w, int h)
 {
-        if (r == 0 || w == 0 || h == 0)
-                return 0;
-
         // Integer precision scaler, using 1/4 of int size
         const int s = 2 << (8 * sizeof(int) / 4);
 
+        if (r == 0 || r == w || s == 0 || h == - (w - r) * 2)
```
tsipinakis, line 418:
```suggestion
        if (r == 0 || h + (w - r) * 2 == 0)
```

Some points:
* `s == 0` is indeed overkill, it's a constant after all.
* What's the point of `r == w`? The final check already checks for a division by 0 in that equation.

And a small readability nitpick, compare everything with 0 to be clear what it's actually checking.

MaximeWack: You are absolutely right!  
I first thought of `r == w` then I completed with invalid values for h without remembering to get rid of the first test.

tsipinakis: @MaximeWack ping? No pressure if you're busy, but I'm waiting for you to update the PR in order to merge this.

MaximeWack: Oops, sorry, forgot about this ! It's updated now.

Thank you !

tsipinakis:
Thanks for the fix!

Issue #307, opened 2017-03-04
Get the history stack from command line


ohque:
Let the users get notification history with something like `dunstcli --get-history` or `dunstcli ls` and then leave it to them to do with the data whatever they want.
One use case that I have in mind specifically is displaying notification count in the panel, as well as some other info based on what kind of notifications are in the history stack at the moment.

This feature request belongs to the general idea of exposing functionality to the command line: #284

11111000000:
Please, help! How to get history? I lost important notification!

tsipinakis:
@11111000000 If more than 20 notifications have been sent since it's not possible unfortunately. The [history_length](https://github.com/dunst-project/dunst/blob/master/dunstrc#L176-L177) setting controls how many notifications are kept in history, and it's set to 20 by default.

History is never saved anywhere, it's just in memory while dunst is running because it's not meant to handle important data it's there as a "I need a way to see what I missed".

snide:
Just a note because I see this thread pops on google when you search for dunst history. You can run `dunstctl history-pop` to pull the last notification. It can be run multiple times to go back in history.

Narrat:
If that is the case, isn't this issue a candidat for closing?

snide:
I didn't make the original issue and it makes a slightly different request, looking for the data, rather than replaying the notification. I just added it in case people run into a similar need.

nikp123:
Managed to implement it however it has REALLY ugly output. Especially because of unbroken newlines messing with the output. Posting code soon...

fwsmit:
I'm pleased to say that the history command has been added. This issue can be closed :)

gchamon:
For later reference, to print all the messages and their timestamps in readable format in bash:

```bash
dunstctl history |
  jq -cr '.data | flatten(1) | .[] | {"html_text": .message.data, "timestamp": .timestamp.data}' |
  while read -r notification_data; do
    html_text=$(jq -r '.html_text' <<<$notification_data)
    timestamp=$(jq -r '.timestamp' <<<$notification_data)
    rendered=$(html2text <<<$html_text)
    if [[ -n $rendered ]]; then
      # Get the system boot time in seconds + your timestamp converted to seconds
      converted_timestamp=$(bc <<<"$(stat -c %Y /proc/1) + ($timestamp / 1000000)")
      rendered_timestamp=$(date +"%Y-%m-%dT%H:%M:%S%z" -d "@$converted_timestamp")
[...]

PR #756, opened 2020-08-22
Add support for keeping notification history in a file (#307)

aesophor:
## Motivation
In #307, @ohque stated that:
> Let the users get notification history, ..., and then leave it to them to do with the data whatever they want.
>
> One use case that I have in mind specifically is displaying notification count in the panel, as well as some other info based on what kind of notifications are in the history stack at the moment.

## Overview
This PR adds support for keeping the notifications in a "history" file, which keeps the notification history even if dunst has been restarted. This file can be deleted at anytime, and when the next notification is received the file will start repopulating again.

The entire history will be stored as a json array which can be easily loaded via Python's json module (or any other language's json parsing library), leaving these history data to the users to do whatever they want. An example with Python3:
```
>>> import json
>>> with open('/home/aesophor/.cache/dunst/history', 'r') as f:
...     j = json.loads(f.read())
...
>>> len(j)
2
>>> j[1]
{'appname': 'dunstify', 'summary': 'Call waiting', 'body': '', 'icon': 'dialog-information', 'raw_icon set': False, 'icon_id': 'dialog-information', 'desktop_entry': '', 'category': '', 'timeout': 10000, 'urgency': 'NORMAL', 'transient': 0, 'formatted': '<b>Call waiting </b>', 'fg': '#F5F5F5', 'bg': '#292d3f', 'frame': '#41485f', 'fullscreen': 'show', 'progress': -1, 'stack_tag': '', 'id': 2, 'actions': {'yes': 'ACCEPT', 'no': 'DECLINE'}, 'script_count': 1, 'scripts': ['/home/aesophor/.local/bin/_play_notification_sound.sh']}
>>>
```

## New Cmdline Option
`-hist/-history_file`: Path to the notification history file
```
$ mkdir -p ~/.cache/dunst
$ dunst -hist ~/.cache/dunst/history
```

## New Config Option
Users may also set this option in the config's `[global]` section,
just make sure the target directory exists and it must be writable. For example,
```
[global]
    ...
    history_file = ~/.cache/dunst/history
```

### DIFF
--- a/config.h
+++ b/config.h
@@ -110,6 +110,8 @@ struct settings defaults = {
 
 .mouse_right_click = (enum mouse_action []){MOUSE_CLOSE_ALL, -1},
 
+.history_file = NULL,
+
 };
 
 struct rule default_rules[] = {
--- a/src/notification.c
+++ b/src/notification.c
@@ -4,12 +4,15 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <glib.h>
 #include <libgen.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
 #include <sys/wait.h>
 #include <unistd.h>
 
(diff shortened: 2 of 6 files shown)


tsipinakis:
I have several objections to merging something like this as opposed to the solution mentioned in #307. I'm pretty sure a similar log-to-file functionality has been proposed before and rejected for the same reasons.

First of all, this can easily turn into a privacy issue as many people would forget they have that set, and that it's logging all notifications to disk. Secondly, the size of the log file is totally unbounded so disk and cpu usage for every program parsing that file will keep increasing unnecessarily, and finally, it causes unnecessary filesystem writes for something that can easily be done in-memory.

Now that we have `dunstctl` merged, I think this can easily be added as an extension to that: `dunstctl history` which would return the current history as a json object.

aesophor:
@tsipinakis 

Thank you very much for your feedback! I think you're right, I'll close this PR for now.

tsipinakis:
For the record, one can easily re-create this behaviour by via the `script` option and a script that simply logs all parameters to a file.

PR #758, opened 2020-09-04
Fix a typo in configuration option in dunst.pod

keer4n:


tsipinakis:
Thanks for the PR, apologies for the delay, the notification fell through.

PR #771, opened 2020-10-29
Remove deprecated shortcuts section

iFreilicht:
The [shortcuts section is deprecated](https://dunst-project.org/documentation/#Shortcut-section-DEPRECATED-SEE-DUNSTCTL) in favor of dunstctl. Having it in the example dunstrc is confusing to new users, at least it was to me.

### DIFF
--- a/dunstrc
+++ b/dunstrc
@@ -267,6 +267,33 @@
     # where there are multiple screens with very different dpi values.
     per_monitor_dpi = false
 
+# The internal keyboard shortcut support in dunst is now considered deprecated
+# and should be replaced by dunstctl calls. You can use the configuration of your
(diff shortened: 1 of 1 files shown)


bebehei:
On the one hand side, I'd agree with you. But this example also served as documentation of the config file. A comment stating to use dunstctl, help all.

iFreilicht:
Alright, I changed the way it's implemented. Section is still there, the [deprecation notice from v1.5.0](https://github.com/dunst-project/dunst/blob/52d67616f1dcd9d4201de3f8096cbc2c09dbf1dd/RELEASE_NOTES) was copied in as a comment, and every option has a comment describing its equivalent dunstctl command.

Issue #772, opened 2020-10-31
Progress bar in notification


fwSmit:
I recently started using Dunst, and I like the customizabilty it offers. I'm missing one feature however: a progress bar. This is related to #318 from a while ago. Not much has been done about this, so I decided I want to implement something similar. 

The way I would do it is a bit different from the issue above. I would use the progress value %p and use it to draw a bar below the text of the notification. The text representation of %p would still be the default, so you would have to activate it via the config (or a command line argument?).

![Image of what it would look like](https://i.ibb.co/hXhccdb/notification-edited.png)
_This is what it would look like_

Is this be a feature that would be considered to be merged into dunst?

tsipinakis:
Since no-one has picked this in so long, any form of it is welcome. In some ways I like your design better than the 'progress bar background' proposed in #318.

> The text representation of %p would still be the default, so you would have to activate it via the config (or a command line argument?).

I'd be fine with making this the default as well, it's a clear improvement from the current situation where, in the default format, progress isn't even visible.

tsipinakis:
Implemented in #775.

PR #775, opened 2020-11-03
Add progress bar

fwSmit:
This PR addresses #772. It add a progress bar at the bottom of the notification when a value is passed via a dbus hint. Settings have been added to turn off the progress bar.  The old behaviour of showing a percentage in text can still be used the same way and even in conjunction with the progress bar.
A highlight colour is added and used for drawing the progress bar.

- [x] Bug: Icons larger than max_icon_size don't get drawn
- [x] Add more configuration variables
- [x] Make sure the border is pixel-aligned to avoid anti-aliassing
- [x] Let the progress bar define the width of the notification as well
- [x] Add tests
- [x] Documentation

_This is what it currently looks like_
notification-dunst-V2
notification-dunst-V2 2

Example usage
``` bash
notify-send -h int:value:10 "Brightness" 
```

### DIFF
--- a/docs/dunst.pod
+++ b/docs/dunst.pod
@@ -142,6 +142,32 @@ the notification on the left border of the screen while a horizontal offset of
 
 =back
 
+=item B<progress_bar> (values: [true/false], default: true)
+
+When an integer value is passed to dunst as a hint (see B<NOTIFY-SEND>), a
+progress bar will be drawn at the bottom of the notification. This
+behavior can be turned off by setting this setting to false.
+
+=item B<progress_bar_height> (default: 10)
+
+The height of the progress bar in pixel. This includes the frame. Make sure
+this value is bigger than twice the frame width.
+
+=item B<progress_bar_min_width> (default: 150)
+
+The minimum width of the progress bar in pixels. The notification is rescaled
+to fit the bar.
+
+=item B<progress_bar_max_width> (default: 300)
+
+The maximum width of the progress bar in pixels. The notification is resized
+to fit the progress bar.
+
+=item B<progress_bar_frame_width> (default: 1)
+
+The frame width of the progress bar in pixels. This value should be smaller
+than half of the progress bar height.
+
 =item B<indicate_hidden> (values: [true/false], default: true)
 
 If this is set to true, a notification indicating how many notifications are
@@ -561,8 +587,8 @@ Specifies the keyboard shortcut that opens the context menu.
 
 The urgency sections work in a similar way to rules and can be used to specify
 attributes for the different urgency levels of notifications (low, normal,
-critical). Currently only the background, foreground, timeout, frame_color and
-icon attributes can be modified.
+critical). Currently only the background, foreground, hightlight, timeout,
+frame_color and icon attributes can be modified.
 
 The urgency sections are urgency_low, urgency_normal, urgency_critical for low,
 normal and critical urgency respectively.
@@ -592,6 +618,12 @@ See COLORS for the value format.
 =item B<-lb/nb/cb color>
 
 Defines the background color for low, normal and critical notifications respectively.
+
+See COLORS for the value format.
+
+=item B<-lh/nh/ch color>
+
+Defines the highlight color for low, normal and critical notifications respectively.
 
 See COLORS for the value format.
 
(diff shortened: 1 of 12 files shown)


src/draw.c
```diff
@@ -618,6 +644,38 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width)
                 cairo_fill(c);
         }
 
+        // progress bar positioning
+        if (have_progress_bar(cl->n)){
+                int progress = cl->n->progress;
```
tsipinakis, line 649:
```suggestion
                int progress = MIN(cl->n->progress, 100);
```
To catch any misuse of the progress value.

test/functional-tests/dunstrc.progress_bar
```diff
@@ -0,0 +1,10 @@
+[global]
+    font = Monospace 8
+    allow_markup = yes
+    format = "<b>%s</b>\n%b"
+    geometry = "0x5-30+20"
+	icon_position = left
```
tsipinakis, line 6: There's some inconsistent indentation here, tabs instead of space.

test/functional-tests/test.sh
```diff
@@ -177,12 +177,35 @@ function geometry {
     keypress
 }
 
+function progress_bar {
+	killall dunst
```
tsipinakis, line 181: Same in this file, some calls use tabs for indentation rather than spaces.

tsipinakis (commented):
Some code style nitpicks and you also forgot to add the new settings to the example `dunstrc`.

This PR was a breeze to review and couldn't even find any breaking bugs. Huge kudos! :)

src/draw.c
```diff
@@ -354,7 +366,12 @@ static struct colored_layout *layout_from_notification(cairo_t *c, struct notifi
 
         pango_layout_get_pixel_size(cl->l, NULL, &(n->displayed_height));
         if (cl->icon) n->displayed_height = MAX(cairo_image_surface_get_height(cl->icon), n->displayed_height);
-        n->displayed_height = MAX(settings.notification_height, n->displayed_height + settings.padding * 2);
+        n->displayed_height = n->displayed_height + settings.padding * 2;
+
+        // progress bar
+        if (have_progress_bar(n)) n->displayed_height += settings.progress_bar_height + settings.padding;
```
tsipinakis, line 372: Please also put a newline after if statements.

docs/dunst.pod
```diff
@@ -142,6 +142,32 @@ the notification on the left border of the screen while a horizontal offset of
 
 =back
 
+=item B<no_progress_bar> (values: [true/false], default: false)
```
tsipinakis, line 145: Oh, I'd also like to avoid having negative no_* settings, I'd prefer if this is progress_bar=[true/false], defaulting to true.

src/draw.c
fwSmit: I fixed it in notification.c to prevent code duplication

test/functional-tests/dunstrc.progress_bar
fwSmit: Got it. I set my vimrc to tabexpand to prevent this in the future.

docs/dunst.pod
fwSmit: Yeah I like that better as well. The reason I did this was because it's not possible to change default true variables in command line. I don't think that will be used much in this case so I'll change it.

fwSmit:
Thank you for the review. I have addressed everything. The tests are failing, because progress values greater than 100 are being changed to 100. Is there a reason why it's better not to change them to 100?

tsipinakis:
> Is there a reason why it's better not to change them to 100?

There can be cases where the progress can be >100%, think for example volume levels with pulseaudio that can go to 150+. So it's best to handle it here so `%p` is still accurate.

fwSmit:
Good point. I have applied your suggestion

src/settings.c
```diff
@@ -355,6 +355,49 @@ void load_settings(char *cmdline_config_path)
...
+
+        settings.progress_bar_frame_width = option_get_int(
+                "global",
+                "progress_bar_frame_width", "-progress_bar_frame_width", defaults.progress_bar_frame_width,
+                "Frame width of the progress bar"
+        );
+
+        settings.progress_bar = option_get_bool(
+                "global",
+                "progress_bar", "-progress_bar", false,
```
tsipinakis, line 384: You can set this to `true` as well, the goal is to have this and `dunstrc` in sync at some point not make it diverge more :)

tsipinakis (approved):

src/settings.c
fwSmit: Oh I forgot that one. I changed it now

tsipinakis:
Merged! Thanks for your work.

PR #776, opened 2020-11-06
Turn icons on by default

fwSmit:
In my opinion this is a saner default.

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -416,7 +416,7 @@ void load_settings(char *cmdline_config_path)
         {
                 char *c = option_get_string(
                         "global",
-                        "icon_position", "-icon_position", "off",
+                        "icon_position", "-icon_position", "left",
                         "Align icons left/right/off"
                 );
 


tsipinakis:
Agreed, you should also change the value in the `dunstrc` to align with this.

fwSmit:
It seems like that is already the case in the current dunstrc

fwSmit:
There are some other cases where the default dunstrc doesn't align with the default without dunstrc (or if that variable isn't defined in the config). I think that's worth fixing as well, because when you're making your own dunst config you want to have sane defaults for the values you haven't defined yet.
The main one I'm thinking about is where the width of the notification is the entire width of the screen when it isn't defined.

I think the problem here is that defaults are defined in two places: the default dunstrc and in the code. It might be worth it to fall back to the default dunstrc when a value isn't found in the config file. That way you can get rid of the default values in code entirely (but what to do when the default dunstrc isn't found?).

tsipinakis:
> It seems like that is already the case in the current dunstrc

You're right, I was looking at an outdated copy. My bad

> It might be worth it to fall back to the default dunstrc when a value isn't found in the config file

That's already done if you're using the Debian version of dunst: The default dunstrc is placed in /etc/xdg/dunst which is one of the fallbacks for config detection. In my view the values in the example config are the sane default i.e. the 'real' default values while the hardcoded values are the values that are most convenient for existing users, so those that don't have that value in their dunstrc yet.


I'd love to sync these two but to some limited extent that would break backward compatibility with old configs and I'd rather avoid it. In other words: Someday!

tsipinakis:
>  The main one I'm thinking about is where the width of the notification is the entire width of the screen when it isn't defined.

Settings like this I'm fine with updating like I said, I guess a first step would be to make a list of what is out of sync and figuring out which ones to update.

PR #779, opened 2020-11-15
Create abstract output and fix wayland bug

fwSmit:
This commit adds an output struct which abstracts the X11 specific
functions and makes it possible to easily create a drop-in wayland
output.

It also fixes a bug in wayland where notifications won't disappear.
This is because wayland doesn't give access to user input when a
client is not in focus. This way it seems like the user is always
idle. The idle functionality is now disabled in Wayland until proper
support is added.

Dunst should now work fine on wayland using X11 compatibility.
Only the idle functionality won't work, obviously.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -1,7 +1,6 @@
 #include "draw.h"
 
 #include <assert.h>
-#include <cairo.h>
 #include <math.h>
 #include <pango/pango-attributes.h>
 #include <pango/pangocairo.h>
@@ -10,6 +9,7 @@
 #include <pango/pango-types.h>
 #include <stdlib.h>
 #include <inttypes.h>
+#include <glib.h>
 
 #include "dunst.h"
 #include "icon.h"
@@ -17,7 +17,15 @@
 #include "markup.h"
 #include "notification.h"
 #include "queues.h"
-#include "x11/x.h"
+#include "output.h"
+#include "settings.h"
+
+struct color {
+        double r;
+        double g;
+        double b;
+        double a;
+};
 
 struct colored_layout {
         PangoLayout *l;
@@ -30,7 +38,8 @@ struct colored_layout {
         const struct notification *n;
 };
 
-struct window_x11 *win;
+const struct output *output;
+window win;
 
 PangoFontDescription *pango_fdesc;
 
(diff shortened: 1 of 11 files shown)


src/output.c
```diff
@@ -0,0 +1,63 @@
...
+        /* wl_is_idle, */
+        /* wl_have_fullscreen_window */
+/* }; */
+
+const struct output* output_create(void)
+{
+        if (is_running_wayland()){
+                LOG_I("System is running wayland");
+        }
+        else{
```
tsipinakis, line 58:
```suggestion
        } else {
```

src/output.c
```diff
@@ -0,0 +1,63 @@
...
+
+        /* wl_get_active_screen, */
+
+        /* wl_is_idle, */
+        /* wl_have_fullscreen_window */
+/* }; */
+
+const struct output* output_create(void)
+{
+        if (is_running_wayland()){
```
tsipinakis, line 55:
```suggestion
        if (is_running_wayland()) {
```

tsipinakis (commented):
Apologies for the delay, looks good from a quick code read aside from some code style nitpicks.
I'll do the final review and merge tomorrow. Glad to see someone pick this up!

fwSmit:
I fixed the code style issues.

src/draw.c
```diff
@@ -250,10 +262,10 @@ static struct dimensions calculate_dimensions(GSList *layouts)
 
 static PangoLayout *layout_create(cairo_t *c)
 {
-        struct screen_info *screen = get_active_screen();
+        const struct screen_info *screen = output->get_active_screen();
 
         PangoContext *context = pango_cairo_create_context(c);
-        pango_cairo_context_set_resolution(context, screen_dpi_get(screen));
+        pango_cairo_context_set_resolution(context, screen->dpi);
```
tsipinakis, line 268: Where is `screen->dpi` initialized? Did you forget to add this to the patch?

fwSmit: Yep, I forgot to diff screen.c. Apparently it doesn't break much when you forget to set it. I added it now.

tsipinakis (approved):
LGTM!

PR #780, opened 2020-11-17
Fix some tests

fwSmit:
When testing, I noticed some warnings, so I fixed them.
Also, the no markup configuration had markup in the config.I deleted that, because that made it harder to check if markup was added.


tsipinakis (approved):
Thanks!

PR #781, opened 2020-11-22
Add wayland support

fwSmit:
Don't merge this in master yet, it's probably best to put this on a separate branch first.

Thanks to a lot of code from the old wayland branch and from mako, I improved dunst's wayland support. 
It now handles mouse input well. This is done in a hacky way by making the sleep in the main loop shorter. I noticed that otherwise the input was only handled when the sleep was over. I'm not sure what the right way to do this is.
The second improvement is the positioning of the notification. Instead of putting the notification at the edge of the screen, it's positioned using zwlr_layer_surface_v1_set_anchor and using some margin. This will probably not work on non-wayland compositors, but I haven't tested that yet.
And thirdly, the Makefile now builds, so that's good. I also made sure to autogenerate the protocol headers with wayland-scanner, so they always stay up-to-date.

These are the things that have to be done, before this can be considered to be merged in master:
- [x] Select X11 output when necesary
- [x] Get rid of the hacky timeout
- [x] Fix code style issues and organise the commits
- [x] Fix visual artifact right after clicking away a notification (when multiple notifications are active).
- [x] Make sure it build when only typing 'make' and not 'make all'
- [x] Fix CI not building
- [x] Fix notification not disappearing after going fullscreen
- [ ] Fix bug where input region isn't correctly updated after closing notification (only after mouse moves). 
- [x] Fix notification not moving after it's resized.
- [x] Resolve merge conflicts


progandy:
I think the proper way to fix the wayland mouse interactions would be to add a wayland source to the mainloop and let it handle all wayland events asynchronously. libgwater includes an implementation if you do not want to write your own.

https://github.com/sardemff7/libgwater/blob/master/wayland/libgwater-wayland.c

fwSmit:
Thanks, that looks like a good solution! I'm not sure how the maintainers feel about adding this as a dependency. It won't be too hard to implement with this as a reference either.

fwSmit:
> I think the proper way to fix the wayland mouse interactions would be to add a wayland source to the mainloop (just like `x_win_reg_source`) and let it handle all wayland events asynchronously. libgwater includes an implementation if you do not want to write your own.
> 
> https://github.com/sardemff7/libgwater/blob/master/wayland/libgwater-wayland.c
I implemented it. Works well.

tsipinakis:
Looking good!

> Don't merge this in master yet, it's probably best to put this on a separate branch first.

Since you're going with the 'one big PR in one go' strategy feel free to work on your branch until you feel you've reached a sufficient milestone to merge this into master, I'll do a full review then (perhaps consider marking the PR as draft until then).

> I'm not sure how the maintainers feel about adding this as a dependency. 

It looks like that library isn't packaged in Debian and I'd assume other distros so for that alone I'd hesitate adding it as a dependency, given that it's only a ~200 line file, I'm ok with embedding it.

fwSmit:
> Since you're going with the 'one big PR in one go' strategy feel free to work on your branch until you feel you've reached a sufficient milestone to merge this into master, I'll do a full review then (perhaps consider marking the PR as draft until then).

I've converted it to a draft.

> It looks like that library isn't packaged in Debian and I'd assume other distros so for that alone I'd hesitate adding it as a dependency, given that it's only a ~200 line file, I'm ok with embedding it.

In the linked repository they recommend to use it as a git submodule, but I think it's easier to embed it as well.

fwSmit:
This PR is now ready for review. The wayland output has almost all of the features of the X11 output. One notable exception is that the wayland version cannot detect fullscreen, thus cannot show only some notifications in fullscreen.
You can however let notifications get on top of fullscreen applications by selecting the layer overlay in the settings.

Here is a summary of the changes in other places:
- I extracted the input handling from x.c and put it in its own file. 
- I added the wayland dependencies to the makefile. You should autogenerate the wayland protocol files with wayland-scanner, but I didn't make that the default for compatibility. Thus the generated files had to be included in the repo (see src/wayland/protocols)
- Embedded libgwater in the repo

fwSmit:
> Bug two is that somewhere along the way Xwayland has become buggy. Notifications don't disappear when clicking them away. The window will instead flicker. It's probably worth to fix this one, so I'll look at it soon.

I bisected it, and the bug was introduced by ed1e3a815a9f80bdd1591faaa8eaa25a70d6b9d4 which applies a fix from master. I'm not sure why it works on wayland and X11, but not on Xwayland.

tsipinakis:
> I bisected it, and the bug was introduced by ed1e3a8 which applies a fix from master. I'm not sure why it works on wayland and X11, but not on Xwayland.

Try out https://github.com/tsipinakis/dunst/tree/action-fix, a better fix for the same bug. Is the issue still present there?

fwSmit:
> Try out https://github.com/tsipinakis/dunst/tree/action-fix, a better fix for the same bug. Is the issue still present there?

I tried it out on wayland, and I didn't see any issues. I'll rebase onto that branch when it is in master.

tsipinakis:
I've pushed the updated fix to master, please rebase when possible.

fwSmit:
> I've pushed the updated fix to master, please rebase when possible.

Thanks, I've rebased ( or rather merged) onto master. This doesn't fix the issue on xwayland right away (flickering when clicking on notification and notifications not disapearing).
I remembered that I forgot idle skipping on xwayland (as there is no way to detect if the user is idle) thinking it would fix the second issue. It fixed both issues however, so that's great. Do you have any reason how wrong idle detection make notifications not disappear by clicking on them?

tsipinakis:
> Do you have any reason how wrong idle detection make notifications not disappear by clicking on them?

Hm, the only way I could see that happening is if a89287cb3e76375bf636764caa2ebec39ecf7d45 wasn't applied, but it looks like it's already merged in your branch. Weird.

tsipinakis:
I'll set up a Wayland VM now and do the review within the next few days. This comment is just to force myself to actually get it done without any more delays ;)

docs/dunst.pod
```diff
@@ -264,7 +268,24 @@ Set to 0 to disable.
...
+
+One of bottom, top or overlay.
+
+Place dunst notifications on the selected layer. Using overlay
+will cause notifications to be displayed above fullscreen windows, though
+this may also occur at top depending on your compositor.
+
+The bottom layer is below all windows and above the background.
+
+Default: top
```
tsipinakis, line 283: Why not have overlay as the default layer? Wouldn't that be more appropriate for dunsts use, and more consistent with the X11 behavior?

I guess this also needs a note that notifications not shown during fullscreen will not be redisplayed like the current `pushback` behavior is.

src/queues.c
```diff
@@ -143,9 +143,12 @@ static bool queues_notification_is_finished(struct notification *n, struct dunst
 
         bool is_idle = status.fullscreen ? false : status.idle;
 
+        /* There is no way to detect if the user is idle
+         * on xwayland, so assume they aren't */
+        if (is_running_xwayland()) is_idle = false;
```
tsipinakis, line 148: Since this is only used to disable idle support, I guess it'd be much cleaner/easier to modify the settings struct to disable it on output creation (if wayland is detected).

src/wayland/wl.c
```diff
@@ -0,0 +1,743 @@
...
+        struct org_kde_kwin_idle_timeout *idle_timeout;
+        bool configured;
+        bool dirty;
+        bool is_idle;
+        bool has_idle_monitor;
+
+	struct {
+		struct wl_pointer *wl_pointer;
+		int32_t x, y;
+	} pointer;
```
tsipinakis, line 68: [nitpick] Indentation error here.

src/wayland/wl.c
```diff
@@ -0,0 +1,743 @@
...
+                return false;
+        }
+        if (ctx.layer_shell == NULL) {
+                LOG_E("compositor doesn't support zwlr_layer_shell_v1");
+                return false;
+        }
+        if (ctx.seat == NULL) {
+                LOG_W("no seat was found, so dunst cannot see input");
+        }
+        else {
```
tsipinakis, line 393:
```suggestion
        } else {
```

src/wayland/wl.c
```diff
@@ -0,0 +1,743 @@
...
+                wl_list_for_each(output, &ctx.outputs, link) {
+                        get_xdg_output(output);
+                }
+                wl_display_roundtrip(ctx.display);
+        }
+        //if (ctx.xdg_output_manager == NULL &&
+        //                strcmp(ctx.config.output, "") != 0) {
+        //        LOG_E("warning: configured an output but compositor doesn't "
+        //                "support xdg-output-unstable-v1 version 2");
+        //}
```
tsipinakis, line 418: Why is this commented out?

tsipinakis (commented):
I did a quick skim of the code and tested it in wayland. I did have some issues with the last notification not disappearing sometimes (missing wake up?) but it's pretty hard to reproduce.

tsipinakis:
Oh, I forgot to add that if it's possible it'd be nice to have a compile-time flag to toggle the wayland/x11 outputs to minimize the dependencies i.e. X11 only users could have an option to not compile with all the wayland libraries.

docs/dunst.pod
fwSmit: Yeah that's probably better. I'll change it

src/wayland/wl.c
fwSmit: Else if on the same line as else as well? (see a few lines below)

src/wayland/wl.c
fwSmit: It didn't compile, and I don't know how to fix that check correctly.

Looking at it again, it is a warning for when you have configured a preferred monitor name and you compositor doesn't support the protocol.
I don't think that it is relevant to us, because we use output numbers. I'll test with a second monitor soon, to see if `monitor` works without xdg-output-unstable-v1.

src/queues.c
fwSmit: Btw, it's only xwayland. Idle management on Wayland works atm (if your compositor supports it).

fwSmit: I agree that's simpler (until we need to know if we're running xwayland elsewhere). I changed it.

fwSmit:
> Oh, I forgot to add that if it's possible it'd be nice to have a compile-time flag to toggle the wayland/x11 outputs to minimize the dependencies i.e. X11 only users could have an option to not compile with all the wayland libraries.
> 
> Edit: But it's in no way a blocker, don't feel obligated to add this if it proves too difficult a problem.

Yeah that's certainly possible. I'll add it

src/wayland/wl.c
fwSmit: As I thought, we don't need xdg-output. This warning can be safely deleted. Actually all the xdg output code can be removed. If we want to have an option to select monitor names to dunst, we will need to add it again. Is it worth to remove all xdg-output code then?

src/wayland/wl.c
tsipinakis: I missed that, good catch. I generally try to follow the [kernel code style](https://www.kernel.org/doc/html/v4.10/process/coding-style.html) with a few deviations (indentation being one of them, as I don't want to break `git blame` with a huge commit updating it).

fwSmit:
> > Oh, I forgot to add that if it's possible it'd be nice to have a compile-time flag to toggle the wayland/x11 outputs to minimize the dependencies i.e. X11 only users could have an option to not compile with all the wayland libraries.
> > Edit: But it's in no way a blocker, don't feel obligated to add this if it proves too difficult a problem.
> 
> Yeah that's certainly possible. I'll add it

This is done now

src/wayland/wl.c
fwSmit: Thanks. that may be helpful in HACKING.md or some contributing doc as well.

src/wayland/wl.c
tsipinakis: Since we use numbers to define monitors in the config currently, I think we can safely remove it. Less is more as they say.

src/settings.c
```diff
@@ -224,7 +224,7 @@ void load_settings(char *cmdline_config_path)
         }
 #endif
 
-        if (settings.force_xwayland) {
+        if (settings.force_xwayland && is_running_wayland()) {
                 /* There is no way to detect if the user is idle
                  * on xwayland, so turn this feature off */
                 settings.idle_threshold = 0;
```
tsipinakis, line 230: A `LOG_W` here informing the user of the change would be good as well.

tsipinakis:
When done with all the comments above, please rebase on master and drop the merge commits (I think it's worth it to keep a clean history for future bisects).

src/wayland/wl.c
fwSmit: I removed all the code regarding xdg-output. That was quite a lot of code for just getting the window name...

fwSmit:
> When done with all the comments above, please rebase on master and drop the merge commits (I think it's worth it to keep a clean history for future bisects).

Could you explain how to do that. Is it literally `git rebase master`? Other than that everything is done.

tsipinakis:
> Could you explain how to do that

Basically `git rebase -i origin/master`, replace `pick` with `drop` to all merge commits and resolve conflicts until the rebase is done. (For good measure get a backup of the branch, it's easy to mess up rebases :) ).

fwSmit:
> > Could you explain how to do that
> 
> Basically `git rebase -i origin/master`, replace `pick` with `drop` to all merge commits and resolve conflicts until the rebase is done. (For good measure get a backup of the branch, it's easy to mess up rebases :) ).

The merge commits I see on github are not visible in the rebase, but it still worked! Thanks for the help! I also updated input.c to be up to date with the current master.

Narrat:
Just butting in, as I noticed something: Maybe 505a763 should additionally be edited or merged with the next commit (via interactive rebase)? As this commits breaks the source because of adding notes from git to the sourcecode (git's notation for a merge conflict).
But if this is of no concern, or planned to do another rebase after the review is done, than this comment can be disregarded.

fwSmit:
> Just butting in, as I noticed something: Maybe [505a763](https://github.com/dunst-project/dunst/commit/505a763cb6e1c62e315aa62345c389f661bd7825) should additionally be edited or merged with the next commit (via interactive rebase)? As this commits breaks the source because of adding notes from git to the sourcecode (git's notation for a merge conflict).
> But if this is of no concern, or planned to do another rebase after the review is done, than this comment can be disregarded.

Thank you, I've fixed it. That's probably better for bisects. Here's another 26 new sha hashes for you :)

fwSmit:
I've opened an issue in sway for the one remaining bug. I'll link it here for clarity https://github.com/swaywm/sway/issues/5911.

ci-build-test
```diff
@@ -0,0 +1 @@
+1
```
tsipinakis, line 1: Was this to force a CI rebuild? If so, it can be removed.

docs/dunst.pod
```diff
@@ -264,7 +268,27 @@ Set to 0 to disable.
...
+this may also occur at top depending on your compositor.
+
+In Wayland, Notifications won't be delayed when in fullscreen (like when
+setting B<fullscreen> to pushback in X11). This is a Wayland limitation.
+
+The bottom layer is below all windows and above the background.
+
+Default: overlay
+
+=item B<use_xwayland> (values: [true/false], default: false) (Wayland only)
```
tsipinakis, line 288: In the code this is `force_xwayland`, better update it here.

docs/dunst.pod
```diff
@@ -789,6 +828,10 @@ Or pushback which is equivalent to delay with the difference that already
 existing notifications are paused and hidden until the focus to the fullscreen
 window is lost.
 
+See B<layer> to change fullscreen behavior in Wayland
+
+Default: show
```
tsipinakis, line 833: Default layer was changed to `overlay`

dunstrc
```diff
@@ -247,6 +247,16 @@
...
 
+    ### Wayland ###
+    # These settings are Wayland-specific. They have no effect when using X11
+
+    # Uncomment this if you want to let notications appear under fullscreen
+    # applications (default: overlay)
+    # layer = top
+
+    # Set this to true to use X11 output on Wayland.
+    use_xwayland = false
```
tsipinakis, line 258: Same here, should be `force_wayland`

tsipinakis (commented):
The `DISABLE_WAYLAND` flag doesn't seem to compile without the wayland libs installed, it fails in `settings.h` as we're using a wayland enum in there.

I don't see a clean way to fix that sadly so we might have to abandon that.

ci-build-test
fwSmit: Yes, though I removed it already.

docs/dunst.pod
fwSmit: This section is about the fullscreen setting, which is x11 only. There is a reference to make users aware of the similar setting for wayland.

fwSmit:
> The `DISABLE_WAYLAND` flag doesn't seem to compile without the wayland libs installed, it fails in `settings.h` as we're using a wayland enum in there.
> 
> I don't see a clean way to fix that sadly so we might have to abandon that.

I didn't document it correctly. You need to use `make WAYLAND=0` to compile without wayland. The flag DISABLE_WAYLAND is then set my make, along with some other things. I had one wayland include in a non-wayland file. That one wasn't needed, so i got rid of it.
I also made sure you can build dust with support for wayland, without needing wayland-protocols installed (because the header files are already genereated in src/wayland/protocls).
Could you test if you can build without wayland installed?

fwSmit:
Added the build info to the README as well

docs/dunst.pod
tsipinakis: Whoops, I misread that.

tsipinakis:
> I didn't document it correctly. You need to use make WAYLAND=0 to compile without wayland.

That's what I meant, I mentioned `DISABLE_WAYLAND` specifically because its not failing in the dependency resolution but in compilation. See the attached build log.

[build.log](https://github.com/dunst-project/dunst/files/5750161/build.log)

fwSmit:
Just so you know, I'll take a few days off. I can finish this off in the beginning of next year.
Happy new year already :)

fwSmit:
> > I didn't document it correctly. You need to use make WAYLAND=0 to compile without wayland.
> 
> That's what I meant, I mentioned `DISABLE_WAYLAND` specifically because its not failing in the dependency resolution but in compilation. See the attached build log.
> 
> [build.log](https://github.com/dunst-project/dunst/files/5750166/build.log)

I think I've fixed it now, but I had to include one enum from the wayland code to make it work. It has become more of a mess than I initially anticipated, but it should work now.

src/wayland/libgwater-wayland.c
```diff
@@ -0,0 +1,192 @@
+/*
```
bebehei, line 1: As far as I can tell from the Makefile, we generate this code. Is there an apparent reason to check it in into the repo?

fwSmit: No it's taken from libgwater, see https://github.com/dunst-project/dunst/pull/781#issuecomment-731775356.
The files in `src/wayland/protocols/` are generated, but since not all distros have a recent version fo `wayland-scanner` (which generates these files) or `wayland-protocols`. That's why these files are checked in the repo and are not generated on compile time (yet).

src/queues.c
```diff
@@ -311,6 +310,21 @@ void queues_notification_close_id(int id, enum reason reason)
 {
         struct notification *target = NULL;
 
+        char* reason_string;
```
bebehei, line 313: This code is suited better somewhere else (depending on its exact purpose).

When we're here:

- We are saying we're closing a notification. (Which notification!?)
- We are not sure, whether there is even a notification, we're able to close. (`target` may be `NULL`)
- We are not sure, if we will even send a DBus signal `NotificationClosed`. (`target.dbus_valid` may be `false`)


Please put it either into `if (target)` or put it to `dbus.c` with the `signal_notification_closed` handling.

bebehei (dismissed):
I've used this branch now on my main desktop. Previously I dunst master with xwayland.

It works, but I'm not yet convinced, this is everything to properly support wayland. Still feels a bit clumsy during daily usage.

Makefile
```diff
@@ -33,6 +33,13 @@ $(error "Failed to query $(PKG_CONFIG) for package 'systemd'!")
 endif
 endif
 
+ifneq (0,${WAYLAND})
+DATA_DIR_WAYLAND_PROTOCOLS ?= $(shell $(PKG_CONFIG) wayland-protocols --variable=pkgdatadir)
```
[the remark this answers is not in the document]
fwSmit: Thanks, I didn't know make worked that way

config.mk
```diff
@@ -20,14 +20,23 @@ VALGRIND ?= valgrind
...
 # uncomment to disable parsing of dunstrc
 # or use "CFLAGS=-DSTATIC_CONFIG make" to build
 #STATIC= -DSTATIC_CONFIG # Warning: This is deprecated behavior
 
 # flags
 DEFAULT_CPPFLAGS = -D_DEFAULT_SOURCE -DVERSION=\"${VERSION}\"
-DEFAULT_CFLAGS   = -g --std=gnu99 -pedantic -Wall -Wno-overlength-strings -Os ${STATIC}
-DEFAULT_LDFLAGS  = -lm
+DEFAULT_CFLAGS   = -g --std=gnu99 -pedantic -Wall -Wno-overlength-strings -Os ${STATIC} ${DISABLE_WAYLAND}
+DEFAULT_LDFLAGS  = -lm -lrt
```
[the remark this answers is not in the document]
fwSmit: I doesn't compile without. Wayland seems to use it https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/101.

docs/dunst.pod
```diff
@@ -82,6 +82,10 @@ starts at 0. See the B<follow> setting.
 Defines where the notifications should be placed in a multi-monitor setup. All
 values except I<none> override the B<monitor> setting.
 
+On Wayland there is no difference between mouse and keyboard focus. When either
```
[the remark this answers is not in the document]
fwSmit: I did that already :). Are there any settings that should be marked both X11 and Wayland?

fwSmit:
> I've used this branch now on my main desktop. Previously I dunst master with xwayland.
> 
> It works, but I'm not yet convinced, this is everything to properly support wayland. Still feels a bit clumsy during daily usage.

Could you elaborate on that? I have compared the wayland and Xwayland versions and the wayland version for me seems way better. Try replacing a long notification with a short one. I believe on Xwayland, you will resize artifacts, while on wayland it's perfect.
I'm curious what you have noticed that's worse on wayland compared to xwayland

src/queues.c
fwSmit: This was just a quick hack for debugging purposes. I think the log might be useful later, so I'll move it to signal_notification_closed.

docs/dunst.pod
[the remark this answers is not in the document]
bebehei: Oh, well sorry. This comment was a leftover. I've later seen that you already did this but forgot to remove the comment.

Makefile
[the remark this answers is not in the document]
bebehei: Well, if properly used, this feature might be quite nice. `make` doesn't save the value of the variable, but its references. It later evaluates the variables when actually needed.


```
SOURCEVAR=b
VAR=a ${SOURCEVAR}
SOURCEVAR=c
all:
	echo ${VAR}
```

It's a weird feature of make.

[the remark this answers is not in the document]
bebehei: Thanks for patching.

config.mk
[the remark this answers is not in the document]
bebehei: Would you mind packing this also into a variable depending on `${WAYLAND}`?

bebehei:
> Could you elaborate on that?

Well, I used dunst in combination with xwayland for about a month. It wasn't a nice experience.

I had some pain points on click actions, close

fwSmit:
> > Could you elaborate on that?
> 
> Well, I used dunst in combination with xwayland for about a month. It wasn't a nice experience, but it worked some way.
> 
> Now using it with native wayland, there didn't stick out any big improvements. I still feel pain points one thing, that the window opens on the wrong output. It doesn't open anymore on the same monitor, but it's not well aligned.
> 
> But please don't count on me here. I'm a sway user for one month now. I do not understand all concepts of wayland yet and maybe these settings have to be done in sway somewhere.

So if I understand correctly, the dunst window opens on the wrong output and is positioned incorrectly on that output? Looking at your dotfiles it seems like you have
```
geometry = "300x5-30-30"
follow = keyboard
```

Note that on wayland the bottom is somehow counted from the top of your status bar (if your status bar is at the bottom). So if you want the notification to appear at the same place as on X11, you might have to subtract the status bar height from that. Now that I think of it, it's probably good to document that.

The follow part I don't understand why it's not working. You will have to give more info about that.

fwSmit:
Another thing to note, I recently discovered there is a wayland protocol for detecting if there is a fullscreen window. I will hopefully implement that protocol for dunst soon, but that will mean that the `layer` setting will be almost useless. Maybe it's wise to move that setting to the experimental section, making it easy to immediately deprecated it.

src/output.c
```diff
@@ -3,9 +3,12 @@
 #include "log.h"
 #include "x11/x.h"
 #include "x11/screen.h"
-/* #include "wayland/wl.h" */ // Not yet
 
-const bool is_running_wayland(void){
+#ifndef DISABLE_WAYLAND
```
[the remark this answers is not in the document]
fwSmit: Got it

fwSmit:
@tsipinakis Do you have time to look at this PR?

src/dunst.c
```diff
@@ -78,6 +78,7 @@ static gboolean run(void *data)
                 draw();
                 output->win_show(win);
         } else {
+                LOG_I("Calling win_hide");
```
tsipinakis, line 81: This is redundant (and the information level is too high for this log)

tsipinakis (approved):
> Well, I used dunst in combination with xwayland for about a month. It wasn't a nice experience, but it worked some way.

Now using it with native wayland, there didn't stick out any big improvements. I still feel pain points one thing, that the window opens on the wrong output. It doesn't open anymore on the same monitor, but it's not well aligned.

Native support is always better than going through the translation layer. Especially since it looks like most distros are moving away from X11.
Now, some pain points might still be there but they can be fixed with later iterations.

---

Sorry for the delay in this, I've been a bit busier than I expected lately. This is ready for a merge. Just a tiny log nitpick I noticed when scanning the code for a final pass ;)

src/dunst.c
fwSmit: I removed it

fwSmit:
> Sorry for the delay in this, I've been a bit busier than I expected lately. This is ready for a merge. Just a tiny log nitpick I noticed when scanning the code for a final pass ;)

No problem. Thanks for reviewing the code. It should be good to go now.

tsipinakis:
Thanks again @fwSmit, that's an insane amount of work you did here! Merged!

PR #791, opened 2020-12-13
Contrib: add progress notify script

fwsmit:
Add progress notify script for audio and brightness notifications


tsipinakis (approved):
Great addition, thanks!

Now, I should probably set something like this up for my system :)

fwSmit:
Yeah I made this for myself as well, but I figured I could write some documentation for it and make a PR.

Issue #792, opened 2020-12-20
Number of unread messages


ritze:
Currently it's not possible to get the number of unread messages when dunst is paused or messages were not closed actively.

The number of unread messages could be used to be shown in a status bar or somewhere else, like other notification services already do.

The primary scope would be when dunst is paused, because you need to focus on a task. To not miss unread messages, you can easily check, if any message were queued since you paused dunst.

tsipinakis:
That's a good idea, however 'unread' needs to be defined more precisely.
Dunst has 3 queues `waiting`, `displayed` and `history`.
`waiting` being a notification that has arrived but hasn't been shown yet (either on standby because of no space, or dunst is paused)
`displayed` being the notifications currently being shown.

In your opinion should `unread` be `waiting+displayed` or just the number in the waiting queue?

ritze:
That's a good point. I let this detail open, since I'm not familiar how dunst manages the notifications.

Why not just adding options for all three? E.g. `dunstctl history-num`, `dunstctl waiting-num`, and `dunstctl missed-num` (for displayed notifications which were not actively "clocked" away). Or just printing all three by calling one command:

```
$ dunstctl status
              Waiting: 0
  Currently displayed: 3
              History: 13
```

This would be the same behavior like for other systemd commands (e.g. `networkctl status`). The disadvantage with one command to show all stats would be that you have to write additional logic around your script, which you use to forward this information.

tsipinakis:
The formatting can be done directly in `dunstctl` so adding all of the above would not be that big a problem.

If anyone wants to work on this there's already the `queues_length_{waiting,displayed,history}` functions and all that's needed is to add the interfaces to [dbus.c](https://github.com/dunst-project/dunst/blob/master/src/dbus.c#L82)

fwSmit:
#793 is merged, so I'll close this issue

PR #793, opened 2020-12-21
Get the number of notifications

ritze:
This will add the dbus properties
* `displayed`
* `waiting`
* `history`

and extends `dunstctl` with the sub-command `status`, which provides this information.

Currently I'm not sure, if `history` is the right term. Maybe `shown` fits here better.

See also the feature request #792.

### DIFF
--- a/dunstctl
+++ b/dunstctl
@@ -14,17 +14,18 @@ show_help() {
 	cat <<-EOH
 	Usage: dunstctl <command> [parameters]"
 	Commands:
-	  action                         Perform the default action, or open the
-	                                 context menu of the notification at the
-	                                 given position
-	  close                          Close the last notification
-	  close-all                      Close the all notifications
-	  context                        Open context menu
-	  history-pop                    Pop one notification from history
-	  is-paused                      Check if dunst is running or paused
-	  set-paused [true|false|toggle] Set the pause status
-	  debug                          Print debugging information
-	  help                           Show this help
+	  action                             Perform the default action, or open the
+	                                     context menu of the notification at the
+	                                     given position
+	  close                              Close the last notification
+	  close-all                          Close the all notifications
+	  context                            Open context menu
+	  history-pop                        Pop one notification from history
+	  is-paused                          Check if dunst is running or paused
+	  set-paused [true|false|toggle]     Set the pause status
+	  status [displayed|history|waiting] Show the number of notifications
+	  debug                              Print debugging information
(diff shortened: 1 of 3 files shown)


dunstctl
```diff
@@ -83,6 +84,17 @@ case "${1:-}" in
 			property_set paused variant:boolean:"$2"
 		fi
 		;;
+	"status")
```
tsipinakis, line 87: `status` is a bit ambiguous, one could assume this is paused status. I'd suggest `count` as the command (unless you have a better idea?)

src/dbus.c
```diff
@@ -85,6 +85,10 @@ static const char *introspection_xml =
     "            <annotation name=\"org.freedesktop.DBus.Property.EmitsChangedSignal\" value=\"true\"/>"
     "        </property>"
 
+    "        <property name=\"displayed\" type=\"u\" access=\"read\" />"
+    "        <property name=\"history\" type=\"u\" access=\"read\" />"
+    "        <property name=\"waiting\" type=\"u\" access=\"read\" />"
```
tsipinakis, line 90: Similar here, I'd name it `displayedLength` to avoid naming conflicts for future expansions.

tsipinakis (commented):
Some naming nitpicks. Otherwise LGTM :+1:

src/dbus.c
ritze: I would name them all in the same way:
* `displayedLength`
* `historyLength`
* `waitingLength`

Without the technical knowledge, it's hard to get, what those properties shall return. Maybe we can find names?

dunstctl
ritze: I have no better idea as `count`. I'll change it…

src/dbus.c
tsipinakis: I gave an example for displayed but I did mean to change all of them :). Your list of names looks good to me.

ritze:
Shall I provide unit tests too? As I can see there are no tests for the dbus interface or `dunstctl` yet.

fwSmit:
Yeah that'd be great. More testing is always better

tsipinakis (approved):
LGTM, as @fwSmit said, if you can also add some tests for this it'd be great but I'm happy to merge as-is if you don't have the time.

ritze:
I guess that it makes sense to merge it now. I won't find time to write tests for the next days or even weeks.

PR #798, opened 2020-12-24
Change config location to /etc/dunst/dunstrc

### LINKED ISSUE #795: Move default config to /etc/xdg/dunst/dunstrc
Currently the dunst config is installed in `/usr/share/dunst/dunstrc`. This is not used by default by dunst, which results in users having to copy the dunstrc to `~/.config/dunst/dunstrc` before being able to use dunst properly.
Dunst doesn't search in `/usr/share/dunst/dunstrc` for config files. It only searches `/etc/xdg/dunst/dunstrc` and `/etc/xdg/dunstrc` for system wide config files. It would be easier for new users if the default config is used if it isn't overridden by a user config file.

Another thing that's worth considering is to let the user config not override the system wide config, but add to it. So if you have a setting that's not defined in the user config, the system wide setting is used. This way we can get rid of the default definitions in `config.h`. Sometimes it's forgotten to add a default setting to both config.h and the default dunstrc.

Edit: see #198 for an old reference to something similar

fwSmit:
Fixes #795 
The makefile still isn't fully xdg compliant. If the user has set `$XDG_CONFIG_DIRS` it should be reflected our install, I believe.
I didn't know the best way to implement that, but [this](https://www.gnu.org/software/make/manual/make.html#Directory-Variables) is some make documentation about all the standard directories, and [this](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html) is the spec.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,7 +204,7 @@ install-dunstctl: dunstctl
 	install -Dm755 dunstctl ${DESTDIR}${BINDIR}/dunstctl
 
 install-doc:
-	install -Dm644 dunstrc ${DESTDIR}${DATADIR}/dunst/dunstrc
+	install -Dm644 dunstrc ${DESTDIR}${SYSCONFDIR}/dunst/dunstrc
 
 install-service: install-service-dbus
 install-service-dbus: service-dbus
@@ -223,7 +223,7 @@ uninstall: uninstall-service uninstall-dunstctl
 	rm -f ${DESTDIR}${BINDIR}/dunstify
 	rm -f ${DESTDIR}${MANPREFIX}/man1/dunst.1
 	rm -f ${DESTDIR}${MANPREFIX}/man1/dunstctl.1
-	rm -rf ${DESTDIR}${DATADIR}/dunst
+	rm -rf ${DESTDIR}${SYSCONFDIR}/dunst
 
 uninstall-dunstctl:
 	rm -f ${DESTDIR}${BINDIR}/dunstctl
--- a/config.mk
+++ b/config.mk
@@ -1,6 +1,7 @@
 # paths
 PREFIX ?= /usr/local
 BINDIR ?= ${PREFIX}/bin
+SYSCONFDIR ?= /etc
 DATADIR ?= ${PREFIX}/share
 # around for backwards compatibility
 MANPREFIX ?= ${DATADIR}/man
--- a/docs/dunst.pod
+++ b/docs/dunst.pod
@@ -35,7 +35,8 @@ rules or using the output in other scripts.
 
 =head1 CONFIGURATION
 
-An example configuration file is included (usually /usr/share/dunst/dunstrc).
+An example configuration file is included (usually /etc/dunst/dunstrc). Note:
+this was previously /usr/share/dunst/dunstrc.
 To change the configuration, copy this file to ~/.config/dunst/dunstrc and edit
 it accordingly.
 
--- a/src/settings.c
+++ b/src/settings.c
@@ -54,6 +54,10 @@ static FILE *xdg_config(const char *filename)
                 g_free(path);
         }
 
+        if (!f) {
+                f = fopen("/etc/dunst/dunstrc", "r");
+        }
+
         return f;
 }
 
--- a/test/test-install.sh
+++ b/test/test-install.sh
@@ -1,27 +1,31 @@
 #!/usr/bin/env bash
 
+# Throw error any time a command fails
 set -euo pipefail
 
 BASE="$(dirname "$(dirname "$(readlink -f "$0")")")"
-PREFIX="${BASE}/install"
+DESTDIR="${BASE}/install"
+PREFIX="/testprefix"
+SYSCONFDIR="/sysconfdir"
 
-make -C "${BASE}" SYSTEMD=1 SERVICEDIR_SYSTEMD="${PREFIX}/systemd" SERVICEDIR_DBUS="${PREFIX}/dbus" PREFIX="${PREFIX}" install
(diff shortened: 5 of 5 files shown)


fwSmit:
Weird, this commit doesn't seem to cause the ci failing. My bisect points to 58d215ddfeaf3fae9f017d05b8c886d7f4aadffd.

tsipinakis:
> that's only on wayland and is not a problem. I can't find what causes the ci failing

The CI is failing in the `test-install` script. See [here](https://github.com/dunst-project/dunst/runs/1605771381#step:8:16).

Re: This PR. I'm thinking in Debian terms where it's usual to have docs and example config files in `/usr/share/doc/<pkg>` and it's now muscle memory to search there when I want something. Looking in `/etc/xdg` is counter-intuitive in that sense. So, I'd opt for having both, one in `/usr` for documentation and the `/etc/xdg` one for actually loading it.

fwSmit:
> > that's only on wayland and is not a problem. I can't find what causes the ci failing
> 
> The CI is failing in the `test-install` script. See [here](https://github.com/dunst-project/dunst/runs/1605771381#step:8:16).

Ah thanks. To fix it I need to make sure the config is installed somewhere else when the prefix is changed. Do I need to make a new CONFIG_PREFIX or change PREFIX to `/` ?


> Re: This PR. I'm thinking in Debian terms where it's usual to have docs and example config files in `/usr/share/doc/<pkg>` and it's now muscle memory to search there when I want something. Looking in `/etc/xdg` is counter-intuitive in that sense. So, I'd opt for having both, one in `/usr` for documentation and the `/etc/xdg` one for actually loading it.

I don't think it's a good idea to place two identical files in different places. Linking them doesn't seem good either. Maybe this is too distro specific. On arch I'm used to the config being something like `/etc/dunst/dunstrc`. There are only a few configs in `/etc/xdg/...`. I'm interested what other people think is the best place for the config.
We can also just make dunst search `/usr/share/doc/dunst/dunstrc`, but I think that's also a bad solution.

tsipinakis:
Linking or making dunst look in /usr/share/doc isn't going to work in Debian at least as everything in that directory is gzipped. Debian-wise I'd always opt to install the config file in both locations as the `/usr/share/doc` directory is muscle memory to all users by now, (and Debian policy suggests it).

The question here is what's best for upstream, default behaviour, does this extend to other distros?

> Do I need to make a new CONFIG_PREFIX or change PREFIX to / ?

`DESTDIR` should be the target 'root' so changing `PREFIX` would break convention here. I'm not sure what the standard variable for `/etc/` files is, if there is one. We should check what other projects do with such files.

fwSmit:
> Linking or making dunst look in /usr/share/doc isn't going to work in Debian at least as everything in that directory is gzipped. Debian-wise I'd always opt to install the config file in both locations as the `/usr/share/doc` directory is muscle memory to all users by now, (and Debian policy suggests it).

[This](https://www.debian.org/doc/manuals/maint-guide/dother.en.html#conffiles) page suggests that conffiles should be under /etc for debian. Every other distro I can find installs it's dotfiles in /etc as well.

> The question here is what's best for upstream, default behaviour, does this extend to other distros?

I think it might be best to install them to `/etc/dunst/dunstrc` and change dunst to look there as well.

> `DESTDIR` should be the target 'root' so changing `PREFIX` would break convention here. I'm not sure what the standard variable for `/etc/` files is, if there is one. We should check what other projects do with such files.

Oh thanks, that's what I was looking for. I changed the test-install script to use that instead of PREFIX. Also made some other improvements to it.

tsipinakis:
> This page suggests that conffiles should be under /etc for debian

I guess the divide here is that I view the dunstrc as both a configuration file (which is what this change is about) but also a big part of dunsts limited documentation/examples. Which is why I hesitate moving it entirely off that directory.

> I think it might be best to install them to /etc/dunst/dunstrc and change dunst to look there as well.

That might be a good compromise, my holdout is that `/etc/xdg` is a rarely used path (no other program on my system has any config there other than autostart files).

Also, don't forget to update the path in dunsts manpage.

fwSmit:
I finished this PR and rebased onto the `wayland-2` branch. There were no conflicts.

tsipinakis:
> I finished this PR and rebased onto the wayland-2 branch. There were no conflicts.

Apparently the merge commit confuses it. Both PRs have conflicts.

fwSmit:
> > I finished this PR and rebased onto the wayland-2 branch. There were no conflicts.
> 
> Apparently the merge commit confuses it. Both PRs have conflicts.

No that's because I added some stuff to wayland-2 because of bebehei's comments. I rebased again and fixed the conflicts now.
I noticed that the changes from #793 also show up in the "files changed" tab. I don't know why, I just rebased on upstream/master.

fwSmit:
Could this be reviewed? Then it can be merged together with #799. Otherwise we will stack up conflicts.

tsipinakis:
Sorry, took a look at this yesterday and forgot to comment. This needs to be rebased. Merge commit is d3dd26a but in mater it's d3dd26a. I feel like this will cause some issues (or at least confuse history) if I merge as-is.

fwSmit:
Yeah I thought so, but didn't know how to fix. I rebased again and it has been cleaned up

tsipinakis (approved):
LGTM.

While reviewing this I couldn't help but think that we should simplify the `test-install` script to only provide the `DESTDIR` variable and let everything else be the default. But that's out of scope of this.

fwSmit:
> While reviewing this I couldn't help but think that we should simplify the `test-install` script to only provide the `DESTDIR` variable and let everything else be the default. But that's out of scope of this.

I don't think that's better. Setting these variables to some non-default value makes sure you aren't hardcoding the defaults in the makefile.

I noticed btw that I didn't update commit message. It should be "Change config location to /etc/dunst/dunstrc". Well I guess it's too late to change it now.

PR #799, opened 2020-12-25
Split man pages

fwsmit:
The dunst man page was getting really big , so I figured it was best to split it up. This PR splits the dunst man page into a section 1 and section 5 man page. Section 1 contains all the documentation to running the dunst command. Section 5 contains the documentation regarding the configuration of dunst. I also updated and added to a few parts of the man page.
Don't merge before #781, because I will have to fix a few conflicts (should have based this on the wayland branch...).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,12 +123,14 @@ test/test: ${OBJ} ${TEST_OBJ}
 	${CC} -o ${@} ${TEST_OBJ} $(filter-out ${TEST_OBJ:test/%=src/%},${OBJ}) ${CFLAGS} ${LDFLAGS}
 
 .PHONY: doc doc-doxygen
-doc: docs/dunst.1 docs/dunstctl.1
+doc: docs/dunst.1 docs/dunst.5 docs/dunstctl.1
 
 # Can't dedup this as we need to explicitly provide the name and title text to
 # pod2man :(
-docs/dunst.1: docs/dunst.pod
+docs/dunst.1: docs/dunst.1.pod
 	${POD2MAN} --name=dunst -c "Dunst Reference" --section=1 --release=${VERSION} $< > $@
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -106,12 +106,14 @@ test/test: ${OBJ} ${TEST_OBJ}
 	${CC} -o ${@} ${TEST_OBJ} $(filter-out ${TEST_OBJ:test/%=src/%},${OBJ}) ${CFLAGS} ${LDFLAGS}
 
 .PHONY: doc doc-doxygen
-doc: docs/dunst.1 docs/dunstctl.1
```
bebehei, line 109: Since, you're adding another manpage, you have to add it to the install targets. It's currently missing in the target `install-dunst`.

bebehei (commented):
I absolutely appreciate it! Gives more clarity about the stuff.

Also ppl from sway and wayland side are used to have multiple manpages in different man sections. Especially section 5 is used there heavily.

Makefile
fwSmit: Added it. While I was at it, I added it to the uninstall and test-install.sh as well.

fwSmit:
> I absolutely appreciate it! Gives more clarity about the stuff.
> 
> Also ppl from sway and wayland side are used to have multiple manpages in different man sections. Especially section 5 is used there heavily.

Yep, that's where I got it from. It might take some getting used to, but in the end it's better.

tsipinakis (approved):

tsipinakis:
As I mentioned [here] this is blocked on #781

fwSmit:
Rebased this one on #798, so it should be good to go as well.

fwSmit:
I rebased again. It's ready for merging now.

tsipinakis (approved):

Issue #767, opened 2020-10-23
Tell scripts everything you know


jbash:
It would be nice for a script to get access to *everything* duntst knows about a notification. My immediate use for this would be to be able to play different sounds for notifications with different categories. Having to guess from the appname isn't ideal. Obviously there are many other possible uses.

Ideally this would include both all the notification's attributes as originally provided to dunst, *and* all of the attributes as subsequently modified by rules. I could imagine it also being useful for a script to know the results of various decisions dunst had made about whether and how to display the notification.

I would suggest passing this stuff via environment variables as described in #653. I don't think it completely obviates #639; the two could be useful together. Another possibility would be to provide all of these things as %-expansions in the script line, but that might be harder to use.

fwSmit:
Maybe it's better to give the script a notification ID and make sure all the neccesary variables are available through dunstctl.
This will probably be more work though.

fwSmit:
@jbash The PR linked above passes info about the notification via environment variables. The variables are listed in [the man page](https://github.com/dunst-project/dunst/pull/802/files#diff-c1889f7158a4cace6e494b1a7c64687b91444308ffe24007ca16e40436e0321e). It would be a bit harder to provide the variables as they were before modified by rules, so they aren't included. Are there any specific use cases you have that would benefit from more variables being provided via environment variables?

jbash:
I don't have any specific use case that would need the unrewritten variables. One can always just not rewrite them. Thanks.

fwSmit:
Okay cool, I'll close this issue then.

Issue #653, opened 2019-08-14
Pass variables as params


bebehei:
I think we should pass the flags in the notification script as environment variables.

https://github.com/dunst-project/dunst/blob/53e7f3d16a747a778b7892a175c224d4c558b44e/src/notification.c#L123-L130

Enhancing this, is a PITA and doesn't scale. We also would allow passing flags (#639) and adding new fields would be easy to document and remove later.

It may be also a fix for #645, as you could easily add the desktop_entry field as an environment variable and you could filter it in the script properly.

tsipinakis:
Why the `help wanted` tag?

bebehei:
idk.

fwSmit:
Implemented by #802

PR #802, opened 2021-01-07
Set environment variables for scripts

fwSmit:
Addresses #767  and  #653. This PR sets environment variables for scripts to use. Let me know if I have to add more or if the naming should be changed. I still need to do the documentation, but I wanted to get some comments before that.

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -133,6 +133,26 @@ void notification_run_script(struct notification *n)
                         if (pid2) {
                                 exit(0);
                         } else {
+                                // Set environment variables
+                                gchar *n_id_str = g_strdup_printf("%i", n->id);
+                                gchar *n_progress_str = g_strdup_printf("%i", n->progress);
+                                gchar *n_timeout_str = g_strdup_printf("%li", n->timeout/1000);
+                                gchar *n_timestamp_str = g_strdup_printf("%li", n->timestamp / 1000);
+                                char* icon_path = get_path_from_icon_name(icon);
+                                safe_setenv("DUNST_APP_NAME",  appname);
+                                safe_setenv("DUNST_SUMMARY",   summary);
+                                safe_setenv("DUNST_BODY",      body);
+                                safe_setenv("DUNST_ICON_PATH", icon_path);
+                                safe_setenv("DUNST_URGENCY",   urgency);
+                                safe_setenv("DUNST_ID",        n_id_str);
+                                safe_setenv("DUNST_PROGRESS",  n_progress_str);
+                                safe_setenv("DUNST_CATEGORY",  n->category);
+                                safe_setenv("DUNST_STACK_TAG", n->stack_tag);
+                                safe_setenv("DUNST_URLS",      n->urls);
+                                safe_setenv("DUNST_TIMEOUT",   n_timeout_str);
+                                safe_setenv("DUNST_TIMESTAMP", n_timestamp_str);
+                                safe_setenv("DUNST_STACK_TAG", n->stack_tag);
+
                                 int ret = execlp(script,
                                                 script,
                                                 appname,
@@ -142,7 +162,7 @@ void notification_run_script(struct notification *n)
                                                 urgency,
                                                 (char *)NULL);
                                 if (ret != 0) {
-                                        LOG_W("Unable to run script: %s", strerror(errno));
+                                        LOG_W("Unable to run script %s: %s", n->scripts[i], strerror(errno));
                                         exit(EXIT_FAILURE);
                                 }
                         }
--- a/src/notification.c
+++ b/src/notification.c
@@ -129,6 +129,7 @@ void notification_run_script(struct notification *n)
                         int status;
                         waitpid(pid1, &status, 0);
                 } else {
+                        // second fork to prevent zombie processes
                         int pid2 = fork();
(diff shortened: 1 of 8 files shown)


src/notification.c
```diff
@@ -133,6 +133,29 @@ void notification_run_script(struct notification *n)
...
+                                setenv("DUNST_STACK_TAG", n->stack_tag, true);
+                                setenv("DUNST_URLS",      n->urls,      true);
+                                setenv("DUNST_TIMEOUT",   n_timeout,    true);
+                                setenv("DUNST_TIMESTAMP", n_timestamp,  true);
+                                setenv("DUNST_STACK_TAG", n->stack_tag, true);
+
+                                g_free(n_id);
+                                g_free(n_progress);
+                                g_free(n_timeout);
+                                g_free(n_timestamp);
```
tsipinakis, line 158: Are the frees necessary here? We're exec'ing out so everything will be lost either way.

fwSmit: Yeah, you're right. I'll delete those frees. I'll delete the if after that as well, since that isn't needed as well.

fwSmit: Why is the logic so complicated anyways? You only have to fork once right?

tsipinakis: Double forking forces a reparent of the child process to init. This means dunst doesn't have to wait to reap the exit code of the script (since we don't care).

tsipinakis: > I'll delete the if after that as well, since that isn't needed as well

If you're talking about the `if (ret != 0) {` one, that's needed in case `exec` fails :)

fwSmit: >Double forking forces a reparent of the child process to init. This means dunst doesn't have to wait to reap the exit code of the script (since we don't care).

Ah that expains it. I'll but a comment there for clarity.

> If you're talking about the if (ret != 0) { one, that's needed in case exec fails :)

 Yeah, but the if statement itself is useless, since we only get there if exec fails.

fwSmit:
Turns out, setenv can block if you feed it NULL. So I made a safe setenv function to prevent that. Also added some documentation.

src/notification.c
```diff
@@ -129,22 +129,41 @@ void notification_run_script(struct notification *n)
...
+                                safe_setenv("DUNST_BODY",      body);
+                                safe_setenv("DUNST_ICON_PATH", icon);
+                                safe_setenv("DUNST_URGENCY",   urgency);
+                                safe_setenv("DUNST_ID",        n_id);
+                                safe_setenv("DUNST_PROGRESS",  n_progress);
+                                safe_setenv("DUNST_CATEGORY",  n->category);
+                                safe_setenv("DUNST_STACK_TAG", n->stack_tag);
+                                safe_setenv("DUNST_URLS",      n->urls);
+                                safe_setenv("DUNST_TIMEOUT",   n_timeout);
+                                safe_setenv("DUNST_TIMESTAMP", n_timestamp);
```
tsipinakis, line 153:
```suggestion
                                safe_setenv("DUNST_TIMEOUT",   n_timeout/1000);
                                safe_setenv("DUNST_TIMESTAMP", n_timestamp/1000);
```

Let's return the times in the same format as provided.

fwSmit: Those are actually strings. I'll go ahead and replace the ints.

src/notification.c
```diff
@@ -129,22 +129,41 @@ void notification_run_script(struct notification *n)
...
-                                int ret = execlp(script,
+                                // Set environment variables
+                                gchar *n_id = g_strdup_printf("%i", n->id);
+                                gchar *n_progress = g_strdup_printf("%i", n->progress);
+                                gchar *n_timeout = g_strdup_printf("%li", n->timeout);
+                                gchar *n_timestamp = g_strdup_printf("%li", n->timestamp);
+                                safe_setenv("DUNST_APP_NAME",  appname);
+                                safe_setenv("DUNST_SUMMARY",   summary);
+                                safe_setenv("DUNST_BODY",      body);
+                                safe_setenv("DUNST_ICON_PATH", icon);
```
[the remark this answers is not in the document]
fwSmit: Oh the man page said it was always an icon path. I'll submit a separate PR for doing just that.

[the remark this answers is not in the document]
fwSmit: I has not become a separate PR, but I extracted the icon path function.

fwSmit:
I added the icon path function to this pr as well. Tests fail, because the old  behaviour of the pixbuf creation was to continue searching the icon path when an invalid icon was found. This behaviour changed to returning the path to the invalid icon, thus not being able to create the pixbuf.
If you want I could add a check if the icon is valid to the path search, but I don't think it's very important.

tsipinakis:
CI appears to be failing now.

fwSmit:
> CI appears to be failing now.

See https://github.com/dunst-project/dunst/pull/802#issuecomment-767806601

tsipinakis:
> See #802 (comment)

Whoops, that should teach me to review code at 1 am :see_no_evil: 

That's fine, I don't expect that there will be many invalid icons and, but please fix the tests :)

fwSmit:
I forgive you :). The tests should now be fixed

tsipinakis (approved):
LGTM

Issue #762, opened 2020-09-17
Dunst doesnt work if screen number >= amount of monitors on that screen


mcz:
I have two X screens, :0.0 and :0.1 and I want to run dunst on the second one. When I try to run dunst, it exits with the message ``dunst: src/x11/screen.c:377: get_active_screen: Assertion `ret >= 0 && ret < screens_len' failed.``.
Running it on screen 0 works. The options "monitor" and "follow" don't change anything.

Some (maybe) useful info:
````
$ DISPLAY=:0 xrandr

Screen 0: minimum 320 x 200, current 1920 x 1080, maximum 16384 x 16384
eDP-1 connected primary 1920x1080+0+0 (normal left inverted right x axis y axis) 344mm x 194mm
   1920x1080     60.02*+  60.01    59.97    59.96    59.93
   1680x1050     59.95    59.88
   1400x1050     59.98
   1600x900      59.99    59.94    59.95    59.82
   (... more resolutions...)
   432x243       59.92    59.57
   320x240       60.05
   360x202       59.51    59.13
   320x180       59.84    59.32

$ DISPLAY=:0.1 xrandr

Screen 1: minimum 320 x 200, current 1920 x 1080, maximum 16384 x 16384
HDMI-1 connected primary 1920x1080+0+0 (normal left inverted right x axis y axis) 509mm x 286mm
   1920x1080     60.00*+
   1600x1200     60.00
   1680x1050     59.88
   1280x1024     75.02    60.02
   1440x900      59.90
   1280x800      59.91
   1024x768      75.03    60.00
   800x600       75.00    60.32
   640x480       75.00    59.94
   720x400       70.08
DP-1 disconnected (normal left inverted right x axis y axis)
HDMI-2 disconnected (normal left inverted right x axis y axis)
````
I think the problem is as follows:  
In `src/x11/screen.c:randr_update()` `XRRMonitorInfo` is used to get the number of Monitors on the `DefaultScreen(xctx.dpy)`. The default screen is taken from the DISPLAY environment variable, which is :0.1, so the default Screen is 1. I have one active monitor on that screen, so `screens_len` gets set to 1.
Later, in `src/x11/screen.c:screen_info()` `ret` gets set from `DefaultScreen()` once again to determine what monitor to use (This is wrong, since screens != monitors). And because my screen's index isn't less than the number of active monitors on that screen, assert aborts.

### Installation info

[...]

PR #809, opened 2021-01-22
Fix process of gettign the active monitor

mcz:
- Get the window with keyboard focus using XGetInputFocus() instead of
  _NET_ACTIVE_WINDOW
- Handle edge cases of keyboard/mouse focus being on different X screen
- Set sane default in case the active monitor can't be determined
  (fixing issue #762)

### DIFF
--- a/src/x11/screen.c
+++ b/src/x11/screen.c
@@ -309,10 +309,7 @@ const struct screen_info *get_active_screen(void)
         if (settings.f_mode == FOLLOW_NONE) {
                 if (settings.monitor >= 0 && settings.monitor < screens_len) {
                         ret = settings.monitor;
-                } else {
-                        ret = XDefaultScreen(xctx.dpy);
                 }
-
                 goto sc_cleanup;
         } else {
                 int x, y;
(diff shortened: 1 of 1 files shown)


mcz:
This doesn't fix the general confusion between screens and monitors. Xrandr and Xinerama allow for multiple monitor to show parts of the same screen. Most references to "screens" in screen.c should be talking about monitors instead. It would be possible to handle multiple screens, but I don't think it would be worth it since most people don't use that part of X11 nowadays.

src/x11/screen.c
```diff
@@ -327,19 +324,25 @@ const struct screen_info *get_active_screen(void)
...
+                                 * the screen the pointer is on, force following
+                                 * the mouse.
                                  */
                                 force_follow_mouse = true;
                         } else {
                                 Window child_return;
-                                XTranslateCoordinates(xctx.dpy, focused, root,
-                                                0, 0, &x, &y, &child_return);
+                                /*
+                                 * The window with input focus might me on a
```
fwSmit, line 338: me → be

src/x11/screen.c
```diff
@@ -386,32 +389,13 @@ const struct screen_info *get_active_screen(void)
...
-                focused = *(Window *)prop_return;
-                XFree(prop_return);
-        }
+        Window focused;
+        int ignored;
+
+        XGetInputFocus(xctx.dpy, &focused, &ignored);
 
+        if (focused == None || focused == PointerRoot)
+                focused = 0;
```
fwSmit, line 398: None is a typedef for 0L, so you can leave that out of the if statement. I would also set focused to None for clarity.  Or are there cases where None != 0L?

tsipinakis: The typedef might not change but it's always best to do what the library devs expect you to in my experience :). It's not worth it for a 5 character reduction.

fwSmit:
> This doesn't fix the general confusion between screens and monitors. Xrandr and Xinerama allow for multiple monitor to show parts of the same screen. Most references to "screens" in screen.c should be talking about monitors instead. It would be possible to handle multiple screens, but I don't think it would be worth it since most people don't use that part of X11 nowadays.

Yeah dunst's X11 implementation isn't the cleanest. Your PR defnitely helps.

As a side note, I'm wondering what do you need two X111 screens for?

mcz:
> I'm wondering what do you need two X111 screens for?
It's just the most convenient way to handle my setup. I have two monitors with my WM (openbox) only running on the main one. The secondary monitor has only suckless's tabbed on it, with a small program listening for MapRequest events and adding the appropriate windows as tabs. It also monitors the number of active tabs, and if there are none, the monitor is turned of with xrandr. This means I only have two monitors when I need them.

src/x11/screen.c
fwSmit: What do you mean? Are you for or against changing it to `focused = None`. I don't think they accidentally typedef'd None to 0.

mcz: None is typedef'd to 0L in Xorg, but that isn't guaranteed by the X11 spec. There might be a X11 implementation where None = 1L.

fwSmit: Allright then this is fine by me.

fwSmit:
Interesting setup. If you squash the commits in one it looks good to me

tsipinakis (approved):
LGTM! Thanks for the PR! (and thanks @fwSmit for the review :) )

Issue #543, opened 2018-09-11
Add option to enable a gap between the icon and the notification text


hamidsafdari:
The current implementation doesn't seem to have an option to specify some gap between the icon and the notification text. I can set a vertical/horizontal padding but all they do is create space for the overall notification. I don't want any padding around the icon but only between the text and the icon. It would also be nice if the same padding would be applied when there is no icon found.

### Installation info

- Version: `1.3.2`
- Install type: `package manager`
- Distro and version: `Archlinux 4.18.5-arch-1 `

blueForestIcarus:
I'd also like to have this feature

fwSmit:
Closed via #810

PR #810, opened 2021-01-23
Adding a config variable that adds extra space between the notification icon and text

GuessWhatBBQ:
This solves/satisfies the feature request in #543 by creating another config variable called ```text_icon_padding``` which defaults to 0, but when set with a pixel value adds that to the regular h_padding to create the desired effect.

Testing is required to see if it breaks any of the regular padding behavior.

### DIFF
--- a/dunstrc
+++ b/dunstrc
@@ -30,7 +30,7 @@
     # The width can be negative.  In this case the actual width is the
     # screen width minus the width defined in within the geometry option.
     geometry = "300x5-30+20"
-	
+
     # Turn on the progess bar
     progress_bar = true
 
@@ -40,14 +40,14 @@
 
     # Set the frame width of the progress bar
     progress_bar_frame_width = 1
-    
+
     # Set the minimum width for the progress bar
     progress_bar_min_width = 150
 
     # Set the maximum width for the progress bar
     progress_bar_max_width = 300
-    
-    
+
+
     # Show how many messages are currently hidden (because of geometry).
     indicate_hidden = yes
 
@@ -75,6 +75,9 @@
 
     # Horizontal padding.
     horizontal_padding = 8
+
+    # Padding between text and icon.
+    text_icon_padding = 0
 
     # Defines width in pixels of frame around the notification window.
     # Set to 0 to disable.
@@ -243,7 +246,7 @@
 
     # Ignore the dbus closeNotification message.
     # Useful to enforce the timeout set by dunst configuration. Without this
-    # parameter, an application may close the notification sent before the 
+    # parameter, an application may close the notification sent before the
     # user defined timeout.
     ignore_dbusclose = false
 
--- a/src/draw.c
+++ b/src/draw.c
@@ -179,6 +179,15 @@ static bool have_dynamic_width(void)
         return (settings.geometry.width_set && settings.geometry.w == 0);
 }
 
+static int get_text_icon_padding()
+{
+        if (settings.text_icon_padding) {
+                return settings.text_icon_padding;
+        } else {
+                return settings.h_padding;
+        }
+}
+
 static bool have_progress_bar(const struct notification *n)
 {
         return (n->progress >= 0 && settings.progress_bar == true);
@@ -241,7 +250,9 @@ static struct dimensions calculate_dimensions(GSList *layouts)
                         w = dim.w;
                         w -= 2 * settings.h_padding;
                         w -= 2 * settings.frame_width;
-                        if (cl->icon) w -= cairo_image_surface_get_width(cl->icon) + settings.h_padding;
+                        if (cl->icon) {
+                                w -= cairo_image_surface_get_width(cl->icon) + get_text_icon_padding();
+                        }
                         layout_setup_pango(cl->l, w);
 
                         /* re-read information */
@@ -335,7 +346,9 @@ static struct colored_layout *layout_init_shared(cairo_t *c, const struct notifi
         } else {
                 width -= 2 * settings.h_padding;
                 width -= 2 * settings.frame_width;
-                if (cl->icon) width -= cairo_image_surface_get_width(cl->icon) + settings.h_padding;
+                if (cl->icon) {
+                        width -= cairo_image_surface_get_width(cl->icon) + get_text_icon_padding();
+                }
                 layout_setup_pango(cl->l, width);
         }
 
@@ -621,7 +634,7 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width)
 
                 // icon position
                 if (settings.icon_position == ICON_LEFT) {
-                        text_x = cairo_image_surface_get_width(cl->icon) + 2 * settings.h_padding;
+                        text_x = cairo_image_surface_get_width(cl->icon) + settings.h_padding + get_text_icon_padding();
                 } // else ICON_RIGHT
         }
         cairo_move_to(c, text_x, text_y);
--- a/config.h
+++ b/config.h
@@ -54,6 +54,7 @@ struct settings defaults = {
 .separator_height = 2,       /* height of the separator line between two notifications */
 .padding = 0,
 .h_padding = 0,              /* horizontal padding */
+.text_icon_padding = 0,     /* padding between icon and text*/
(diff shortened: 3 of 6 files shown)


GuessWhatBBQ:
It would also be nice if the config file allowed more fine tuned control of padding in general. For example adding padding to text and icons separately and also specify the direction of the padding (top/bottom/left/right)

fwSmit:
The PR looks good. I haven't tested it yet, though. Could you add some documentation in `dunst.pod` about the new setting?

> This solves/satisfies the feature request in #543 by creating another config variable called `text_icon_padding` which defaults to 0, but when set with a pixel value adds that to the regular h_padding to create the desired effect.

As currently worded in dunstrc, it seems like the `text_icon_padding` replaces the horizontal padding. I think that's the most intuitive for users. Could you consider implementing it that way, instead of adding to `h_padding`?

> It would also be nice if the config file allowed more fine tuned control of padding in general. For example adding padding to text and icons separately and also specify the direction of the padding (top/bottom/left/right)

It's worth thinking about, but we should be careful making the config too long or complicated.

GuessWhatBBQ:
I see what you mean about the wording.

But just so we're clear, lets say this is the behavior without any ```text_icon_padding```
```
h_padding=3
text_icon_padding=0

_________________________
|---00000---IamAmessa---|
|---00000---ge       ---|
|---00000---Summary  ---|
|---00000---Body     ---|
‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
```
**0 = Icon 
\- = Guaranteed padding regardless of message size**

---
According to you it should be like this
```
h_padding=3
text_icon_padding=2

_________________________
|---00000--IamAmessag---|
|---00000--e         ---|
|---00000--Summary   ---|
|---00000--Body      ---|
‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
```
---
Compared to what my PR does right now:
```
h_padding=3
text_icon_padding=2

_________________________
|---00000-----IamAmes---|
|---00000-----sage   ---|
|---00000-----Summary---|
|---00000-----Body   ---|
‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾‾
```

The think an implementation like the first one would require a bit more work, but I'm concerned about whether this is truly the desired effect.

fwSmit:
I don't know what's most intuitive to users, but `text_icon_padding` suggests to me that it's the distance between the text and the icon.
About implementation, it's not much harder. You can just subtract the horizontal padding after getting the config value (make sure to check if it's zero before that though). It's a little less clean, but it makes documenting it easier.

GuessWhatBBQ:
I think that should do it

dunstrc
```diff
@@ -76,6 +76,10 @@
     # Horizontal padding.
     horizontal_padding = 8
 
+    # Padding between text and icon.
+    text_icon_padding = 0
```
fwSmit, line 80: This should be commented by default and set to a non-zero value now.

dunstrc
```diff
@@ -76,6 +76,10 @@
...
+
+
```
fwSmit, line 82: One newline too many

src/draw.c
```diff
@@ -622,6 +636,9 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width)
                 // icon position
                 if (settings.icon_position == ICON_LEFT) {
                         text_x = cairo_image_surface_get_width(cl->icon) + 2 * settings.h_padding;
```
fwSmit, line 638: This should be only 1 times h_padding

src/draw.c
```diff
@@ -335,7 +342,14 @@ static struct colored_layout *layout_init_shared(cairo_t *c, const struct notifi
         } else {
                 width -= 2 * settings.h_padding;
                 width -= 2 * settings.frame_width;
-                if (cl->icon) width -= cairo_image_surface_get_width(cl->icon) + settings.h_padding;
+                if (cl->icon) {
+                        if (settings.text_icon_padding) {
+                                width -= cairo_image_surface_get_width(cl->icon) + settings.text_icon_padding;
+                        }
+                        else {
+                                width -= cairo_image_surface_get_width(cl->icon) + settings.h_padding;
```
fwSmit, line 350: You could simplify this by adding a function `get_text_icon_padding()` which returns either h_padding or text_icon_padding. By adding that, you don't need if's here.

fwSmit:
Thank you for taking the time to change it. After these comments, it should be good.

src/draw.c
fwSmit: Yeah, the drawing code is pretty rough (and untested)

dunstrc
GuessWhatBBQ: Does it make a difference ? Not to mention none of the other disabled by default variables are commented out. I don't think that would maintain consistency

src/draw.c
GuessWhatBBQ: That modularization could also allow the fine tuning controls I mentioned before

dunstrc
fwSmit: Oh I forgot that setting it to zero disables it

src/draw.c
fwSmit: Yeah it would be good to do that. It's probably best to first modularize it, then add tests and lastly add some fine tuning controls.

src/draw.c
GuessWhatBBQ: I didnt actually modify that bit. I think it's 2*h_padding to provide the gap between icon and text the old way. Changing it would break current setups

fwSmit: Oh sorry I didn't look good enough. It's good like this :)

src/draw.c
GuessWhatBBQ: I could move a few code chunks around to make it a bit more modularized but should I make it part of this PR ?

src/draw.c
```diff
@@ -241,7 +241,14 @@ static struct dimensions calculate_dimensions(GSList *layouts)
                         w = dim.w;
                         w -= 2 * settings.h_padding;
                         w -= 2 * settings.frame_width;
-                        if (cl->icon) w -= cairo_image_surface_get_width(cl->icon) + settings.h_padding;
+                        if (cl->icon) {
+                                if (settings.text_icon_padding) {
+                                        w -= cairo_image_surface_get_width(cl->icon) + settings.text_icon_padding;
+                                }
+                                else {
```
tsipinakis, line 248:
```suggestion
                                } else {
```

src/draw.c
```diff
@@ -335,7 +342,14 @@ static struct colored_layout *layout_init_shared(cairo_t *c, const struct notifi
         } else {
                 width -= 2 * settings.h_padding;
                 width -= 2 * settings.frame_width;
-                if (cl->icon) width -= cairo_image_surface_get_width(cl->icon) + settings.h_padding;
+                if (cl->icon) {
+                        if (settings.text_icon_padding) {
+                                width -= cairo_image_surface_get_width(cl->icon) + settings.text_icon_padding;
+                        }
+                        else {
```
tsipinakis, line 349:
```suggestion
                        } else {
```

tsipinakis (changes requested):
Some code style issues, lets keep things consistent. Code looks good but I haven't tested it yet, I'll do that tomorrow.

There's also a binary file that made its way in there `src/settings.h.gch`, please remove it :).

fwSmit:
LGTM, thank you!

tsipinakis (approved):
LGTM. However, from the quick tests I did the looks of the increased padding looked weird to me. I'm curious about the usecase here (perhaps it just doesn't suit my config?).

That issue has 5 upvotes so I'll merge this either way.

GuessWhatBBQ:
I set my h_padding to 0 because I dont like the icon getting smaller. This has the added side effect of making the icon and text almost touch and that looks weird to me. So that's why the added gap is required like this.

2021-01-27-084550_456x84_scrot

fwSmit:
Makes sense. I'll merge it so I can fix more conflicts in #799.

PR #814, opened 2021-02-02
Wayland improvements

fwsmit:
This PR adds fullscreen detection for compositors that support the protocol (sway master does, but there is no release version that does yet). It also adds basic touchscreen handling copied from mako, although I haven't been able to test it since I don't have a touch device with linux.
Minor fix is made, where the idle timer would be set to 0 when idle_timeout was 0, instead of disabling the timer. This resulted in lots of debug messages.


fwSmit:
This code is for review ready now. You can test fullscreen detection only on the latest sway version. There is one edge case not being dealt with, namely two "toplevels" being fullscreen on the same output. It's possible according to the protocol, but I don't think it's possible in sway.

tsipinakis:
I'm not sure about the touch support part. It's untested as you said, no-one has requested it yet.

Doesn't wayland translate touch=left click? In which case I think leaving the default behaviour would be fine.

fwSmit:
> Doesn't wayland translate touch=left click? In which case I think leaving the default behaviour would be fine.

I don't think wayland translates touch actions to left mouse button. That would mean it triggers twice on touch. Wouldn't make much sense in code.

I feel like it can't hurt to leave the touch code in.

tsipinakis:
> I don't think wayland translates touch actions to left mouse button. That would mean it triggers twice on touch. Wouldn't make much sense in code.

My point is that my assumption for a sane default for wayland is:
```
If (application supports touch) {
    send touch events
} else {
    send touch events as left click
}
```

Is that not the case? It seems weird that for all programs to work with touch they have to be _explicitly_ coded for touch.

fwSmit:
> Is that not the case? It seems weird that for all programs to work with touch they have to be explicitly coded for touch.

I've asked on the wayland irc to be sure. It's not required by the spec to translate the touch inputs to left mouse button, thus some compositors might not do it. I don't know what compositors do or don't do it though.

Most applications won't have to be explicitly coded for touch, because they are using a GUI toolkit like QT or GTK. The wayland protocol seems to be quite low level, so maybe it will eventually make sense to have libraries that do the wayland interfacing. You can't query things like mouse position, but you have to keep track of it with position change events.

tsipinakis (approved):

tsipinakis:
Hm, in that case lets try it out. (And hope it doesn't break anything)

kajzersoze:
Fullscreen detection works with Sway 1.6 ?

fwSmit:
> Fullscreen detection works with Sway 1.6 ?

Yep

kajzersoze:
Well, I'm trying not to have notifications shown when fullscreen, under X11 this line did that `fullscreen = delay`, doesn't seem to work under Wayland, do I need something else to configure as well ?

fwSmit:
Could you open an issue about that with more info like a Dunst version included?

PR #819, opened 2021-02-10
Fix background and highlight rules not working

### LINKED ISSUE #818: Rules highlight and background not working
Tested with this in my dunstrc

```
[any]
    summary = "*"
    background = "#D62929"
    foreground = "#F9FAF9"
```

This should result in all notifications having a red background, but since bfcf655b1e13f8c5884b36203fa950a70443dbb3 that's not working.

fwSmit:
Fixes #818


tsipinakis (approved):
Should've seen this in the review, whoops!

Issue #804, opened 2021-01-08
Dunst script option doesn't provide image path from spotify's notifications


Barbarossa93:
When using a script accepting arguments from dunst's script option, the icon path of spotify notifications isn't included. I would expect there to be a path provided since I can see album art in the actual dunst notification. I know it's not an issue with the script since I can get icons for other notifications, specifically when I send a notification with a specified image.

I apologize if this is a problem on spotify's end; it's not clear to me where the problem is stemming.

### Installation info

- Version: `1.5.0`
- Install type: `package`
- Distro and version: ` Arch`

bebehei:
Spotify's notification uses a raw image. This is binary data, sent over the wire. (In comparison: The icons are just text strings, which get resolved then by dunst after receiving the notification.)

We can't provide the binary data in a meaningful way to a script.

bebehei:
See here:

https://github.com/dunst-project/dunst/blob/0e6997b6fcb9bb89c961597123945bc0075445c9/src/notification.c#L105-L143

IIRC spotify, doesn't send an iconname, which then results in an empty parameter in your script. (Which is absolutely fine, the `icon` parameter is optional in the notification spec)

Barbarossa93:
Oh bummer, figured it was something on their end.  Thanks for the quick response! I assume this is unfixable so I'll go ahead and close this.

fwSmit:
If you really want all icons to be usable by scripts, you could add something like
```diff
diff --git a/src/draw.c b/src/draw.c
index bb70b1c..2f9e1fc 100644
--- a/src/draw.c
+++ b/src/draw.c
@@ -311,6 +311,8 @@ static struct colored_layout *layout_init_shared(cairo_t *c, const struct notifi
 
         if (settings.icon_position != ICON_OFF && n->icon) {
                 cl->icon = gdk_pixbuf_to_cairo_surface(n->icon);
+                GError * error = NULL;
+                gdk_pixbuf_save(n->icon, "/tmp/dunst_icon.jpg", "jpeg", &error, NULL);
         } else {
                 cl->icon = NULL;
         }
```
and make your script use `/tmp/dunst_icon.jpg` if it doesn't get an icon via arguments. If you make the icon name the app name, it would even work for multiple notifications at the same time.

PR #823, opened 2021-02-11
 Add history_pop_id feature

### LINKED ISSUE #806: Force send spotify notifications with icon
When a new song is played in spotify, I get a notification via dunst. But sometimes I want to know the song info in the middle of the song. Hence I made a small script like this:
`dunstify "$(spotifycli --song)" "$(spotifycli --artist)"
`
The script uses spotifycli and is called via a keybinding. The problem is, it doesnt show any album artwork ( because I am not asking it show). I wanted to know how can I ask it show this info with album artwork. And is there a better way to get song info in the middle of the song?

### Installation info

- Version: `1.5.0-1`
- Install type: `pacman`
- Distro and version: `ArcoLinux`

fwSmit:
This allows users to bring back a specific notification back from
history, instead of only the last one.

Fixes #806

### DIFF
--- a/src/queues.c
+++ b/src/queues.c
@@ -351,6 +351,25 @@ void queues_history_pop(void)
 }
 
 /* see queues.h */
+bool queues_history_pop_id(int id)
+{
+        if (g_queue_is_empty(history))
+                return false;
+
+        struct notification *n = queues_get_by_id(id);
+        if (!n)
+                return false;
+
+        if (!g_queue_remove(history, n))
+                return false;
+
+        n->redisplayed = true;
+        n->timeout = settings.sticky_history ? 0 : n->timeout;
+        g_queue_insert_sorted(waiting, n, notification_cmp_data, NULL);
+        return true;
+}
+
+/* see queues.h */
 void queues_history_push(struct notification *n)
 {
         if (!n->history_ignore) {
--- a/src/queues.c
+++ b/src/queues.c
@@ -190,7 +190,7 @@ int queues_notification_insert(struct notification *n)
                 }
                 inserted = true;
         } else {
-                n->id = ++next_notification_id;
+                n->id = next_notification_id++;
         }
 
         if (!inserted && STR_FULL(n->stack_tag) && queues_stack_by_tag(n))
--- a/dunstctl
+++ b/dunstctl
@@ -21,7 +21,8 @@ show_help() {
 	  close-all                         Close the all notifications
 	  context                           Open context menu
 	  count [displayed|history|waiting] Show the number of notifications
-	  history-pop                       Pop one notification from history
+	  history-pop [ID]                  Pop latest notification from history. If ID is
+	                                    provided, pop the notification with that ID instead.
 	  is-paused                         Check if dunst is running or paused
 	  set-paused [true|false|toggle]    Set the pause status
 	  debug                             Print debugging information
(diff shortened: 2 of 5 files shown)


src/queues.c
```diff
@@ -350,6 +350,25 @@ void queues_history_pop(void)
...
+        struct notification *n = queues_get_by_id(id);
+        if (!n)
+                return false;
+
+        if (!g_queue_remove(history, n))
+                return false;
+
+        n->redisplayed = true;
+        n->timeout = settings.sticky_history ? 0 : n->timeout;
+        g_queue_insert_sorted(waiting, n, notification_cmp_data, NULL);
```
tsipinakis, line 368: This can be deduped with `queues_history_pop`. I think the cleanest way is `queues_history_pop_n(notification *n)` and make both functions use it.

tsipinakis (commented):
You also forgot to update the manpage for dunstctl :)

src/queues.c
fwSmit: Sure, I'll take a look.
Also, I noticed that when you replace a notification, the old notification still exists in history. Should we purge the replaced notification from history, or make it so that the history-pop-id function only pops the last version? Right now it pops all version one after the other

tsipinakis: It's invalid from the spec to replace a notification that has already been closed. So we can leave it as it is.

fwSmit: I don't see it mentioned in the spec

tsipinakis: From `NotificationClosed`:
> The ID specified in the signal is invalidated before the signal is sent and may not be used in any further communications with the server. 

So after that signal no further communication should reference that ID.

fwSmit: Oh then this whole feature is invalid according to the spec. I do not see a valid way to achieve this otherwise, other than resending the notification. That kind of defeats the point

fwSmit: Hmm, we could do a history-pop with app-id. That would be fine for most cases and easier to use

tsipinakis: Dunstctl is itself outside the spec either way. The spec only specifies what happens between the client and the server, anything else is extra. So in that space we can do whatever.

Thought I do have doubts about the usefulness of this feature, other than the issue linked I don't how for what else it'd be used for - it's a pretty hacky solution.

fwSmit: I think I have to agree with you there. The application could always resend a notification instead. This will probably be a more hacky solution in most cases.

fwSmit: Shall I close the PR then?

tsipinakis: Feel free to close, I'm open to implementing something like this with a valid use-case. One thing I can think of is for scripts in combination with `dunstctl`. But I'm not sure what use popping a notification from history would have.

I guess the real problem in #806 is 'you can't get the icon for a raw icon notification' which is a pretty tough one to solve.

fwSmit: > I guess the real problem in #806 is 'you can't get the icon for a raw icon notification' which is a pretty tough one to solve.

It's not really hard to make a working solution for that though. See https://github.com/dunst-project/dunst/issues/804#issuecomment-756697382 for a patch that saves the icon to a file. We could add two rules: one rule to enable icon saving to a file, and a second one to specify where the icon should be saved.

tsipinakis: That's no really a solution IMO. It adds unnecessary IO writes, and might even introduce race conditions.

A proper solution would be
a) Somehow pass the file contents to the script via bash (I don't think that's possible, bash doesn't like binary content)
b) Somehow request the icon from dunst via dunctl

Any other ideas are appreciated but these are the only paths I see currently.

fwSmit: > That's no really a solution IMO. It adds unnecessary IO writes, and might even introduce race conditions.

Race conditions maybe, but you can avoid unneccesary IO by saving the file to a tmpfs (/run and sometimes /tmp as well).

> A proper solution would be
> a) Somehow pass the file contents to the script via bash (I don't think that's possible, bash doesn't like binary content)

Maybe a file descriptor would work? But that's more complicated than it should be.

> b) Somehow request the icon from dunst via dunctl
> 
> Any other ideas are appreciated but these are the only paths I see currently.

I think we need a good way to point to old notifications then, so a better ID system. I'll leave it for now though, but if we come up with more use cases for needing to point to old notifications we can think more about it.

PR #830, opened 2021-02-19
Drop support for DUNST_COMMAND_*

tsipinakis:
These commands may be used for a DoS attack. Specifically, an
application that's running from a sandbox (e.g.  websites with
notification permission) can toggle the permissions.

This is a low severity issue but since dunstctl is available which
performs the exact same task there is no reason to leave this in.

Credit to Vít Šesták for reporting this.

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -452,6 +452,12 @@ void notification_init(struct notification *n)
         /* Process rules */
         rule_apply_all(n);
 
+        if (g_str_has_prefix(n->summary, "DUNST_COMMAND_")) {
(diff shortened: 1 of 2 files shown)


tsipinakis:
Drop in replacements for the removed commands:
* `notify-send DUNST_COMMAND_PAUSE` -> `dunstctl set-paused true`
* `notify-send DUNST_COMMAND_RESUME` -> `dunstctl set-paused false`
* `notify-send DUNST_COMMAND_TOGGLE` -> `dunstctl set-paused toggle`

PR #834, opened 2021-02-22
Wayland: fallback to X11 output when initialization fails

### LINKED ISSUE #833: 1.6.1 fails to start on wayland compositors without zwlr_layer_shell_v1 support
The new wayland code works fine on sway, wayfire and other wlroots-based compositors (and probably recent kwin as it supports layer-shell). Thanks a lot for implementing this feature!
However on a compositors without the required protocol support (weston, gnome) dunst fails to start with error message about missing protocols.

Steps to reproduce:
1. Start weston with [xwayland support enabled in weston.ini](https://wayland.freedesktop.org/xserver.html#heading_toc_j_5)
2. `systemctl --user import-envronment DISPLAY WAYLAND_DISPLAY`
3. `systemctl --user start dunst`
4. The same result is reproducible when running `dunst` from the terminal. Deleting `WAYLAND_DISPLAY` env variable allows `dunst` to start in X11 mode

Expected behavior:
`dunst` fails to register required wayland interfaces and falls back to X11 backend.

Actual results: 
```
Feb 21 15:51:54 localhost.localdomain systemd[987]: Starting Dunst notification daemon...
Feb 21 15:51:54 localhost.localdomain dunst[13765]: ERROR: compositor doesn't support zwlr_layer_shell_v1
Feb 21 15:51:54 localhost.localdomain systemd-coredump[13769]: Process 13765 (dunst) of user 1000 dumped core.

                                                               Stack trace of thread 13765:
                                                               #0  0x00007f8340559913 g_logv (libglib-2.0.so.0 + 0x5b913)
                                                               #1  0x00007f8340559b83 g_log (libglib-2.0.so.0 + 0x5bb83)
                                                               #2  0x00005632bd9e0ebe wl_init (dunst + 0x14ebe)
                                                               #3  0x00005632bd9d8633 main (dunst + 0xc633)
                                                               #4  0x00007f834034b1e2 __libc_start_main (libc.so.6 + 0x281e2)
                                                               #5  0x00005632bd9d8cee _start (dunst + 0xccee)
```

### Installation info

- Version: `1.6.1`
- Install type: package (found while preparing update for Fedora)
- Distro and version: ` Fedora 33`

fwSmit:
Fixes #833

### DIFF
--- a/src/output.c
+++ b/src/output.c
@@ -8,7 +8,7 @@
 #include "wayland/wl.h"
 #endif
 
-const bool is_running_wayland(void) {
+bool is_running_wayland(void) {
         char* wayland_display = getenv("WAYLAND_DISPLAY");
         return !(wayland_display == NULL);
 }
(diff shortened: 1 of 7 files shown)


alebastr:
Works great for all environments I tried. Thanks for the quick fix!

src/output.c
```diff
@@ -55,18 +55,40 @@ const struct output output_wl = {
...
+}
+
+#ifdef ENABLE_WAYLAND
+const struct output* get_wl_output() {
+        const struct output* output = &output_wl;
+        if (output->init()) {
+                return output;
+        } else {
+                LOG_W("Couldn't initialize wayland output. Falling back to X11 output.");
+                output->deinit();
```
tsipinakis, line 74: `deinit` is a noop for wayland. Perhaps we should properly deinit to avoid any interactions with having a partially initialized backend.

fwSmit: That was my intention, but the deinit was linked to the wrong function. I linked it now, and fixed the wayland deinit. I'm not very confident everything is deallocated, so take a close look at it. It doesn't crash at least when exiting from every possible abort point.

tsipinakis: Deallocation isn't that important, it's only a few bytes. The main point is to make sure that no events fire on the wayland side. This would confuse the main logic and be a debugging hell :)

tsipinakis (approved):
LGTM

PR #840, opened 2021-03-10
queues: correctly sort notifications even when sort is false.

### LINKED ISSUE #838: Disabled Sorting Order Is Not Honored While Dunst Is Paused
I have the "sort" option set to false within the Dunst configuration, which currently sorts notifications in the following, vertically stacked, order:
**10
9
8
7
6
5
4
3
2
1**

When generated as follows:
**notify-send 1
notify-send 2
notify-send 3
notify-send 4
notify-send 5
notify-send 6
notify-send 7
notify-send 8
notify-send 9
notify-send 10**

However, when pausing Dunst, the notification order seems`*` to change back to the default behavior, as if the "sort" option was set to true, where the ordering of the numbered notifications shown above is reversed (_1..10 top to bottom, rather than 10..1_), but only while Dunst is paused.  For example, the following:
**notify-send 1
notify-send 2
notify-send 3
notify-send 4
dunstctl set-paused true
notify-send 5
notify-send 6
dunstctl set-paused false
notify-send 7
notify-send 8
dunstctl set-paused true
notify-send 9
notify-send 10
dunstctl set-paused false**

Generates:
**8
7
4
3
2
1
5
6
9
10**

But the expected results were that it would show the notifications in the same order, whether or not Dunst is being paused/unpaused, ie:
**10
9
8
7
6
5
4
3
2
1**

Looking through the issue history, I found past bug reports related to the sort option:
[#Add option to always sort messages chronologically based on when they arrived? · Issue #301](https://github.com/dunst-project/dunst/issues/301)
[#Queue handling with settings.sort · Issue #431](https://github.com/dunst-project/dunst/issues/431)

But I don't think they're directly related to this bug, as they don't reference the sorting behavior while Dunst is paused.

`*` Also, it's worth noting that the urgency of the notification doesn't seem to impact the ordering when the "sort" config option is disabled, but it definitely does when this option is enabled, and seems to behave as described while Dunst is either paused or unpaused.

### Installation info

- Version: Dunst - A customizable and lightweight notification-daemon v1.6.1
- Install type: manual, from source
- Distro and version: Xorg using Linux distro

fwSmit:
The sort name is a bit misleading. It should actually be called
sort_urgency or something.
Notifications are still not sorted based on time, but ID.

Fixes #838


fwSmit:
I fixed the leak in the tests.
I also added a workaround for a glib bug that was annoying me. It produced duplicated lines in the test output. It should be fixed in glib, but it takes a while to trickle down

PR #854, opened 2021-04-19
wayland hidpi support

### LINKED ISSUE #836: Wayland HiDPI Font Scaling Blurriness
When scaling my HiDPI screens in Wayland, the text is blurry. The font is the same between dunst and the terminal/waybar.

I'm happy to try any tweaking or configuration ideas. Setting `GDK_SCALE=2` didn't seem to make a difference.

Monitor (scaling=1.3)
dunst-monitor

Laptop screen (scaling=2)
dunst-laptop

### Installation info

- Version: `1.6.1`
- Install type: `modified Debian package`
- Distro and version: `Debian Bullseye`

fwSmit:
Fixes #836
@bebehei This PR currently only implements scale detection on wayland. I'm making this PR already so you know I'm working on it.
To test it on sway, it's really easy:
`swaymsg output "*" scale 1.5`

Sway will message a scale of 2 to dunst and it will scale it down to 1.5 afterwards.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -141,10 +141,11 @@ static struct color layout_get_sepcolor(struct colored_layout *cl,
 
 static void layout_setup_pango(PangoLayout *layout, int width)
 {
+        int scale = output->get_scale();
         pango_layout_set_wrap(layout, PANGO_WRAP_WORD_CHAR);
-        pango_layout_set_width(layout, width * PANGO_SCALE);
+        pango_layout_set_width(layout, width * scale * PANGO_SCALE);
         pango_layout_set_font_description(layout, pango_fdesc);
-        pango_layout_set_spacing(layout, settings.line_height * PANGO_SCALE);
+        pango_layout_set_spacing(layout, settings.line_height * scale * PANGO_SCALE);
 
         PangoAlignment align;
         switch (settings.align) {
@@ -192,9 +193,20 @@ static bool have_progress_bar(const struct notification *n)
         return (n->progress >= 0 && settings.progress_bar == true);
 }
 
+static void get_text_size(PangoLayout *l, int *w, int *h, int scale) {
+        pango_layout_get_pixel_size(l, w, h);
+        // scale the size down, because it may be rendered at higher DPI
+
+        if (w)
+                *w /= scale;
+        if (h)
+                *h /= scale;
+}
+
 static struct dimensions calculate_dimensions(GSList *layouts)
 {
         struct dimensions dim = { 0 };
+        int scale = output->get_scale();
 
         const struct screen_info *scr = output->get_active_screen();
         if (have_dynamic_width()) {
@@ -221,10 +233,10 @@ static struct dimensions calculate_dimensions(GSList *layouts)
         for (GSList *iter = layouts; iter; iter = iter->next) {
                 struct colored_layout *cl = iter->data;
                 int w=0,h=0;
-                pango_layout_get_pixel_size(cl->l, &w, &h);
+                get_text_size(cl->l, &w, &h, scale);
                 if (cl->icon) {
-                        h = MAX(cairo_image_surface_get_height(cl->icon), h);
-                        w += cairo_image_surface_get_width(cl->icon) + settings.h_padding;
+                        h = MAX(get_icon_height(cl->icon, scale), h);
+                        w += get_icon_width(cl->icon, scale) + settings.h_padding;
                 }
                 h = MAX(settings.notification_height, h + settings.padding * 2);
                 dim.h += h;
@@ -250,15 +262,15 @@ static struct dimensions calculate_dimensions(GSList *layouts)
                         w -= 2 * settings.h_padding;
                         w -= 2 * settings.frame_width;
                         if (cl->icon) {
-                                w -= cairo_image_surface_get_width(cl->icon) + get_text_icon_padding();
+                                w -= get_icon_width(cl->icon, scale) + get_text_icon_padding();
                         }
                         layout_setup_pango(cl->l, w);
 
                         /* re-read information */
-                        pango_layout_get_pixel_size(cl->l, &w, &h);
+                        get_text_size(cl->l, &w, &h, scale);
                         if (cl->icon) {
-                                h = MAX(cairo_image_surface_get_height(cl->icon), h);
-                                w += cairo_image_surface_get_width(cl->icon) + settings.h_padding;
+                                h = MAX(get_icon_height(cl->icon, scale), h);
+                                w += get_icon_width(cl->icon, scale) + settings.h_padding;
                         }
                         h = MAX(settings.notification_height, h + settings.padding * 2);
                         dim.h += h;
@@ -299,6 +311,7 @@ static struct colored_layout *layout_init_shared(cairo_t *c, const struct notifi
 {
         struct colored_layout *cl = g_malloc(sizeof(struct colored_layout));
         cl->l = layout_create(c);
+        int scale = output->get_scale();
 
         if (!settings.word_wrap) {
                 PangoEllipsizeMode ellipsize;
@@ -346,7 +359,7 @@ static struct colored_layout *layout_init_shared(cairo_t *c, const struct notifi
                 width -= 2 * settings.h_padding;
                 width -= 2 * settings.frame_width;
                 if (cl->icon) {
-                        width -= cairo_image_surface_get_width(cl->icon) + get_text_icon_padding();
+                        width -= get_icon_width(cl->icon, scale) + get_text_icon_padding();
                 }
                 layout_setup_pango(cl->l, width);
         }
@@ -367,6 +380,7 @@ static struct colored_layout *layout_from_notification(cairo_t *c, struct notifi
 {
 
         struct colored_layout *cl = layout_init_shared(c, n);
+        int scale = output->get_scale();
 
         /* markup */
         GError *err = NULL;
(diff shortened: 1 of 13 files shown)


fwSmit:
@adamcstephens  @bebehei The wayland HiDPI support is implemented! Could you test it? With an integer scaling factor everything should look razor sharp, and with a non-integer scaling factor it is rendered in the nearest scaling above that and scaled down from there. Everything should work pretty well. The only thing I've noticed is that some icons are smaller than when redered with a scale of 1, but that can probably be solved with min_icon_size.
(I'll fix the tests later)

fwSmit:
The tests are now fixed. Icons are now also scaling properly, even when they're low resolution. The only bug remaining, which I probably won't fix is that the icons are not immediately updated on scaling. Only new notification have the bigger icons.

adamcstephens:
I'm not seeing a difference with this patch. Let me know if there's any info I can provide to help, and I'm also on IRC.

fwSmit:
What compositor are you using? This is tested on sway. Debug output of Dunst would help as well

adamcstephens:
I'm using sway 1.5.1 on Debian Bullseye.

Here's the debug output:

```
$ ~/bin/dunst -verbosity debug
INFO: Using Wayland output
INFO: New output found - id 0
DEBUG: Ignoring wake up
DEBUG: RUN
DEBUG: Fullscreen queried: 0
INFO: Idle status queried: 0
INFO: Wayland: Hiding window
WARNING: No icon found in path: 'dialog-information'
DEBUG: Waking up
DEBUG: RUN
DEBUG: Fullscreen queried: 0
INFO: Idle status queried: 0
DEBUG: Buffer size (scaled) 300x60
DEBUG: Sleeping for 59959 ms
INFO: Pointer handle button 272: 1
INFO: Pointer handle button 272: 0
DEBUG: Waking up
DEBUG: RUN
DEBUG: Fullscreen queried: 0
INFO: Idle status queried: 0
DEBUG: Queues: Closing notification for reason: user
INFO: Wayland: Hiding window
```

fwSmit:
What scale are you using? A non-integer scale will not look perfectly sharp. Try a scale of 2 and see if it looks better with this PR than before

paretje:
@fwSmit Judging by mako's pango init, I suspect you need to set the pango [scale attribute](https://developer.gnome.org/pango/1.46/pango-Text-Attributes.html#pango-attr-scale-new) for better non-integer scaling of text.

fwSmit:
> @fwSmit Judging by mako's pango init, I suspect you need to set the pango [scale attribute](https://developer.gnome.org/pango/1.46/pango-Text-Attributes.html#pango-attr-scale-new) for better non-integer scaling of text.

No unfortunately the wayland protocol only allows for integer scaling to be passed to an application (probably with good reason). The way sway allow for non-integer scaling is by passing the next integer above your scaling and scale it down afterwards. Most of the times it shouldn't matter, because you'll use it when you have a 4k screen which will work perfectly with 2x scaling.

This is what it looks like on my pc (2x scale):
20210428_22h27m26s_grim
Without HiDPI support

20210428_22h27m57s_grim
With HiDPI support

The text is definitely improved compared to before, but the icon might be worse, because it's blurry. I haven't looked into it yet, but I suspect the scaling method we use with gdk is different from the scaling method used by sway.

Another notification (2x scale):
20210428_22h42m13s_grim
Without HiDPI support

20210428_22h41m59s_grim
With HiDPI support

The line height seems to be slightly higher with this PR, so I'll look into if the pango scale function might help with that. Have you tried this PR by the way @paretje?

paretje:
Ah, sorry, it was the most obvious difference when I had a quick look at the codebases. I didn't notice scale was an int ...

On my personal laptop, I have a 14'" QHD screen, and I apply a scaling factor of 1.25. Since I switched to sway, I have been using mako as my notification daemon, which works fine, but I'm missing some features. I knew dunst had Wayland support now, and wanted to try it. I ended up trying your PR, but I noticed the text to be very blurry. I don't have this issue with mako, which was why I ended having a look at the code.

dunst:
dunst

mako:
mako

fwSmit:
Okay from your screenshot it looks like it can definitely be done better. I'll have to experiment a bit and maybe copy some code. Does a scaling factor of 2 not have any issues on your machine?

paretje:
Hmmm, yes, it doesn't really work with scaling factor 2 either for me.

fwSmit:
Can you do a picture with the latest Dunst release as well?

fwSmit:
It seems like they are exactly the same. I think there must be something wrong in the scale detection then. I've added a debug print to see what scale you are using.

fwSmit:
After zooming into those images I've found at least one difference that has nothing to do with the scaling. Mako apparently uses subpixel rendering for their fonts. Still, the 1.25x result from dunst doesn't look great.

Below are zoomed in versions of you screenshots:

Mako 1.25x
20210430_14h56m40s_grim

Dunst 1.25x
20210430_14h56m53s_grim

Dunst 2x
20210430_14h57m04s_grim

paretje:
Yes, I had noticed it and tried it, but it didn't really help. But, I now applied this additional diff:
```diff
diff --git a/src/draw.c b/src/draw.c
index f3e3a6a..5df7298 100644
--- a/src/draw.c
+++ b/src/draw.c
@@ -142,6 +142,8 @@ static struct color layout_get_sepcolor(struct colored_layout *cl,
 static void layout_setup_pango(PangoLayout *layout, int width)
 {
         int scale = output->get_scale();
+        LOG_D("Setup pango with scale %i", scale);
+
         pango_layout_set_wrap(layout, PANGO_WRAP_WORD_CHAR);
         pango_layout_set_width(layout, width * scale * PANGO_SCALE);
         pango_layout_set_font_description(layout, pango_fdesc);

```

And now I get (using 1.25):
```
MESSAGE: 'allow_markup' is deprecated, please use 'markup' instead.
INFO: Max width was set but max_icon_size is unlimited. Limiting icons to 150 pixels
MESSAGE: The option 'icon_folders' is deprecated, please use 'icon_path' instead.
MESSAGE: The frame section is deprecated, width has been renamed to frame_width and moved to the global section.
MESSAGE: The frame section is deprecated, color has been renamed to frame_color and moved to the global section.
INFO: Using Wayland output
INFO: New output found - id 0
DEBUG: Output scale changed to 2
DEBUG: Ignoring wake up
DEBUG: RUN
DEBUG: Fullscreen queried: 0
INFO: Idle status queried: 0
INFO: Wayland: Hiding window
DEBUG: Waking up
DEBUG: RUN
DEBUG: Fullscreen queried: 0
INFO: Idle status queried: 0
DEBUG: Setup pango with scale 1
DEBUG: Buffer size (scaled) 300x52
DEBUG: Sleeping for 9967 ms
```

So, while the output scale is updated to 2, which sounds right given it's an integer, when we actually setup pango, `output->get_scale()` returns 1.

paretje:
Oh, now I see why. `wl_get_scale` always returns 1 unless you use a fixed monitor? I can confirm that setting follow to none fixes the scaling for me.
```c
int wl_get_scale(void) {
        struct dunst_output *output = get_configured_output();
        if (output)
                return output->scale;
        else
                return 1;
}
```

fwSmit:
Thanks for debugging. The get_scale function may indeed return the default of 1 when it cannot detect what the current output is. This happens when `follow` is something else than `none`.  We then leave it up to the compositor to decide on what output to place the dunst window, making it harder (but not impossible) for us to know what output the window is currently on. On my system when I do this, the text becomes way bigger than it's supposed to be
It should now work for the case where there is no output though, because it now returns that output instead of NULL (let the compositor decide). Do you have multiple monitors, or is the follow value just there doing nothing?
I'll try out if taking the largest scale factor among the monitors and see if the compositor scales the buffer down if needed.

fwSmit:
It should use the right scale in any monitor configuration. It works fine to use a larger scale if the current output cannot be found.
I'll still have to look if the text rendering can be more crisp, but it doesn't necessarily have to be in this PR if the text looks good enough.

paretje:
No, I don't use multiple monitors, it probably was just part of the default config at the time. When I sit at my desk, I use an external monitor, but my laptop monitor is disabled.

dunst does crash when I switch between monitors though. I don't think (dis)connecting monitors is actually handled at the moment? The immediate cause is the fact that the initial value of scale is 0, so if for some reason there is no output, scale 0 is used, resulting in a division by zero in `get_text_size`. I suggest to use 1 for robustness.

fwSmit:
Okay thanks for mentioning. That will be an easy fix. Have you compared the text between Dunst and mako now? Is there still a lot of difference or only the subpixel rendering?

paretje:
No, dunst looks fine now, thank you.

fwSmit:
> No, dunst looks fine now, thank you.

Cool! The divide by 0 issue is fixed as well. I'll see if I'll implement subpixel rendering as well. It should be only a few lines of code.

adamcstephens:
I can confirm that this PR looks good now, and you've resolved the blurriness for me. It works for multiple monitors with different scales.

Tested on:
* single screen at scale of 2.2
* dual screen, scales of 1.32 and 2.2

fwSmit:
> 
> 
> I can confirm that this PR looks good now, and you've resolved the blurriness for me. It works for multiple monitors with different scales.
> 
> Tested on:
> 
>     * single screen at scale of 2.2
> 
>     * dual screen, scales of 1.32 and 2.2

Did you try it with `follow=none` or something else? When `follow` is mouse or keyboard, it'll be rendered at 2.2x res and scaled down to 1.32 if needed.

adamcstephens:
> Did you try it with `follow=none` or something else? When `follow` is mouse or keyboard, it'll be rendered at 2.2x res and scaled down to 1.32 if needed.

I just tested both `follow=none` and `follow=mouse`, with either display as primary, and it looks good in every scenario I tested.

fwSmit:
It sounds like you tested it pretty well. I don't see any issues either, except the icons not resizing immediately upon changing the scale. Subpixel font rendering will have to wait a bit as well, since it's not as easy as I thought. This PR is finished, though. We'll just have to wait for someone to merge it (might take a while, since the maintainers are busy).

tsipinakis:
Sorry for the delay in merging this. I can't do much to test it as I'm not on wayland (yet) but I'll trust you saying it looks good. :+1:

Issue #233, opened 2015-07-21
Horizontally centered banners (GNOME 3.16+-like)


sgtpep:
It would be great to be able to place notifications like GNOME do since 3.16: https://wiki.gnome.org/Design/OS/Notifications.

tsipinakis:
This can be partly achieved with the current `geometry` setting, just set the X to +/- half the width of the display. The window won't expand based on the notification length but it's a start.

jzbor:
It would be nice if this setting where independent from the screen width as it may vary on a laptop with external screen

victorz:
> This can be partly achieved with the current `geometry` setting, just set the X to +/- half the width of the display. The window won't expand based on the notification length but it's a start.

With this method, you'd have to also subtract half of the notification width, so `horiz_offset = (screen_width - notification_width) / 2`. Still, this won't work properly with `shrink = yes`, because that makes the notification width dynamic, and the offset can't be made dynamic.

fwSmit:
This issue is solved with #855 (WIP)

WhyNotHugo:
I believe this can be closed with #855 merged?

Issue #82, opened 2012-12-21
Height should take into account that notifications differ in height


Markus00000:
If I am not mistaken, setting height to 10 is intended to allow dunst grow to 10 lines.

Now that there can be a frame, padding, a larger line height and lines that wrap, it is hard to set height to a sensible value.

About 8 to 30 notifications will fit my screen depending on their content. What value would I set height to?

Therefore I propose to measure height in pixels (as is width, x, y, padding and frame).

Setting height to a number of pixels would cause any notification that does not fit into the remaining pixels to be queued until there are enough free pixels. Or expressed more simply:

```
if window_height_max - window.height >= notification.height:
    notification.show()
else:
    notification.queue()
```

WhyNotHugo:
I believe this can be closed with #855 merged?

fwSmit:
> I believe this can be closed with #855 merged?

Yes, thank you :+1:

PR #855, opened 2021-04-22
Redesigning the window geometry/dimension settings

### LINKED ISSUE #336: Redesigning the window geometry/dimension settings

### LINKED ISSUE #518: Not correctly wrapping text with shrink enabled.
Setting a max width on notifications, along with shrink and word_wrap seems to wrap content based on a notification's min width instead of the width of any already present notifications.

Center aligned
scrot

Right aligned
scrot

### Installation info
- Version: 1.3.2
- Install type: package
- Distro and version: Void Linux

fwSmit:
This PR implements #336, building on top of settings refactor v2
See https://github.com/dunst-project/dunst/issues/336#issuecomment-812845742 for a description of the latest idea
Fixes #336 #82 #233
Seems to fix #518


fwSmit:
Weird. No CI is running in this PR anymore

fwSmit:
It somehow works again. Sorry for extra PR

etrnal70:
Just tried this PR today. Is the `offset` setting broken ? Because I don't see any change when set it with any value. I'm using wayland on sway

docs/dunst.5.pod
```diff
@@ -64,47 +64,59 @@ keyboard focus.
...
+        top-center
+        top-right
+        bottom-left
+        bottom-center
+        bottom-right
+        left-center
+        center
+        right-center
+
+=item B<offset> format: (horizontal, vertical)
```
etrnal70, line 111: Looks like the format is wrong, should be like `10x300`

etrnal70:
image

Shouldn't hidden notification only show hidden count ? (No icon and progress bar)

docs/dunst.5.pod
fwSmit: Thank you! I've updated it. I think that immediately solves your other problem as well.

fwSmit:
> image
> 
> Shouldn't hidden notification only show hidden count ? (No icon and progress bar)

I'll consider this a separate issue. While investigating this, I also noticed that the timeout of the hidden notifications is also ignored.

fwSmit:
Thanks for testing @etrnal70. I'll go ahead and merge this soon.

docs/dunst.5.pod
```diff
@@ -64,47 +64,59 @@ keyboard focus.
...
 
-It's important to note that the positive and negative sign B<DOES> affect the
-position even if the offset is 0. For example, a horizontal offset of +0 puts
-the notification on the left border of the screen while a horizontal offset of
--0 at the right border. The same goes for the vertical offset.
-
-=back
+Examples:
+        origin = top-right
+        offset = (10, 300) # a margin of 10 pixels from the right and 300 pixels from the top
```
etrnal70, line 119: I forgot this, example should be `10x300`

fwSmit: Thank you, it's fixed now

PR #856, opened 2021-04-24
Fix link to "Building" section in readme

mgsloan:


bebehei:
In the mean time Github provides a ToC automatically. There is no need to assemble a ToC manually.

I'd guess you can remove the whole ToC section.

bebehei:
Removing the ToC section would also prevent such mistakes at their source.

fwSmit:
Don't we use the same file somewhere on the website?

PR #858, opened 2021-05-01
Add CFLAGS with -Werror for CI

bebehei:
This should fix dunst-project/docker-images#4

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -6,6 +6,7 @@ DATADIR ?= ${PREFIX}/share
 # around for backwards compatibility
 MANPREFIX ?= ${DATADIR}/man
 MANDIR ?= ${MANPREFIX}
+EXTRA_CFLAGS ?=
 
 DOXYGEN ?= doxygen
 FIND ?= find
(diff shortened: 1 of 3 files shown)


bebehei:
As now seen in the CI: `-Werror` is used.

It's not possible to inject CFLAGS via environment in `make`, since this overwrites the default variable. :-/

I think this is more favorable over hardcoding `-Werror` in the docker images for testing.

config.mk
```diff
@@ -6,6 +6,7 @@ DATADIR ?= ${PREFIX}/share
 # around for backwards compatibility
 MANPREFIX ?= ${DATADIR}/man
 MANDIR ?= ${MANPREFIX}
+CFLAGS_CI ?=
```
fwSmit, line 9: I would call it `EXTRA_CFLAGS` and document it somewhere. It might be useful for other things as well.

fwSmit:
I also thought about just adding -Werror to the DEFAULT_CFLAGS, but that might be a bit annoying in development.

config.mk
bebehei: I've renamed it to `EXTRA_CFLAGS`.

bebehei:
Having `-Werror` hardcoded in the Makefile is very annoying, especially for users compiling the code on another distro, which potentially throws new errors.

fwSmit:
> Having `-Werror` hardcoded in the Makefile is very annoying, especially for users compiling the code on another distro, which potentially throws new errors.

Agreed. 

By the way, I would expect at least the arch build to fail now, because on my system it gives a warning about a using deprecated function. This happens in a recent version of glib. Do the images have to be updated before the warning will show up?

bebehei:
> Do the images have to be updated before the warning will show up?

Yes.

fwSmit:
Am I able to update the images as well or do you need some sort of credentials?

fwSmit (approved):

bebehei:
Afaik you have to have a docker hub account. I can add you to the orga then.

All you have to do, should be a `make build` locally and a `make push`.

It'll then push the freshly built images to docker hub.

fwSmit:
Okay I made an account. My username is `fwsmit`. I don't think I'll need access immediately, but it'll be useful to be able to fix it when things go wrong. Also, I've made a fix for the warning from glib: 1d3822ee8b2263be472e13ea685e7beb69071fb1. Before that is merged the images shouldn't be updated.

bebehei:
I've added you in the docker project. You're now an owner there. You can update it whenever you want. If you need help, I can give you suppoort.

PR #860, opened 2021-05-03
Update Pango Markup reference link

ohadeytan:
It seems the old [link](https://developer.gnome.org/pango/stable/pango-Markup.html) is unreachable.

### DIFF
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -292,7 +292,7 @@ Allow a small subset of html markup in notifications
     <u>underline</u>
 
 For a complete reference see
-<https://developer.gnome.org/pango/stable/pango-Markup.html>
+<https://developer.gnome.org/pygtk/stable/pango-markup-language.html>
 
 =item B<strip>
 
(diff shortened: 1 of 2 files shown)


bebehei:
The whole `stable` docs branch seems to be unavailable for the pango project. Not just the markup page.

Using a link, [pinning `1.46` as the version](https://developer.gnome.org/pango/1.46/pango-Markup.html), makes the page available again.

But linking to the `pygtk` project doesn't make any sense for me. We don't use it at all.

One should find out whether the unavailable `stable` branch is a temporary mistake of the Gnome devs or whether it was intent.

bebehei:
[I've asked the GNOME devs directly.](https://gitlab.gnome.org/GNOME/pango/-/issues/557).

I'm disliking the pygtk link, since the docs are built on 2011-03-12. This won't be a good solution either. **This is over ten years old now!**

ohadeytan:
You are probably right, I'm closing. 
Sorry for the rushy pull request.

PR #867, opened 2021-05-14
Add icon-static server capability

hastinbe:
Dunst supports displaying the first frame of an icon, thus it seems reasonable that it should announce that capability with `icon-static`.

`"icon-static" | Supports display of exactly 1 frame of any given image array. This value is mutually exclusive with "icon-multi", it is a protocol error for the server to specify both.` 
from https://developer.gnome.org/notification-spec/#hints

I check for this capability in another project. notify-osd is the only other I know that supports/announces it, however dunst doesn't when it could since it has this already. I've been relying upon the `image-path` hint as an alternative which is more or less a fire and forget approach but it'd be nice to know if the capability exists


fwSmit:
I didn't know we supported this :-)
LGTM, thank you.

fwSmit (approved):

PR #883, opened 2021-06-28
fixed doxygen generation

FuJa0815:
`make doc-doxygen` did not run, because some parameters were not named correctly


fwSmit:
Thank you for fixing. I myself don't use the doxygen documentation, which is why it's not correct everywhere.
I'm curious, what did you want to use the doxygen for?

bebehei (approved):
Thank you, this seems to be good.

bebehei:
> I'm curious, what did you want to use the doxygen for?

We've initially added doxygen to have a standard and baseline to have documented all the functions. If you use doxygen, you can easily have other code browsers, too. Some people might not want to use `git grep`.

PR #884, opened 2021-06-28
X11: enable dpi scaling

fwSmit:
Note: since dpi scaling was originally implemented for wayland which
only supports integer scaling (and let compositors do the non-integer
scaling from there), the X11 implementation only supports integer
scaling as well.
The scaling factor is derived from the active screen's dpi.

### DIFF
--- a/src/x11/x.c
+++ b/src/x11/x.c
@@ -943,7 +943,11 @@ static void x_shortcut_init(struct keyboard_shortcut *ks)
 }
 
 int x_get_scale(void) {
-        return 1;
+        const struct screen_info *scr_info = get_active_screen();
+        int scale = round(scr_info->dpi/96.);
+        LOG_D("X11 dpi: %i", scr_info->dpi);
+        LOG_D("X11 scale: %i", scale);
+        return scale;
 }
 
 /* vim: set ft=c tabstop=8 shiftwidth=8 expandtab textwidth=0: */


fwSmit:
@frebib You can go ahead and try this PR. You can set the scaling factor by changing the monitor's dpi value. Note that only integer scaling is supported.

frebib:
I'm not sure if I'm doing this correctly, but when scaled to 192 dpi, it seems to do this
https://frebib.net/s/18ef8bd.png

After scaling back to 96 it displays fine
https://frebib.net/s/66360fd.png

Icons seem to scale strangely too, like they're one update behind. They don't scale up, but they do scale up after I've set the dpi back to 96 again, which is odd. Maybe I'm holding it wrong, I can't tell. It seems that also the geometry updates are delayed too?

I'll say that I don't actually use integer scaling myself. My laptop is ~3440 wide (iirc) and I tend to use 1.5x scale (144dpi)

src/x11/x.c
```diff
@@ -943,7 +943,11 @@ static void x_shortcut_init(struct keyboard_shortcut *ks)
 }
 
 int x_get_scale(void) {
-        return 1;
+        const struct screen_info *scr_info = get_active_screen();
+        int scale = round(scr_info->dpi/96.);
```
frebib, line 947: Throwing away fractional scaling this easily seems a waste. dunst already supports fractional font scaling. Can't we do the same for geometry?

fwSmit: The problem is that wayland only supports integer scaling. It would be more work to change that to fractional scaling. For now I've just repurposed the scaling code from wayland.

> dunst already supports fractional font scaling. Can't we do the same for geometry?

In what way do we support fractional font scaling? By setting the font size?

fwSmit:
> I'm not sure if I'm doing this correctly, but when scaled to 192 dpi, it seems to do this
> https://frebib.net/s/18ef8bd.png
> 
> After scaling back to 96 it displays fine
> https://frebib.net/s/66360fd.png
> 
> Icons seem to scale strangely too, like they're one update behind. They don't scale up, but they do scale up after I've set the dpi back to 96 again, which is odd. Maybe I'm holding it wrong, I can't tell. It seems that also the geometry updates are delayed too?
> 
> I'll say that I don't actually use integer scaling myself. My laptop is ~3440 wide (iirc) and I tend to use 1.5x scale (144dpi)
> 
> Edit: lemme work out how to turn on debug logging and I'll see what it thinks it's doing:
> It's picking up the scale, but not actually scaling the geometry it seems
> 
> ```
> Jun 28 12:42:01 frebib-PC.nerdhouse.io dunst[80375]: DEBUG: X11 dpi: 192
> Jun 28 12:42:01 frebib-PC.nerdhouse.io dunst[80375]: DEBUG: X11 scale: 2
> ```

Could you try setting a multiple of 92 as dpi? For example 184.

vincentbernat:
It doesn't work for me either. If I have two notifications, the second one is not displayed until a third one comes. Using a multiple of 92 does not help.

vincentbernat:
The following patch fix it for me:

```diff
diff --git a/src/draw.c b/src/draw.c
index f3e3a6a38874..b12486951342 100644
--- a/src/draw.c
+++ b/src/draw.c
@@ -289,6 +289,9 @@ static struct dimensions calculate_dimensions(GSList *layouts)
                 dim.w = text_width + 2 * settings.h_padding;
                 dim.w += 2 * settings.frame_width;
         }
+        dim.w *= scale;
+        dim.h *= scale;
+        dim.corner_radius *= scale;

         return dim;
 }
```

vincentbernat:
Looking at the code, I see this is not totally correct, notably when there is a negative width or when shrinking is enabled or when there is a progress bar. Also, I didn't test with Wayland either...

fwSmit:
> Looking at the code, I see this is not totally correct, notably when there is a negative width or when shrinking is enabled or when there is a progress bar. Also, I didn't test with Wayland either...

Yeah, but it's a good starting point. It seems from that code that it's mainly the window dimensions that were wrong in my patch and that the text scaling worked.
The best way to fix it is to scale the width and height as late as possible. So only once you starting drawing or requesting a window from x should you scale it. That's also the way I did it for wayland.

vincentbernat:
I get this patch instead:

```diff
diff --git a/src/x11/x.c b/src/x11/x.c
index 09405bd0cc6a..f504095f457a 100644
--- a/src/x11/x.c
+++ b/src/x11/x.c
@@ -161,8 +161,9 @@ static bool x_win_composited(struct window_x11 *win)
 void x_display_surface(cairo_surface_t *srf, window winptr, const struct dimensions *dim)
 {
         struct window_x11 *win = (struct window_x11*)winptr;
-        x_win_move(win, dim->x, dim->y, dim->w, dim->h);
-        cairo_xlib_surface_set_size(win->root_surface, dim->w, dim->h);
+        int scale = x_get_scale();
+        x_win_move(win, dim->x - dim->w * (scale - 1), dim->y, dim->w * scale, dim->h * scale);
+        cairo_xlib_surface_set_size(win->root_surface, dim->w * scale, dim->h * scale);

         XClearWindow(xctx.dpy, win->xwin);

@@ -171,7 +172,7 @@ void x_display_surface(cairo_surface_t *srf, window winptr, const struct dimensi
         cairo_show_page(win->c_ctx);

         if (settings.corner_radius != 0 && ! x_win_composited(win))
-                x_win_corners_shape(win, dim->corner_radius);
+                x_win_corners_shape(win, dim->corner_radius * scale);
         else
                 x_win_corners_unshape(win);

```

Note the hack around `dim->x` which is not correct in all cases. I wonder how it works with Wayland as you never use `dim.x` and `dim.y`.

fwSmit:
> I get this patch instead:
> 
> ```diff
> diff --git a/src/x11/x.c b/src/x11/x.c
> index 09405bd0cc6a..f504095f457a 100644
> --- a/src/x11/x.c
> +++ b/src/x11/x.c
> @@ -161,8 +161,9 @@ static bool x_win_composited(struct window_x11 *win)
>  void x_display_surface(cairo_surface_t *srf, window winptr, const struct dimensions *dim)
>  {
>          struct window_x11 *win = (struct window_x11*)winptr;
> -        x_win_move(win, dim->x, dim->y, dim->w, dim->h);
> -        cairo_xlib_surface_set_size(win->root_surface, dim->w, dim->h);
> +        int scale = x_get_scale();
> +        x_win_move(win, dim->x - dim->w * (scale - 1), dim->y, dim->w * scale, dim->h * scale);
> +        cairo_xlib_surface_set_size(win->root_surface, dim->w * scale, dim->h * scale);
> 
>          XClearWindow(xctx.dpy, win->xwin);
> 
> @@ -171,7 +172,7 @@ void x_display_surface(cairo_surface_t *srf, window winptr, const struct dimensi
>          cairo_show_page(win->c_ctx);
> 
>          if (settings.corner_radius != 0 && ! x_win_composited(win))
> -                x_win_corners_shape(win, dim->corner_radius);
> +                x_win_corners_shape(win, dim->corner_radius * scale);
>          else
>                  x_win_corners_unshape(win);
> ```
> 
> Note the hack around `dim->x` which is not correct in all cases.

Only when scale is less than 1 right? 

> I wonder how it works with Wayland as you never use `dim.x` and `dim.y`.

The wayland implementation uses anchors and specifies the distance from a certain edge of the screen. It calculates that from dim. See https://github.com/dunst-project/dunst/blob/master/src/wayland/wl.c#L724. I'm planning to depreciate the geometry setting in favour of that way of specifying the dimensions.

vincentbernat:
❦  4 July 2021 06:07 -07, fwSmit:

>> Note the hack around `dim->x` which is not correct in all cases.
>
> Only when scale is less than 1 right?

I mean when width is positive or when the width takes more than half the screen.

>> I wonder how it works with Wayland as you never use `dim.x` and `dim.y`.
>
> The wayland implementation uses anchors and specifies the distance
> from a certain edge of the screen. It calculates that from dim. See
> https://github.com/dunst-project/dunst/blob/master/src/wayland/wl.c#L724.
> I'm planning to depreciate the geometry setting in favour of that way
> of specifying the dimensions.

In the meantime, do you want a proper PR or can you just take my diff
and put it in your PR? I think it only needs an additional `if` for
positive width and a safeguard to ensure we don't scale more than the
screen size.

fwSmit:
> ❦ 4 July 2021 06:07 -07, fwSmit:
> > Note the hack around `dim->x` which is not correct in all cases. Only when scale is less than 1 right?

> I mean when width is positive

If you mean with width the x-direction offset of the geometry then yes. Same thing would have to be done for the y-direction. But it's better to scale the width value right before you calculate the window position, then no compensation would have to be done.

> or when the width takes more than half the screen.

You mean that when you scale it, it takes more than the entire width of the screen? I don't think this would be really important. to handle.

> > I wonder how it works with Wayland as you never use `dim.x` and `dim.y`. The wayland implementation uses anchors and specifies the distance from a certain edge of the screen. It calculates that from dim. See https://github.com/dunst-project/dunst/blob/master/src/wayland/wl.c#L724. I'm planning to depreciate the geometry setting in favour of that way of specifying the dimensions.
> In the meantime, do you want a proper PR or can you just take my diff and put it in your PR? I think it only needs an additional `if` for positive width and a safeguard to ensure we don't scale more than the screen size.

A PR would be appreciated! I have one note on my own code: make sure x_get_scale doesn't return 0. Could you fix that as well?

```
int scale = MAX(1, round(scr_info->dpi/96.));
```

PR #885, opened 2021-06-28
wayland: don't warn the user if idle_threshold = 0

fwSmit:


tsipinakis:
Merged as part of settings refactor #803 . This can be closed?

fwSmit:
This PR is slightly different than the commit in #803, but they do the same thing. So I'll close this one.

PR #886, opened 2021-06-30
queues: require the same appid for stacking a notification

fwsmit:
This has been suggested in https://gitlab.freedesktop.org/xdg/xdg-specs/-/issues/77 to prevent name collisions.

### DIFF
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -841,9 +841,9 @@ Updates the icon of the notification, it should be a path to a valid image.
 =item C<set_stack_tag>
 
 Sets the stack tag for the notification, notifications with the same (non-empty)
-stack tag will replace each-other so only the newest one is visible. This can be
-useful for example in volume or brightness notifications where you only want one of
-the same type visible.
+stack tag and the same appid will replace each-other so only the newest one is
+visible. This can be useful for example in volume or brightness notifications
+where you only want one of the same type visible.
 
 The stack tag can be set by the client with the 'synchronous',
 'private-synchronous' 'x-canonical-private-synchronous' or the
--- a/src/queues.c
+++ b/src/queues.c
@@ -247,7 +247,8 @@ static bool queues_stack_by_tag(struct notification *new)
                 for (GList *iter = g_queue_peek_head_link(allqueues[i]); iter;
                             iter = iter->next) {
                         struct notification *old = iter->data;
-                        if (STR_FULL(old->stack_tag) && STR_EQ(old->stack_tag, new->stack_tag)) {
+                        if (STR_FULL(old->stack_tag) && STR_EQ(old->stack_tag, new->stack_tag)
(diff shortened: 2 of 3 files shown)


tsipinakis (approved):

tsipinakis:
Valgrind seems to be complaining. n1 isn't beeing freed for some reason?

fwSmit:
@tsipinakis It was a problem with a `notification_ref` too many. I don't yet understand completely, but it's just a test so if valgrind doesn't complain it's fine. This PR should be ready to merge now

fitrh:
Hi, how can i achieve the same result as below configuration after this commit ? Before this, i can set different state for each low, mid, and high volume but still stacking by using same stack tag, but now it seems same appname required for make it stacking

```dosini
[volume]
    appname = "Audio"
    format = " <span font='32'>%s <span foreground='#a9b1d6'>%b</span>%%</span>"
    foreground = "#0db9d7"
    frame_color = "#0db9d7"
    highlight = "#0db9d7"
    history_ignore = yes
    new_icon = NONE

[volume-high]
    appname = "Audio High"
    format = " <span font='32'>%s <span foreground='#a9b1d6'>%b</span>%%</span>"
    foreground = "#ff7a93"
    frame_color = "#ff7a93"
    highlight = "#ff7a93"
    history_ignore = yes
    new_icon = NONE

[volume-low]
    appname = "Audio Low"
    format = " <span font='32'>%s <span foreground='#a9b1d6'>%b</span>%%</span>"
    foreground = "#a9b1d6"
    frame_color = "#a9b1d6"
    highlight = "#a9b1d6"
    history_ignore = yes
    new_icon = NONE

[volume-mute]
    appname = "Audio Mute"
    format = " <span font='32'>%s %b%%</span>"
    foreground = "#444b6a"
    frame_color = "#444b6a"
    highlight = "#444b6a"
    history_ignore = yes
    new_icon = NONE
```

fwSmit:
> Hi, how can i achieve the same result as below configuration after this commit ? Before this, i can set different state for each low, mid, and high volume but still stacking by using same stack tag, but now it seems same appname required for make it stacking

It's expected that this doesn't work anymore. There are multiple ways to fix this:

- Provide all information for the notification properties via the notification command:
 `dunstify -h string:fgcolor:#444b6a ...`
  You should set the same appname. The markup can only be changed in the body (if this is a problem, make an issue about setting the format via hints).
- Overwrite the notification by id:
  `dunstify -r $ID`
  You can keep the other things you did. Keep in mind this is kind of hacky.

fitrh:
> - Provide all information for the notification properties via the notification command:
>   dunstify -h string:fgcolor:#444b6a ...
Ah, I see, thank you.

OliverLew:
Sorry, I was mistaking the appid and notification id (I thought there was only one id-thing). The behavior did change for me after 1.7.0, but shouldn't be related to this PR. Please forget about that comment.

PR #889, opened 2021-07-02
xdg-open can open URLs, don't rely on specific hardcoded browser.

hramrach:


bebehei (approved):

fwSmit:
PR looks good. I agree that it's better to use xdg-open. Could you add xdg-open as an optional dependency in the readme?

tsipinakis:
Conflicts with `config,h` removal. Otherwise LGTM

Issue #754, opened 2020-08-18
Provide an option to apply Xorg scaling factor to the window geometry?


asrcpq:
I want to have consistent size of dunst notification window when switching between multiple monitors with different resolution and different dpi scaling level. I have a laptop with FHD builtin display and when I connect it to an external UHD monitor(with Xft.dpi doubled) I always find the notification window too narrow.

frebib:
This is still a problem, please re-open this issue. dunst will dynamically redraw when the DPI changes, but the geometry is not updated. Restarting in that case seems overkill, especially when most of the work is already done live

tsipinakis:
Hm, I agree handling this would be useful. There is some dpi handling but it doesn't extend to the window dimensions.

fwSmit:
@frebib @asrcpq  Is anyone still wanting this feature and willing to test it? I could pretty easily implement this (for integer scaling factors), since it's implemented for wayland now already.

frebib:
Yes, I'd very much like this and am willing to run experimental builds. Notifications are at most a luxury for me anyway :)

fwsmit:
This is implemented by #890

PR #890, opened 2021-07-05
X11: enable dpi scaling

vincentbernat:
The scaling factor is derived from the active screen's dpi. This may
or may not match what is expected. Other sources could be font DPI or
Gdk/WindowScalingFactor.

Currently, only integer scaling is supported.

This reuses most of the code that was present for Wayland and the
change is limited to compute a scaling factor and putting the window
at the right position and at the right size. The code to compute the
window position in `draw.c` was only used for X11, so it was moved
directly in `x.c`.

See #884 for the discussion.

I would also be interested to support fractional scaling. From my understanding, this is not possible with Wayland, right? For X11, we can and the change would mostly be replacing `scale` by `scale*10` and divide it again by 10 everywhere.

### DIFF
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -106,6 +106,15 @@ the notification on the left border of the screen while a horizontal offset of
 
 =back
 
+=item B<scale> (default: 0, X11 only)
+
+Specifies a scale factor for dimensions to adapt notifications to
+HiDPI screens. This scales the notification geometry and it's
+contents. It is not recommended to use a fractional scaling factor, as
+this may result in things being one pixel off. Try to use a whole
+number scaling factor and adjust the font size and other sizes as
+needed. If 0 is specified, the scale factor is auto-detected.
+
 =item B<progress_bar> (values: [true/false], default: true)
 
 When an integer value is passed to dunst as a hint (see B<NOTIFY-SEND>), a
--- a/src/draw.c
+++ b/src/draw.c
@@ -774,31 +774,6 @@ static struct dimensions layout_render(cairo_surface_t *srf,
         return dim;
 }
 
-/**
- * Calculates the position the window should be placed at given its width and
- * height and stores them in \p ret_x and \p ret_y.
- */
-static void calc_window_pos(int width, int height, int *ret_x, int *ret_y)
-{
-        const struct screen_info *scr = output->get_active_screen();
-
-        if (ret_x) {
-                if (settings.geometry.negative_x) {
-                        *ret_x = (scr->x + (scr->w - width)) + settings.geometry.x;
-                } else {
-                        *ret_x = scr->x + settings.geometry.x;
-                }
-        }
-
-        if (ret_y) {
-                if (settings.geometry.negative_y) {
-                        *ret_y = scr->y + (scr->h + settings.geometry.y) - height;
-                } else {
-                        *ret_y = scr->y + settings.geometry.y;
-                }
-        }
-}
-
 void draw(void)
 {
         assert(queues_length_displayed() > 0);
@@ -821,7 +796,6 @@ void draw(void)
                 first = false;
         }
 
-        calc_window_pos(dim.w, dim.h, &dim.x, &dim.y);
         output->display_surface(image_surface, win, &dim);
 
         cairo_surface_destroy(image_surface);
--- a/src/draw.c
+++ b/src/draw.c
@@ -144,11 +144,11 @@ static struct color layout_get_sepcolor(struct colored_layout *cl,
 
 static void layout_setup_pango(PangoLayout *layout, int width)
 {
-        int scale = output->get_scale();
+        double scale = output->get_scale();
         pango_layout_set_wrap(layout, PANGO_WRAP_WORD_CHAR);
-        pango_layout_set_width(layout, width * scale * PANGO_SCALE);
+        pango_layout_set_width(layout, round(width * scale * PANGO_SCALE));
         pango_layout_set_font_description(layout, pango_fdesc);
-        pango_layout_set_spacing(layout, settings.line_height * scale * PANGO_SCALE);
+        pango_layout_set_spacing(layout, round(settings.line_height * scale * PANGO_SCALE));
 
         PangoAlignment align;
         switch (settings.align) {
@@ -196,20 +196,20 @@ static bool have_progress_bar(const struct notification *n)
         return (n->progress >= 0 && settings.progress_bar == true);
 }
 
-static void get_text_size(PangoLayout *l, int *w, int *h, int scale) {
+static void get_text_size(PangoLayout *l, int *w, int *h, double scale) {
         pango_layout_get_pixel_size(l, w, h);
         // scale the size down, because it may be rendered at higher DPI
 
         if (w)
-                *w /= scale;
+                *w = round(*w / scale);
         if (h)
-                *h /= scale;
+                *h = round(*h / scale);
 }
 
 static struct dimensions calculate_dimensions(GSList *layouts)
 {
         struct dimensions dim = { 0 };
-        int scale = output->get_scale();
+        double scale = output->get_scale();
 
         const struct screen_info *scr = output->get_active_screen();
         if (have_dynamic_width()) {
@@ -235,7 +235,7 @@ static struct dimensions calculate_dimensions(GSList *layouts)
         int text_width = 0, total_width = 0;
         for (GSList *iter = layouts; iter; iter = iter->next) {
                 struct colored_layout *cl = iter->data;
-                int w=0,h=0;
+                int w=0, h=0;
                 get_text_size(cl->l, &w, &h, scale);
                 if (cl->icon) {
                         h = MAX(get_icon_height(cl->icon, scale), h);
@@ -314,7 +314,7 @@ static struct colored_layout *layout_init_shared(cairo_t *c, const struct notifi
 {
         struct colored_layout *cl = g_malloc(sizeof(struct colored_layout));
         cl->l = layout_create(c);
-        int scale = output->get_scale();
+        double scale = output->get_scale();
 
         if (!settings.word_wrap) {
                 PangoEllipsizeMode ellipsize;
@@ -383,7 +383,7 @@ static struct colored_layout *layout_from_notification(cairo_t *c, struct notifi
 {
 
         struct colored_layout *cl = layout_init_shared(c, n);
-        int scale = output->get_scale();
+        double scale = output->get_scale();
 
         /* markup */
         GError *err = NULL;
@@ -452,7 +452,7 @@ static GSList *create_layouts(cairo_t *c)
 }
 
 
-static int layout_get_height(struct colored_layout *cl, int scale)
+static int layout_get_height(struct colored_layout *cl, double scale)
 {
         int h;
         int h_icon = 0;
@@ -499,15 +499,15 @@ static int frame_internal_radius (int r, int w, int h)
  * The top corners will get rounded by `corner_radius`, if `first` is set.
  * Respectably the same for `last` with the bottom corners.
  */
-void draw_rounded_rect(cairo_t *c, int x, int y, int width, int height, int corner_radius, int scale, bool first, bool last)
+void draw_rounded_rect(cairo_t *c, int x, int y, int width, int height, int corner_radius, double scale, bool first, bool last)
 {
-        width *= scale;
-        height *= scale;
-        x *= scale;
-        y *= scale;
-        corner_radius *= scale;
+        width = round(width * scale);
+        height = round(height * scale);
+        x = round(x * scale);
+        y = round(y * scale);
+        corner_radius = round(corner_radius * scale);
 
-        const float degrees = M_PI / 180.0;
+        const double degrees = M_PI / 180.0;
 
         cairo_new_sub_path(c);
 
@@ -557,8 +557,8 @@ void draw_rounded_rect(cairo_t *c, int x, int y, int width, int height, int corn
 /**
  * A small wrapper around cairo_rectange for drawing a scaled rectangle.
  */
-void draw_rect(cairo_t *c, int x, int y, int width, int height, int scale) {
-        cairo_rectangle(c, x * scale, y * scale, width * scale, height * scale);
+static void draw_rect(cairo_t *c, double x, double y, double width, double height, double scale) {
+        cairo_rectangle(c, round(x * scale), round(y * scale), round(width * scale), round(height * scale));
 }
 
 static cairo_surface_t *render_background(cairo_surface_t *srf,
@@ -571,7 +571,7 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
                                           bool first,
                                           bool last,
                                           int *ret_width,
-                                          int scale)
+                                          double scale)
 {
         int x = 0;
         int radius_int = corner_radius;
@@ -635,10 +635,12 @@ static cairo_surface_t *render_background(cairo_surface_t *srf,
         if (ret_width)
                 *ret_width = width;
 
-        return cairo_surface_create_for_rectangle(srf, x * scale, y * scale, width * scale, height * scale);
+        return cairo_surface_create_for_rectangle(srf,
+                                                  round(x * scale), round(y * scale),
+                                                  round(width * scale), round(height * scale));
 }
 
-static void render_content(cairo_t *c, struct colored_layout *cl, int width, int scale)
+static void render_content(cairo_t *c, struct colored_layout *cl, int width, double scale)
 {
         const int h = layout_get_height(cl, scale);
         int h_without_progress_bar = h;
@@ -667,7 +669,7 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, int
                         text_x = get_icon_width(cl->icon, scale) + settings.h_padding + get_text_icon_padding();
                 } // else ICON_RIGHT
         }
-        cairo_move_to(c, text_x * scale, text_y * scale);
+        cairo_move_to(c, round(text_x * scale), round(text_y * scale));
 
         cairo_set_source_rgba(c, cl->fg.r, cl->fg.g, cl->fg.b, cl->fg.a);
         pango_cairo_update_layout(c, cl->l);
@@ -695,7 +697,7 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, int
                         image_x = settings.h_padding;
                 } // else ICON_RIGHT
 
-                cairo_set_source_surface(c, cl->icon, image_x * scale, image_y * scale);
+                cairo_set_source_surface(c, cl->icon, round(image_x * scale), round(image_y * scale));
                 draw_rect(c, image_x, image_y, image_width, image_height, scale);
                 cairo_fill(c);
         }
@@ -728,8 +730,14 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, int
                 cairo_fill(c);
                 // border
                 cairo_set_source_rgba(c, cl->frame.r, cl->frame.g, cl->frame.b, cl->frame.a);
-                // TODO draw_rect instead of cairo_rectangle resulted in blurry lines. Why?
-                cairo_rectangle(c, (frame_x + half_frame_width) * scale, (frame_y + half_frame_width) * scale, (progress_width - frame_width) * scale, progress_height * scale);
+                // TODO draw_rect instead of cairo_rectangle resulted
+                // in blurry lines due to rounding (half_frame_width
+                // can be non-integer)
+                cairo_rectangle(c,
+                                (frame_x + half_frame_width) * scale,
+                                (frame_y + half_frame_width) * scale,
+                                (progress_width - frame_width) * scale,
+                                progress_height * scale);
                 cairo_set_line_width(c, frame_width * scale);
                 cairo_stroke(c);
         }
@@ -742,7 +750,7 @@ static struct dimensions layout_render(cairo_surface_t *srf,
                                        bool first,
                                        bool last)
 {
-        int scale = output->get_scale();
+        double scale = output->get_scale();
         const int cl_h = layout_get_height(cl, scale);
 
         int h_text = 0;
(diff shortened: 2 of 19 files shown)


frebib:
This seems to work on my computer, where I presume it's intended that the current DPI is rounded up to the nearest integer scale which is used to calculate the geometry. I tested 96dpi (1.0x), 144dpi (1.5x) and 192dpi (2.0x) where I got the geometry scale 1x, 2x, 2x respectively.

I'll chime in and say whilst this is better than no scaling at all, fractional scaling would be a lot more useful for the majority of people (myself included), so hopefully this improvement can lead to that next! 
> For X11, we can and the change would mostly be replacing scale by scale*10 and divide it again by 10 everywhere.

I'm not sure this would work great to be honest, see cases like 1.25x (120dpi), it would be slightly off. I think it'd be better to pass around the full scale value as a float and clamp/round it for wayland but not for X11.

Either way, this is a great improvement and a good start to solving #754 fully. Thanks for the hard work @vincentbernat and @fwSmit

fwSmit:
Thank you for the PR

> The scaling factor is derived from the active screen's dpi. This may
> or may not match what is expected. Other sources could be font DPI or Gdk/WindowScalingFactor.

That could be an option. It's mostly that the screen dpi detection was already there. I believe advantages of screen dpi include (correct me if I'm wrong)
- It can be updated on the fly
- It's built into an (universal) X11 extension
- It is automatically assigned (???)

I don't know about the other options, but I'm open to change it. Depending on which one we choose, an option in the dunst settings to overwrite the scale value may be useful.

> Currently, only integer scaling is supported.
> 
> This reuses most of the code that was present for Wayland and the
> change is limited to compute a scaling factor and putting the window
> at the right position and at the right size. The code to compute the
> window position in `draw.c` was only used for X11, so it was moved
> directly in `x.c`.

Yes seems good to me.

> I would also be interested to support fractional scaling. From my understanding, this is not possible with Wayland, right? For X11, we can and the change would mostly be replacing `scale` by `scale*10` and divide it again by 10 everywhere.

Why not change `scale` to a float? I'm trying to think of disadvantages. I can come up with
- Sizes may not end up at whole pixels. You can either round, making thinks potentially look not aligned. Or you can alias, which results in smooth edges.

The advantage of the `scale*10` way is that when the scale is a multiple 10, everything will be exact, where as with a float the scale of 1 may still result in distortion (depending on if you round).

fwSmit:
To explain the Wayland situation a bit more: the scale factor can only be an integer. The nearest integer above the scale factor is chosen, and then the image is scaled down by interpolation. I understand that non-integer scaling is very useful and also very possible for dunst, so let's look at the best way to implement it.

src/x11/x.c
```diff
@@ -161,8 +161,24 @@ static bool x_win_composited(struct window_x11 *win)
 void x_display_surface(cairo_surface_t *srf, window winptr, const struct dimensions *dim)
 {
```
fwSmit, line 162: Just a note here. dim->x and dim->y aren't used anywhere anymore. This is just temporary though, because after #855 the window position code will be shared again.

src/x11/x.c
```diff
@@ -161,8 +161,24 @@ static bool x_win_composited(struct window_x11 *win)
...
         struct window_x11 *win = (struct window_x11*)winptr;
-        x_win_move(win, dim->x, dim->y, dim->w, dim->h);
-        cairo_xlib_surface_set_size(win->root_surface, dim->w, dim->h);
+        const struct screen_info *scr = get_active_screen();
+        int scale = x_get_scale();
```
fwSmit, line 165: Could you pass the scale as a variable to the function instead? It is known in `void draw()` already

fwsmit (changes requested):
Looks good overall just a small change needed. If you want you can implement instant changing when the DPI changes by calling `wake_up()` when a dpi change event is emitted (note that there's a bug that icons still don't scale instantly).

vincentbernat:
❦  6 July 2021 15:29 -07, fwSmit:

>> The scaling factor is derived from the active screen's dpi. This may
>> or may not match what is expected. Other sources could be font DPI or
>> Gdk/WindowScalingFactor.
>
> That could be an option. It's mostly that the screen dpi detection was
> already there. I believe advantages of screen dpi include (correct me
> if I'm wrong)
> - It can be updated on the fly
> - It's built into an (universal) X11 extension
> - It is automatically assigned (???)
>
> I don't know about the other options, but I'm open to change it.
> Depending on which one we choose, an option in the dunst settings to
> overwrite the scale value may be useful.

I just mention this to document it somewhere. GTK applications are not
reacting directy to XRandR DPI, the watch for XSETTINGS changes, but
watching XRandR DPI is not unknown territory as Chromium does that too.
I don't think we need to do something now. If the issue popup, you can
search for it and find this PR ;-)

>> I would also be interested to support fractional scaling. From my
>> understanding, this is not possible with Wayland, right? For X11, we
>> can and the change would mostly be replacing `scale` by `scale*10`
>> and divide it again by 10 everywhere.
>
> Why not change `scale` to a float? I'm trying to think of disadvantages. I can come up with
> - Sizes may not end up at whole pixels. You can either round, making
> thinks potentially look not aligned. Or you can alias, which results
> in smooth edges.
>
> The advantage of the `scale*10` way is that when the scale is a
> multiple 10, everything will be exact, where as with a float the scale
> of 1 may still result in distortion (depending on if you round).

Yes, that was the main reason. But we can also round everything and it
should be fine. As it would handle 1.25x case better, I'll do that.

src/x11/x.c
vincentbernat: I was going to remove it from struct, but some other parts of the code are still using it.

src/x11/x.c
vincentbernat: It's also called from `x_win_show()`.

fwSmit: Oh it's fine then.

fwSmit:
> Yes, that was the main reason. But we can also round everything and it should be fine. As it would handle 1.25x case better, I'll do that.

Ok you can go ahead and make `scale` a float. Let's see what happens.

vincentbernat:
Done. Tested on X11 with 1, 1.25, 1.5 and 2. No issue detected! This is quite a big change due to the need to add rounding everywhere. This is not strictly necessary but it may result as an off-by-one error if we don't do that. Maybe we could live with that if you want to reduce the changeset.

fwSmit:
I've taken a quick look and it looks good. If I understand it correctly the `+ 0.5` is because C rounds toward zero, so yes that can stay.
Could you add a setting for overriding the scale? It should also work for wayland.

frebib:
Just tested this and it seems to work a treat on X11 with 1.0x and 1.5x scale! Thanks so much

src/wayland/wl.c
```diff
@@ -635,7 +635,7 @@ static void schedule_frame_and_commit();
 
 // Draw and commit a new frame.
 static void send_frame() {
-        int scale = wl_get_scale();
+        int scale = wl_get_scale() + 0.5;
```
fwSmit, line 638: This `+0.5` is doing nothing at the moment. `wl_get_scale` only returns integers now (converted to float). I would either implement a floating point scale for wayland as well (by changing a few more things to float and adding +0.5's), or changing `wl_get_scale` to return an int.

fwsmit (commented):
For me the notification became blurry from this patch and it looked like it was (half a) pixel below where it should be.

Before
20210708_22h06m46s_grim

After
20210708_22h06m36s_grim

Note also that a separator seems  to appear in the bottom image, but I have the color set to my background color. It will probably be fixed when you address the comments

src/wayland/wl.c
vincentbernat: The signature needs it to return a float because it is called as `output->get_scale()` in `draw.c`.

fwSmit: Right. Still, the `+ 0.5` isn't needed here at the moment, so please remove them here where possible

src/draw.c
```diff
@@ -554,8 +554,8 @@ void draw_rounded_rect(cairo_t *c, int x, int y, int width, int height, int corn
 /**
  * A small wrapper around cairo_rectange for drawing a scaled rectangle.
  */
-void draw_rect(cairo_t *c, int x, int y, int width, int height, int scale) {
-        cairo_rectangle(c, x * scale, y * scale, width * scale, height * scale);
+void draw_rect(cairo_t *c, int x, int y, int width, int height, float scale) {
+        cairo_rectangle(c, x * scale + 0.5, y * scale + 0.5, width * scale + 0.5, height * scale + 0.5);
```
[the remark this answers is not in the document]
vincentbernat: Ack. Let's just put `round()` everywhere instead of mixing.

src/wayland/wl.c
vincentbernat: I did remove them.

frebib:
I can confirm that I also noticed the text aliasing at 1.0x scale at closer inspection with 
66a7efa, but it appears to be back to "normal" with the latest revision (7048e74) and the scaling seems to still work as expected 👍🏻

src/x11/x.c
```diff
@@ -942,8 +958,12 @@ static void x_shortcut_init(struct keyboard_shortcut *ks)
...
 }
 
-int x_get_scale(void) {
-        return 1;
+float x_get_scale(void) {
+        const struct screen_info *scr_info = get_active_screen();
+        float scale = MAX(1, scr_info->dpi/96.);
+        LOG_D("X11 dpi: %i", scr_info->dpi);
+        LOG_D("X11 scale: %f", scale);
+        return scale;
```
fwSmit, line 966: Could you add a scale setting that overrides this and wl_get_scale?

vincentbernat: Done.

docs/dunst.5.pod
```diff
@@ -100,6 +100,12 @@ the notification on the left border of the screen while a horizontal offset of
 
 =back
 
+=item B<scale> (default: 0)
+
+Specifies a scale factor for dimensions to adapt notifications to the
+screen DPI. This does not apply to text. If 0 is specified, the scale
```
fwSmit, line 106: Screen dpi or wayland scale factor

fwSmit:
I will test it later today, but other then the little documentation thing it looks very good

docs/dunst.5.pod
vincentbernat: That's the source of auto-detection, not really the purpose. I have pushed another tentative to not mention DPI. Otherwise, please propose the whole text.

fwSmit: It's documentation for the dpi scaling at the same time :)
The text should also be scaled with the scale factor. If that's not the case then something is wrong.

I would suggest this as text:
```
Specifies a scale factor for dimensions to adapt notifications to HiDPI
screens. This scales the notification geometry and it's contents.
It is not recommended to use a fractional scaling factor, as this may result in
things being one pixel off. Try to use a whole number scaling factor and adjust
the font size and other sizes as needed. If 0 is specified, the scale factor is
auto-detected.
```

fwsmit (commented):
The fractional scaling does work, but there are a couple of artifacts introduced by the rounding. It's fine to have those for a first effort, but I would like to keep the default to not use fractional scaling. So could you make x_get_scale only return rounded scale factors?

docs/dunst.5.pod
vincentbernat: Unfortunately, on X11, it does not scale text. Text is handled using Xft/DPI (from XSETTINGS, with a fallback to X resource database). I don't think this is very important, unless you use the `scale` settings, in this case, you don't control the whole scale.

vincentbernat:
What kind of artifacts? Having to force a fractional scaling through settings only means dunst has to be restarted when the DPI change (for example, when you connect external screens). Maybe I can add -1 to autodetect *and* keep fractional scaling?

vincentbernat:
If the artifacts are due to the scaling factor using non fractional values, like 1.04 because user did not set Xft.DPI (which may be possible if you don't care about HiDPI), maybe we could also whitelist 1.5 and 1.25.

Linux is kind of a hell for people using HiDPI displays. The general mechanism is to get text size from Xft/DPI and interface scale from Gdk/WindowScalingFactor (which is not fractional) and to listen for changes on XSETTINGS. However, many apps fallback to various mechanisms, like monitor DPI settings or more usually Xft.DPI. So, the best we can do when using HiDPI display is to set XRandR DPI, Xft.DPI, Xft/DPI, and Gdk/WindowScalingFactor to consistent values. Most apps then just work fine or don't handle DPI at all. Having to mangle each app is cumbersome. That's why I would like to default to something working out of the box.

fwSmit:
> What kind of artifacts? Having to force a fractional scaling through settings only means dunst has to be restarted when the DPI change (for example, when you connect external screens). Maybe I can add -1 to autodetect _and_ keep fractional scaling?

I've noticed two things:
The border is visible at the place of the separator. In the below pic I set the border to green and the separator to red to show the effect.

scale: 1.25x
20210711_20h05m38s_grim

Progress bar border isn't placed correctly. Note the space at the left:

scale: 1x
20210711_20h08m34s_grim

The second one is a regression, because the border was placed in a non-whole pixel pixel position. Cairo strokes are specified by the center of the stroke, so that's why a non-whole pixel position was needed.
Note that the border of the progress bar is drawn differently than the border of the notification. The progress bar border is drawn with a stroke and the other one is just a rounded rectangle. 

All in all, it's a little more complicated than rounding everything. I'm currently feeling like we will be better off not rounding and letting anti-aliassing do it's work, but I'm curious what you think

> If the artifacts are due to the scaling factor using non fractional values, like 1.04 because user did not set Xft.DPI (which may be possible if you don't care about HiDPI), maybe we could also whitelist 1.5 and 1.25.

I don't think it's a problem on 1.04 scale, but it is on 1.25 as seen above. Whitelisting certain values doesn't seem like a good solution to me. I think it's better to improve the scaling just a bit more.

> Linux is kind of a hell for people using HiDPI displays. The general mechanism is to get text size from Xft/DPI and interface scale from Gdk/WindowScalingFactor (which is not fractional) and to listen for changes on XSETTINGS. However, many apps fallback to various mechanisms, like monitor DPI settings or more usually Xft.DPI. So, the best we can do when using HiDPI display is to set XRandR DPI, Xft.DPI, Xft/DPI, and Gdk/WindowScalingFactor to consistent values. Most apps then just work fine or don't handle DPI at all. Having to mangle each app is cumbersome. That's why I would like to default to something working out of the box.

Agreed, I'm trying to make dunst as good out of the box as possible. One should need to do no configuration to make it work well.

docs/dunst.5.pod
fwSmit: Weird I thought we controlled text size directly. The scale variable is kind of useless on X11 then. On wayland it also doesn't work currently. It doesn't scale the window dimensions, so the notification is cut off.

vincentbernat:
I am not familiar with Cairo at all, so I don't know what other software do. It's a bit odd the regression at 1x, since nothing has really changed. We multiply by 1 or divide by 1. Maybe using a float instead of a double is the reason? Cairo is using double. Conversion from float to double may introduce this kind of things?

As for fractional scaling, that's mostly why Gnome was really reluctant to support it for anything else than text. But if you have a HiDPI screen at 1.5x scale, not scaling the interface does not give a very good result. That's why I think recent versions of Gnome started supporting fractional scaling.

I think that whitelisting known ratio (1.25 and 1.5) by default and using integer ratio otherwise would be a good compromise. Setup matching one of those two ratios have most likely configured their environment for it (by opposition to people using 1x scale with no Xft.DPI set and where dunst is using computed DPI from XRandR) and are willing to live with the little imperfections. This would make the setup work out of the box for most people in the way they like it. As long as we find the regression at 1x with Cairo.

docs/dunst.5.pod
vincentbernat: We could make it an X11 only setting. Wayland does not really need it because there is only one way to get the scaling, so we are unlikely to be wrong. And since it does not scale text, we can just have a 3-state setting: integer-scaling, integer-scaling++, fractional-scaling. I am not good with names.

fwSmit:
We could round the scale to multiples of 0.25. But again, I haven't tested what happens at scales like 1.04, so let's test that first.

I've explained the regression for 1x above already. It has to do with the line needing to be on a non-whole pixel for certain line widths. If the line width is 3 for example, you have to specify the location of the line by it's middle. So if the left of the line has to be on pixel 40, the middle is on pixel 41.5.

docs/dunst.5.pod
fwSmit: Yeah but if the text size doesn't change from this scale variable it's pretty much useless. Is there a way to draw larger text independent from the X dpi?

vincentbernat: I don't know. But if there is, I wouldn't expect Dunst to override Xft/DPI (or Xft.DPI). Text scaling is pretty much the only thing that's easy to do on Linux without much surprise.

vincentbernat:
What I don't understand is that it shouldn't be a regression as we didn't change that. We still pass the same values as before to Cairo, except instead of being int, they are floats (but rounded). From what I understand, the problem should already be present in master?

src/draw.c
```diff
@@ -725,9 +727,11 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, int
                 cairo_fill(c);
                 // border
                 cairo_set_source_rgba(c, cl->frame.r, cl->frame.g, cl->frame.b, cl->frame.a);
-                // TODO draw_rect instead of cairo_rectangle resulted in blurry lines. Why?
-                cairo_rectangle(c, (frame_x + half_frame_width) * scale, (frame_y + half_frame_width) * scale, (progress_width - frame_width) * scale, progress_height * scale);
-                cairo_set_line_width(c, frame_width * scale);
+                draw_rect(c,
```
fwSmit, line 730: The regression is here. Before it was using floats directly to cairo_rectangle

vincentbernat: Oh, good catch! Should we convert draw_rect to floats?

fwSmit:
Just a heads up, I'll be gone for a few weeks, so don't expect this to be merged soon

src/draw.c
vincentbernat: I have converted it to double, but this is not enough because of rounding.

Unrelated, but I also noticed I have used `float` while most of the existing code and Cairo are using `double`. So, I have converted everything to `double` (and use of `round()` introduced a conversion to `double` as well).

vincentbernat:
I have also discovered the functional tests. The following ones are failing:
* `-300x`
* `-300x1-20-20`
* `x1` (which is also mislabelled, I'll push a fix very shortly)

I think we should fix it, but you said previously this didn't matter. Did I understand correctly?

src/draw.c
fwSmit: > I have converted it to double, but this is not enough because of rounding.

Yeah it's better then to leave the old cairo_rectangle line that was there. You can go ahead and leave the TODO out though.

> Unrelated, but I also noticed I have used `float` while most of the existing code and Cairo are using `double`. So, I have converted everything to `double` (and use of `round()` introduced a conversion to `double` as well).

:+1:

fwSmit:
> I have also discovered the functional tests. The following ones are failing:
> 
>     * `-300x`
> 
>     * `-300x1-20-20`
> 
>     * `x1` (which is also mislabelled, I'll push a fix very shortly)
> 
> 
> I think we should fix it, but you said previously this didn't matter. Did I understand correctly?

I can't remember saying anything about the functional tests, but I think it's worth it to fix them. They're quite outdated, so a lot of the settings have to be updated. The functional tests are useful, since they test things that can't really be tested with automated tests.

fwSmit:
Now that #803 is merged this PR has merge conflicts. You'll need to update the added setting to the new format. Let me know if you need help with that.

fwSmit:
@vincentbernat Are you able to rebase the PR and fix the last few things?

vincentbernat:
Sorry, I am a bit low on time currently. Dunno exactly when I will get more time.

fwSmit:
Okay, no problem. Maybe someone else wants to finish it, or I can do it some time.

vincentbernat:
Let me do it quickly.

vincentbernat:
OK, it's rebased.

vincentbernat:
Hold on, it's crashing, I didn't test correctly...

vincentbernat:
OK, all good for me.

vincentbernat:
And I have added tests too.

fwsmit (commented):
Thank you for the quick reply and fix! It works very well for me now. There's two things remaining as far as I can tell:
- The scale variable doesn't change the window size in wayland. I would suggest making `scale` a X11 only option, since sway on wayland has a better way to set scale anyways.
- The text size on X11 doesn't change with a change in `scale`. Is that expected behaviour? It's not a dealbreaker for me, so it can be merged with this quirk.

vincentbernat:
❦ 10 August 2021 03:01 -07, fwSmit:

> - The scale variable doesn't change the window size in wayland. I
> would suggest making `scale` a X11 only option, since sway on wayland
> has a better way to set scale anyways.

Is it only a matter of adding a "(X11 only)" in the description?

> - The text size on X11 doesn't change with a change in `scale`. Is
> that expected behaviour? It's not a dealbreaker for me, so it can be
> merged with this quirk.

Yes, this is expected. We discussed a bit about this, and that's how
scaling generally work for X11. I don't think it will surprise most
people. And people should just use `scale=0` since it will match the
size of the text.

fwSmit:
> Is it only a matter of adding a "(X11 only)" in the description?

Yes and removing the `setting.scale` from wl_get_scale().

> Yes, this is expected. We discussed a bit about this, and that's how scaling generally work for X11. I don't think it will surprise most people. And people should just use `scale=0` since it will match the size of the text.

Okay that's fine by me. If people get confused we can always add a note in the documentation

vincentbernat:
OK, updated.

docs/dunst.5.pod
```diff
@@ -106,6 +106,15 @@ the notification on the left border of the screen while a horizontal offset of
 
 =back
 
+=item B<scale> (default: 0)
```
fwSmit, line 109: Could you add `(X11 only)` here as well?

fwsmit (approved):
Thank you, I only have one small comment

docs/dunst.5.pod
vincentbernat: Done.

fwSmit:
Thank you for your work!

svenstaro:
A release with this would be great! :)

fwSmit:
I agree a release is due. What do you think @tsipinakis?

PR #895, opened 2021-08-11
Fix history_ignore and match_transient options

Suyooo:
Hi everyone,

this pull request fixes two small problems I noticed with my config, that seem to have snuck in with the recent settings refactor.

The first one is that the `history_ignore` setting was not being read (showing a warning about the setting not existing), even though the implementation was still there. Checking `settings_data.h`, there seemed to be a `frame` option that was simply a duplicate of `skip_display`.

I was unsure whether there was a point to it, and assumed it was a typo, so I replaced it with the definition for the `history_ignore` option, and the setting works fine again. (I also took the opportunity to add the option to the docs.)

The second one is that the `match_transient` filter was not working correctly. This seems to be caused by `string_parse_bool` only writing a `bool` even though the setting is stored as an `int` in the struct, resulting in a problem since it's initialized with `-1`. [Here's a small program to demonstrate it.](https://gist.github.com/Suyooo/d92cf518ee42f9df76335ed8ae75c348)

To fix this, I simply replaced the parser with `string_parse_enum`, which writes the full `int` instead of being truncated when passing through `string_parse_bool`. I did the same for other, similar settings - options where a boolean value is defined, but stored in an `int` - for consistency.

If you feel like there are better approaches available to fix these, please let me know and I'll edit the PR.


fwSmit (approved):
Thank you for the PR, it looks perfect. Just to explain your questions:

> The first one is that the `history_ignore` setting was not being read (showing a warning about the setting not existing), even though the implementation was still there. Checking `settings_data.h`, there seemed to be a `frame` option that was simply a duplicate of `skip_display`.
> 
> I was unsure whether there was a point to it, and assumed it was a typo, so I replaced it with the definition for the `history_ignore` option, and the setting works fine again. (I also took the opportunity to add the option to the docs.)

This was indeed just a typo.

> 
> The second one is that the `match_transient` filter was not working correctly. This seems to be caused by `string_parse_bool` only writing a `bool` even though the setting is stored as an `int` in the struct, resulting in a problem since it's initialized with `-1`. [Here's a small program to demonstrate it.](https://gist.github.com/Suyooo/d92cf518ee42f9df76335ed8ae75c348)
> 
> To fix this, I simply replaced the parser with `string_parse_enum`, which writes the full `int` instead of being truncated when passing through `string_parse_bool`. I did the same for other, similar settings - options where a boolean value is defined, but stored in an `int` - for consistency.

The `string_parse_bool` function only works for values that are actually stored in a boolean type as you demonstrated. Thus it's good that you replaced it where neccesary. (But if you use `string_parse_int` for boolean types it will overwrite the next variable as well) I've checked and it seems that `string_parse_bool` is only used for boolean types now.

PR #903, opened 2021-09-10
Set cursor to left_ptr when mouse is over a notification on wayland

GorrillaRibs:
As in mako here: https://github.com/emersion/mako/commit/85a3e38e9334cdd31dd43802b6850255986a72f2

Currently (on wayland) the cursor remains what it was when the mouse enters the notification window (i.e. from a terminal or a text entry box it'll be a vertical bar), this fixes that.

Code is taken from [mako](https://github.com/emersion/mako) (which used to have the same issue) and adapted a little to fit dunst.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -56,4 +56,5 @@ pkg_config_packs += libnotify
 
 ifneq (0,${WAYLAND})
 pkg_config_packs += wayland-client
+pkg_config_packs += wayland-cursor
 endif
--- a/src/wayland/wl.c
+++ b/src/wayland/wl.c
@@ -7,6 +7,7 @@
 #include <stdbool.h>
 #include <wayland-client.h>
 #include <wayland-client-protocol.h>
+#include <wayland-cursor.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <sys/mman.h>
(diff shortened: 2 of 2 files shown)


config.mk
```diff
@@ -56,4 +56,5 @@ pkg_config_packs += libnotify
 
 ifneq (0,${WAYLAND})
 pkg_config_packs += wayland-client
+pkg_config_packs +=wayland-cursor
```
fwSmit, line 59:
```suggestion
pkg_config_packs += wayland-cursor
```

src/wayland/wl.c
```diff
@@ -70,17 +71,20 @@ struct wl_ctx {
         } pointer;
 
 	struct {
-		struct wl_touch *wl_touch;
-		struct {
-			int32_t x, y;
-		} pts[MAX_TOUCHPOINTS];
-	} touch;
+                struct wl_touch *wl_touch;
```
fwSmit, line 74: Please change the tabs to spaces. In the rest of the code as well

src/wayland/wl.c
```diff
@@ -579,6 +590,40 @@ bool wl_init() {
                 LOG_W("compositor doesn't support zwlr_foreign_toplevel_v1. Fullscreen detection won't work");
         }
 
+	// Set up the cursor. It needs a wl_surface with the cursor loaded into it.
+	// If one of these fail, mako will work fine without the cursor being able to change.
```
fwSmit, line 594:
```suggestion
	// If one of these fail, dunst will work fine without the cursor being able to change.
```

src/wayland/wl.h
```diff
@@ -28,5 +28,6 @@ bool wl_have_fullscreen_window(void);
 // should be multiplied by this value, but don't use it to multiply other
 // values. All sizes should be in unscaled units.
 double wl_get_scale(void);
+
```
fwSmit, line 31: Could you remove this unrelated change?

fwsmit (changes requested):
Thanks for porting this patch. I've got a few style comments, but other than that it looks good

src/wayland/wl.c
GorrillaRibs: Done (I think - tried to do it automatically w/ nvim and it looks ok, but probably worth double checking)

src/wayland/wl.h
GorrillaRibs: Done!

GorrillaRibs:
I think I've fixed everything brought up, but I haven't really used pull requests on github before, so I might've missed something - please let me know if there's any issues!

fwSmit:
@GorrillaRibs Looks great! If you could squash the commits into the first one I'll merge the PR.

fwsmit (approved):

GorrillaRibs:
Done!

fwSmit:
Thank you :)

PR #908, opened 2021-09-11
Better default installation locations for systemd and dbus units

### LINKED ISSUE #907: Don't use pkg-config for SERVICEDIR_*
### Installation info

- Version: Dunst - A customizable and lightweight notification-daemon v1.6.1-107-g3055efd
- Install type: manually
- Distro and version:  Ubuntu 20.04.3

Expected behaviour:
Software installed under $PREFIX (/usr/local)

Actual behaviour:
Systemd and dbus units overwrite distribution versions in /usr/share.

`debsums dunst`
> /usr/bin/dunst                                                                OK
> /usr/lib/systemd/user/dunst.service                                       FAILED
> /usr/share/dbus-1/services/org.knopwob.dunst.service                      FAILED
> /usr/share/doc/dunst/NEWS.gz                                                  OK
> /usr/share/doc/dunst/changelog.Debian.gz                                      OK
> /usr/share/doc/dunst/copyright                                                OK
> /usr/share/doc/dunst/dunstrc.gz                                               OK
> /usr/share/man/man1/dunst.1.gz                                                OK

Please don't use pkg-config to determine where to put the systemd and dbus service units. As the man pages of systemd/dbus-daemon clearly state: those directories (returned by pkg-config) are for *packages* of the distro. Locally installed software should always go into /usr/local or whatever PREFIX the builder fancies.

I just wanted to check the latest git version because I noticed some warnings with the dunst version of Ubuntu (1.4.1) under (X)wayland. I never uninstall distro packages just to test an upstream version. One cannot easily uninstall the dunst package anyway since it is depended on by update-notifier in Ubuntu, for instance. So users who install without thinking about this might in the long run end up with two inconsistent versions of dunst, the distro version and the local one, because the package manager will happily overwrite files in /usr/share on update/upgrade as will make install.

Long story short: **Never *ever* install into directories managed by the distro package manager**, unless you are very certain of what you are doing and asking for in the process. ;) (I have never had the need for it, that is precisely why there is /usr/local which always comes first in any [sane] search path). Many users have no clue about this which clearly rules out the current procedure as default. I did read the build instructions and I do know about SERVICEDIR_SYSTEMD and SERVICEDIR_DBUS but nobody should need to use those parameters unless they have some non-default setup. I am building quite a lot of software from source and very seldomly need to mess with the default build configuration. This should not be one of those occasions. ;)

Now, here is what I think the best way of doing this is:
- Install the systemd user service in /usr/local/lib/systemd/user or maybe /etc/systemd/user (see Table 2 in systemd.unit(5)). I think the 1st directory makes it easy enough to just install to $(PREFIX)/lib/systemd/user.
- Install the dbus unit to $(PREFIX)/share/dbus-1/services or maybe query $XDG_DATA_DIRS, but I think the latter only matters for users that have altered the default. Those should know to take care of where these files go by themselves.

Please, please, *please* change this. If you don't, at least warn other, less observant, users about this, and loudly so.

Cheers,
Peter

WhitePeter:
This PR proposes better default locations, so locally installed dunst versions can coexist with distro version, as it should be. This fixes #907.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,30 +8,7 @@ ifneq ($(wildcard ./.git/),)
 VERSION := $(shell ${GIT} describe --tags)
 endif
 
-ifeq (,${SYSTEMD})
-# Check for systemctl to avoid discrepancies on systems, where
-# systemd is installed, but systemd.pc is in another package
-systemctl := $(shell command -v ${SYSTEMCTL} >/dev/null && echo systemctl)
-ifeq (systemctl,${systemctl})
-SYSTEMD := 1
-else
-SYSTEMD := 0
-endif
-endif
-
-SERVICEDIR_DBUS ?= $(shell $(PKG_CONFIG) dbus-1 --variable=session_bus_services_dir)
-SERVICEDIR_DBUS := ${SERVICEDIR_DBUS}
-ifeq (,${SERVICEDIR_DBUS})
-$(error "Failed to query $(PKG_CONFIG) for package 'dbus-1'!")
-endif
-
-ifneq (0,${SYSTEMD})
-SERVICEDIR_SYSTEMD ?= $(shell $(PKG_CONFIG) systemd --variable=systemduserunitdir)
-SERVICEDIR_SYSTEMD := ${SERVICEDIR_SYSTEMD}
-ifeq (,${SERVICEDIR_SYSTEMD})
-$(error "Failed to query $(PKG_CONFIG) for package 'systemd'!")
-endif
-endif
+SYSTEMD ?= $(shell $(PKG_CONFIG) --silence-errors ${SYSTEMDAEMON} || echo 0)
 
 ifneq (0,${WAYLAND})
 DATA_DIR_WAYLAND_PROTOCOLS ?= $(shell $(PKG_CONFIG) wayland-protocols --variable=pkgdatadir)
--- a/README.md
+++ b/README.md
@@ -107,10 +107,10 @@ sudo make install
 - `BINDIR=<PATH>`: Set the `dunst` executable's path (Default: `${PREFIX}/bin`)
 - `DATADIR=<PATH>`: Set the path for shared files. (Default: `${PREFIX}/share`)
 - `MANDIR=<PATH>`: Set the prefix of the manpage. (Default: `${DATADIR}/man`)
-- `SYSTEMD=(0|1)`: Disable/Enable the systemd unit. (Default: detected via `pkg-config`)
+- `SYSTEMD=(0|1)`: Disable/Enable the systemd unit. (Default: autodetect systemd)
 - `WAYLAND=(0|1)`: Disable/Enable wayland support. (Default: 1 (enabled))
-- `SERVICEDIR_SYSTEMD=<PATH>`: The path to put the systemd user service file. Unused, if `SYSTEMD=0`. (Default: detected via `pkg-config`)
-- `SERVICEDIR_DBUS=<PATH>`: The path to put the dbus service file. (Default: detected via `pkg-config`)
+- `SERVICEDIR_SYSTEMD=<PATH>`: The path to put the systemd user service file. Unused, if `SYSTEMD=0`. (Default: `${PREFIX}/lib/systemd/user`)
+- `SERVICEDIR_DBUS=<PATH>`: The path to put the dbus service file. (Default: `${DATADIR}/dbus-1/services`)
 - `EXTRA_CFLAGS=<FLAGS>`: Additional flags for the compiler.
 
 **Make sure to run all make calls with the same parameter set. So when building with `make PREFIX=/usr`, you have to install it with `make PREFIX=/usr install`, too.**
--- a/config.mk
+++ b/config.mk
@@ -6,6 +6,8 @@ DATADIR ?= ${PREFIX}/share
 # around for backwards compatibility
 MANPREFIX ?= ${DATADIR}/man
 MANDIR ?= ${MANPREFIX}
+SERVICEDIR_DBUS ?= ${DATADIR}/dbus-1/services
+SERVICEDIR_SYSTEMD ?= ${PREFIX}/lib/systemd/user
 EXTRA_CFLAGS ?=
 
 DOXYGEN ?= doxygen
@@ -15,7 +17,7 @@ GIT ?= git
 PKG_CONFIG ?= pkg-config
 POD2MAN ?= pod2man
 SED ?= sed
-SYSTEMCTL ?= systemctl
+SYSTEMDAEMON ?= systemd
 VALGRIND ?= valgrind
 
 # Disable systemd service file installation,


Makefile
```diff
@@ -8,30 +8,7 @@ ifneq ($(wildcard ./.git/),)
...
-endif
-
-ifneq (0,${SYSTEMD})
-SERVICEDIR_SYSTEMD ?= $(shell $(PKG_CONFIG) systemd --variable=systemduserunitdir)
-SERVICEDIR_SYSTEMD := ${SERVICEDIR_SYSTEMD}
-ifeq (,${SERVICEDIR_SYSTEMD})
-$(error "Failed to query $(PKG_CONFIG) for package 'systemd'!")
-endif
-endif
+SYSTEMD ?= $(shell $(SYSTEMDAEMON) --version >/dev/null || echo 0)
```
fwSmit, line 11: On my system (Arch) the binary `systemd` is not in my path. I would have to run `/usr/lib/systemd/systemd` for it to work. I'm not sure what a portable way to check it is. You would probably have to check at least `/usr/lib/systemd/systemd and `systemd`.
See also: https://bugs.archlinux.org/task/35195

WhitePeter: Thanks for pointing that out. I just checked and found that '/usr/bin/systemd' is symlinked to '/lib/systemd/systemd' on Ubuntu, so the real executable is not in PATH either. I think the change I just added resolves this.

fwSmit:
Looks good to me! If you could squash the whitespace fix into the first commit and the systemd check in the second commit I'll merge it

WhitePeter:
> Looks good to me! If you could squash the whitespace fix into the first commit and the systemd check in the second commit I'll merge it

I'll do that, but I am not quite sure how. I am not very experienced with git, let alone github. So, if you don't mind, could you elaborate a bit on how I do that properly? I think I am in the process of messing things up on my local repo and don't want to push. I tried 'git reset 1fe9b3e' removed the bogus white line and 'git commit --amend config.mk'. But that results in a different commit hash, which is to be expected, I guess from my limited understanding of git. Is this, what I am supposed to do?

I know, this is not the right place to ask for help on general git questions, which I considers this to be, so a simple pointer to the relevant documentation will suffice and I will do my homework. ;) So far I could not find how to change a commit after the fact, other than what I have outlined above.

fwSmit:
It's fine you can ask it here. The easiest way to do it is to do `git rebase -i 1fe9b3e06c4416c2541bbd5076be6aa2bbf5339b^`. You can then move around the relevant commits below the ones they need to be squashed into and replace pick with fixup.
It's wise to keep a backup of your repo by git cloning it to a different location on your pc

WhitePeter:
Thanks for your help so far. I think I have got my local repo to the state you wanted: 3 commits with the fixups discussed here. But, since my local repo and the one hosted on github diverged git refuses to push with the following message:
>  ! [rejected]        master -> master (non-fast-forward)
error: failed to push some refs to 'git@github.com:WhitePeter/dunst.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

Reading the man page of git-push, I think I have found a solution: --force or --force-with-lease. The latter sounds like the right option. Is this what I have to do?

fwSmit:
No just --force. You can do git diff origin/master before that to check. There should be no difference.

bebehei:
Why do we drop the queries to `pkg-config` is there an apparent reason? I admit the systemd-check can be improved, but I don't get why we hardcode the `SERVICEDIR_SYSTEMD` and `SERVICEDIR_DBUS`. That's what pkg-config is actually good for.

WhitePeter:
> Why do we drop the queries to `pkg-config` is there an apparent reason?

Have you read the issue report this is linked to or the commit messages? ;)

> I admit the systemd-check can be improved, but I don't get why we hardcode the `SERVICEDIR_SYSTEMD` and `SERVICEDIR_DBUS`. That's what pkg-config is actually good for.

Apparently it is not, because it returns the directories meant for distro packages, i.e. /usr/{share,lib}... instead of /usr/local/{share,lib}/... where *locally* built software is supposed to be installed. Distro maintainers shouldn't have a problem with overriding the defaults. It's as simple as:
```sh
SERVICEDIR_SYSTEMD=$(pkg-config systemd --variable=systemduserunitdir) \
SERVICEDIR_DBUS=$(pkg-config dbus-1 --variable=session_bus_services_dir) \
make
```

But the number of users building locally very likely outweighs the number of maintainers by orders of magnitude. And the latter only have to setup there package build scripts once. But I doubt they ever relied on the defaults, since PREFIX is /usr for distro packages, anyway. And, alas, change the PREFIX to /usr and the directories in question will be the ones that pkg-config would return anyway. ;)
Also, I opted for hardcoding because other stuff gets hardcoded as well, when there is no room for ambiguity. Have a look at the referenced documentation in my first commit. Those are well documented directories which are very unlikely to change. I mean, just look at everything above SYSTEMDIR* in config.mk. That stuff is hardcoded as well, because it is equally well documented.

The current behavior is just wrong! For once, stuff gets installed outside the default PREFIX (/usr/local) and it will overwrite files belonging to the distro and in turn the distros package manager will overwrite those on the next upgrade. And stuff like that should never happen _by default_.

For reference:
https://www.pathname.com/fhs/pub/fhs-2.3.html#USRLOCALLOCALHIERARCHY

fwSmit:
While on the topic of overwriting, the config file also gets written to a package config location. It wasn't a problem for you, since the debian package installs the system config in a different location, but it should get fixed anyways.
The config is currently installed as `/etc/dunst/dunstrc` by default, but that should be `/usr/local/etc/dunst/dunstrc`. It must still be possible for packagers to install the config file as `/etc/dunst/dunstrc`, so it shouldn't always be `${PREFIX}/etc/dunst/dunstrc`.

WhitePeter:
> While on the topic of overwriting, the config file also gets written to a package config location. It wasn't a problem for you, since the debian package installs the system config in a different location, but it should get fixed anyways.

I am still on the fence about that, hence no report, yet. ;) The actual correct location is in /etc/xdg. I noticed, because I got warned about obolete or missing options. Plus, we are talking about /etc here. No sane package manager should overwrite stuff there without asking. So, yes, dunstrc would replace the distro version, but from then on the file should not be changed by anything automatically unless it happens to be identical to the system version, and then it is a non-issue.
BTW, if I am not mistaken, make install will happily overwrite *any* dunstrc, even it has been changed from the shipped one. I am also not happy about that. ;-/

> The config is currently installed as `/etc/dunst/dunstrc` by default, but that should be `/usr/local/etc/dunst/dunstrc`. It must still be possible for packagers to install the config file as `/etc/dunst/dunstrc`, so it shouldn't always be `${PREFIX}/etc/dunst/dunstrc`.

It won't get picked up there either. I did some research on this and was still weighing the options. There seem to be none on a default installation, at least not in Ubuntu, because $XDG_CONFIG_DIRS is unset and according to the Base directory spec [1], it then defaults to just /etc/xdg, which is very unfortunate. Looks like an oversight, either on the side of freedesktop.org or Debian/Ubuntu. I cannot speak for other distros. I did try to just move dunstrc to /usr/local/etc/xdg/dunst/dunstrc, but it gets ignored, unless I set XDG_CONFIG_DIRS manually in /etc/environment.

[1] https://specifications.freedesktop.org/basedir-spec/latest/

WhitePeter:
I am currently working on a separate pull request for the issue with dunstrc. Or should I push it here?

WhitePeter:
> The config is currently installed as `/etc/dunst/dunstrc` by default, but that should be `/usr/local/etc/dunst/dunstrc`. **It must still be possible for packagers to install the config file as `/etc/dunst/dunstrc`**, so it shouldn't always be `${PREFIX}/etc/dunst/dunstrc`.

'SYSCONFDIR=/etc make'? It is a parameter after all. I am only talking about and changing *defaults* here. "Default" is the antonym of fault, i.e. no value provided is a fault which has to be defaulted. I expect the casual user, myself included, to make those "faults", while one can expect that package maintainers know what they are doing. ;)

bebehei:
> Have you read the issue report this is linked to or the commit messages? ;)

Oh no, sorry. I've missed the issue ref.

I second, that installing files into `/usr/lib` or `/usr/share` is a bad pattern. But when we implemented the code, systemd/dbus did not look into `/usr/local/...`, so this wasn't usable at all. That's also the reason, why we use `pkg-config`, which installs it into `/usr/share`/`/usr/lib` directly instead of its pendant in `/usr/local`.

This might be historic and DBus might respect these folders, too.

WhitePeter:
> I second, that installing files into `/usr/lib` or `/usr/share` is a bad pattern. But when we implemented the code, systemd/dbus did not look into `/usr/local/...`, so this wasn't usable at all. That's also the reason, why we use `pkg-config`, which installs it into `/usr/share`/`/usr/lib` directly instead of its pendant in `/usr/local`.
> 
> This might be historic and DBus might respect these folders, too.

It most definitely does on Ubuntu. The curious thing is that even running 'env XDG_CONF_DIRS=/usr/local/etc/xdg /usr/local/bin/dunst' without any actual changes to the code picks up a dunstrc in the relative path 'dunst/dunstrc'. I can see a difference in the number of warnings about "non-existent" (obsolete?) settings.

BTW, I have sent a separate pull request #910 for the etc stuff, because it only semi-related and a bit more involved, nothing too fancy though. ;)

fwSmit:
Thank you for your work. It looks good to me.

fwSmit (approved):

PR #913, opened 2021-09-18
wl.c: Fix segfault on shutdown

### LINKED ISSUE #912: Segfault on termination introduced by 91c6e93
91c6e9355aec5e0c9d0377ff6a1c93e2a2f23559 seems to have introduced this segfault on termination by CTRL-C or SIGTERM.

### Installation info

- Version: v1.6.1-113-g91c6e93
- Install type: built from git
- Distro and version: Ubuntu 20.04.3

Steps to reproduce:
```sh
$ make
$ ./dunst&
$ kill %%
[1]+  Segmentation fault      ./dunst
```
One commit prior all was fine:
```sh
$ git checkout 91c6e93\^
$ make
$ ./dunst&
$ kill %%
[1]+  Done                    ./dunst
```
Not to complain, since I am running swaywm (Wayland), I do like the changed cursor behaviour. Would be nice though for dunst to shutdown cleanly. ;)

WhitePeter:
This fixes #912.

### DIFF
--- a/src/wayland/wl.c
+++ b/src/wayland/wl.c
@@ -672,13 +672,13 @@ void wl_deinit() {
         if (ctx.registry)
                 wl_registry_destroy(ctx.registry);
 
-        if (ctx.display)
-                wl_display_disconnect(ctx.display);
-
         if (ctx.cursor_theme != NULL) {
                 wl_cursor_theme_destroy(ctx.cursor_theme);
                 wl_surface_destroy(ctx.cursor_surface);
         }
+
+        if (ctx.display)
+                wl_display_disconnect(ctx.display);
 }
 
 static void schedule_frame_and_commit();


WhitePeter:
Sorry, missed a closing brace the first time.

WhitePeter:
Seeing that one cannot do the cleanup unless there is still a display context, wouldn't it be better to just enclose the whole body of wl_deint() in if(ctx.display)? Or do the disconnect after wl_deint() has run, i.e. in the calling function?

fwSmit:
Some things can still be done without the display, like freeing the buffers.
In what situation would we enter this function with the display being NULL?

WhitePeter:
> Some things can still be done without the display, like freeing the buffers.

OK, didn't know that.
> In what situation would we enter this function with the display being NULL?

Don't know, haven't looked *that* closely at the code. ;) I only checked the diff and, since this happens at shutdown, it was kinda obvious, what the fault was. And there is an if statement in there checking for ctx.display already, so I thought at least that should stay or be checked by the caller before entering. But then again, I know too little about the code base, hence the question. Should I maybe just add a comment to warn about doing this kind of stuff *after* wl_display_disconnect()?

fwSmit:
> > Some things can still be done without the display, like freeing the buffers.
> 
> OK, didn't know that.
> > In what situation would we enter this function with the display being NULL?
> 
> Don't know, haven't looked *that* closely at the code. ;) I only checked the diff and, since this happens at shutdown, it was kinda obvious, what the fault was. And there is an if statement in there checking for ctx.display already, so I thought at least that should stay or be checked by the caller before entering. But then again, I know too little about the code base, hence the question. Should I maybe just add a comment to warn about doing this kind of stuff *after* wl_display_disconnect()?

Nah it's fine like this. I'l have a closer look at the code, since I don't know the answer either. If there is a case then mosy other calls in the function will result in a segfault as well.

Thanks for the fix, I'll merge it

Issue #796, opened 2020-12-22
Feature: include keyword in configuration


fwSmit:
The include keyword would work as follows. On any line in the configuration file you could write `include` followed by a file name (this could also be changed to only before the first section. This file name can have a "/*" at the end to say that the entire content of a directory has to be included. When reading a config file, every time an include keyword is read, dunst will read the relevant file before continuing.

Example:

`~/.config/dunst/dunstrc`
```ini
# use manjaro theme
include /etc/xdg/dunst/themes/manjaro

# override certain colours of that theme
[urgency_normal]
    background = "#bbbbbb"
```

`/etc/xdg/dunst/themes/manjaro`
```ini
[urgency_normal]
    background = "#285577"
    foreground = "#ffffff"
```

This feature would be very useful for things like themes. Dunst could ship with a few default themes that can be enabled by just including them in your dunstrc. Other things like layouts and icon sets could be included this way as well.

tsipinakis:
Hm, the problem here is that the ini format is not really designed for such a functionality.

* What would happen if a section is defined in both files? Merge them?
  - If a setting is included in both files which one takes precedent?
  - If, similarly, a setting is included in 2 included files which one takes precedent?

I think we would need to move the config to an entirely new language to achieve something like this cleanly.

fwSmit:
> Hm, the problem here is that the ini format is not really designed for such a functionality.

True, but I think it's possible to implement in a not too confusing way. They only thing that would be confusing here is that the include would be ignoring the section. This can be partially avoided by only allowing includes at the beginning and the end of the file.

This is very rough pseudocode of what it would look like.
```c
struct settings;

void parsefile(file) {
        for(all lines) {
                if (isKey)
                        set(key, value, section);
                else if (isInclude)
                        parsefile(f);
        }
}
```

This would be equivalent to the, IMO harder to implement, c style includes where the contents of the file are just dumped in place of the include. If you do it this way you would have to allow duplicate sections (I don't know how they're currently handled).

To answer your questions:

> What would happen if a section is defined in both files? Merge them?

Every section is just handled when the parser sees it. The variables get the value that is set last.

>  If a setting is included in both files which one takes precedent?

The last one. So if the example in my last comment, the background colour will be `#bbbbbb`. 

>  If, similarly, a setting is included in 2 included files which one takes precedent?

The setting from the last included file.

> I think we would need to move the config to an entirely new language to achieve something like this cleanly.

[...]

Issue #719, opened 2020-05-29
Reload config by SIGUSR1


nick87720z:
I found such feature in tint2, with at least one use case at tint2conf which does it on config save. Not sure if it's more effective than e.g. direct use of inotify. Auto-reload on file change could be done by separate tools like fsniper.

tsipinakis:
This has been requested before (#63) however, dunst was built around the assumption that the settings struct is never modified so adding that would require a rework of some other areas of the code.

Additionally, `SIGUSR1` is already taken by the `pause` command, so it'll probably have to go into dunstctl as `dunstctl reload` or something similar.

tangxinfa:
If dunst supported reload, we can switch theme dynamically.

On my desktop, i exported colors from current emacs theme, and apply it to
polybar, rofi, dunst. But restart dunst will have the risky of lose notifications.

pbrisbin:
I would love to see this feature.

> dunst was built around the assumption that the settings struct is never modified so adding that would require a rework of some other areas of the code.

I haven't looked at the code, but when `polybar` receives `SIGUSR1` to reload config, its output seems to indicate that it replaces its own process with a new one, rather than actually "reloading config" (it shows exactly the same shutting down and starting up messages). Would that be an easier approach that doesn't require making the settings struct mutable?

> Additionally, SIGUSR1 is already taken by the pause command, so it'll probably have to go into dunstctl as dunstctl reload or something similar.

I don't mind a `dunstctl` command for this, but I wanted to mention `SIGUSR2` as another option.

fwsmit:
> I would love to see this feature.

Note that currently you can restart dunst perfectly fine by killing it and letting it autostart again. You can also `systemctl restart dunst` if you use systemd. The only thing is that your notification history is lost.

> dunst was built around the assumption that the settings struct is never modified so adding that would require a rework of some other areas of the code.

I don't expect this to be a big problem. Most of the settings struct will be perfectly fine with being modified.

> I haven't looked at the code, but when `polybar` receives `SIGUSR1` to reload config, its output seems to indicate that it replaces its own process with a new one, rather than actually "reloading config" (it shows exactly the same shutting down and starting up messages). Would that be an easier approach that doesn't require making the settings struct mutable?

This is already possible without it being built into dunst. A proper reload should at least save the notification history. This would be easiest when the same process is kept.

> > Additionally, SIGUSR1 is already taken by the pause command, so it'll probably have to go into dunstctl as dunstctl reload or something similar.
[...]

PR #917, opened 2021-09-21
Settings: Be more compliant with XDG BD spec.

WhitePeter:
Rollback some of the changes in commit 694c596 to be compliant with the
Base Directory Spec, when we can be, and provide an override. By default
there will be *no* fallback value set, if and only if XDG_CONFIG_DIRS is
unset. This is a deliberate digression from the spec. If one sets
XDG_CONFIG_DIRS to any value, including an empty string, dunst behaves
like the spec mandates, with all side effects that entails. This is
mostly only relevant if one installs to a local prefix, i.e. /usr/local,
or, more precisely, if the system-wide dunstrc is expected to be in a
different location than the spec default, i.e. /usr/local/etc/xdg.
Distro packages should not be affected by this, because dunstrc will be
expected in /etc/xdg, anyway, which happens to be the default if
XDG_CONFIG_DIRS is unset or empty.

See the commit and PR messages attached to above commit for more info.

Also, some cosmetic changes.


fwSmit:
I realized that we forgot to update the man page. Could you write something for the man page as well?

WhitePeter:
> I realized that we forgot to update the man page. Could you write something for the man page as well?

I'm on it.

WhitePeter:
> > I realized that we forgot to update the man page. Could you write something for the man page as well?
> 
> I'm on it.

This might take some time, because I want to get ${SYSCONFDIR} into the man page in a flexible way, so it gets updated to the correct location at build time. It looks like pod2man is building the man pages. Is doxygen in any way also involved in that--doesn't look that way at first glance?
Since I know virtually nothing about both pod2man and doxygen, can you give me some pointers to get a head start?

Currently I am thinking of moving 'docs/dunst.1.pod' to 'docs/dunst.1.pod.in' and preprocessing it with something similar to the '$(SED)' scripts for the '*.service' files in 'Makefile'. But that seems hackish to me, since doxygen provides expansion of parameters--I have snooped in the source code for the unbound DNS resolver, which makes use of that quite heavily. I couldn't find anything yet on how to do that with pod2man, though. Any hint is very welcome.

fwSmit:
> > > I realized that we forgot to update the man page. Could you write something for the man page as well?
> > 
> > 
> > I'm on it.
> 
> This might take some time, because I want to get `${SYSCONFDIR}` into the man page in a flexible way, so it gets updated to the correct location at build time. It looks like pod2man is building the man pages. Is doxygen in any way also involved in that--doesn't look that way at first glance?
> Since I know virtually nothing about both pod2man and doxygen, can you give me some pointers to get a head start?
> 
> Currently I am thinking of moving `docs/dunst.1.pod` to 'docs/dunst.1.pod.in' and preprocessing it with something similar to the `$(SED)` scripts for the `*.service` files in `Makefile`. But that seems hackish to me, since doxygen provides expansion of parameters--I have snooped in the source code for the unbound DNS resolver, which makes use of that quite heavily. I couldn't find anything yet on how to do that with pod2man, though. Any hint is very welcome.

Yeah try to not create an intermediate file. You could probably pipe the file to grep and pipe it to pod2man.

Just a note on this PR: I still haven't decided if this is settings lookup scheme is the best, so you might need to change some things.

WhitePeter:
> Yeah try to not create an intermediate file. You could probably pipe the file to grep and pipe it to pod2man.

I will look into that.
> Just a note on this PR: I still haven't decided if this is settings lookup scheme is the best, so you might need to change some things.

All right, let me know. Basically it is almost the same as prior to #910. The only change is the override for the default of g_get_system_config_dirs(). But, since I just found out about the internal documentation, I am thinking to ditch that function altogether and build an array from the concatenation of $XDG_CONFIG_HOME:$XDG_CONFIG_DIRS:$SYSCONFDIR.

That way there is full control of the default and it should pave the way to not stopping at the first found file. My midterm goal is to take all files into account and walk through said array from least important to most important, overwriting values as they come. That is essentially what shells do when they initialize themselves and also what the BD spec seems to expect as well, since it talks about "information" instead of files.

WhitePeter:
I was forced offline for a while, because of an outage in my area, so I got bored and went ahead to achieve said 'midterm goal'. This works as expected on my end. Now one only needs to put settings in the more important config files that are actually different from those in less important ones or the internal defaults. In my limited preliminary tests this works quite nicely, i.e. change a setting in ${SYSCONFDIR}/dunst/dunstrc, uncomment the same in ${XDG_CONFIG_HOME}/dunst/dunstrc and you should see that the former setting gets used. But, since I mostly stick to the default config anyway (no need for bells and whistles) I have only checked some random changes and it worked.

Have a look at apply_settings() which I split out to overwrite settings over and over. From my limited understanding of the code it should just work.

src/settings.c
```diff
@@ -100,52 +145,50 @@ void check_and_correct_settings(struct settings *s) {
...
+                        config_file = stdin;
+                else
+                        config_file = fopen(cmdline_config_path, "r");
+
+                if (config_file) {
+                        apply_settings(config_file);
+                        return;
+                } else
+                        // FIXME: Really DIE or just warn and carry on with defaults?
+                        DIE("Cannot find config file: '%s'", cmdline_config_path);
```
fwSmit, line 181: You can go ahead and make this a warning

src/settings.c
```diff
@@ -100,52 +145,50 @@ void check_and_correct_settings(struct settings *s) {
...
+void load_settings(char *cmdline_config_path) {
+        FILE *config_file = NULL;
+        if (cmdline_config_path) {
+                if (STR_EQ(cmdline_config_path, "-"))
+                        config_file = stdin;
+                else
+                        config_file = fopen(cmdline_config_path, "r");
+
+                if (config_file) {
+                        apply_settings(config_file);
```
fwSmit, line 177: The config given by the command line should be the most important config file. Now it's the least important

src/settings.c
```diff
@@ -100,52 +145,50 @@ void check_and_correct_settings(struct settings *s) {
...
-                config_file = xdg_config("/dunstrc");
-        }
-
-        if (!config_file) {
-                LOG_W("No dunstrc found.");
-        }
-
-        load_ini_file(config_file);
         settings_init();
         set_defaults();
```
fwSmit, line 153: Currently `apply_settings` initializes the settings struct and sets the defaults. If you use apply_settings multiple times, the settings will reset to default before every use. `settings_init` and `set_defaults` should only happen once.

README.md
```diff
@@ -106,6 +106,9 @@ sudo make install
 - `PREFIX=<PATH>`: Set the prefix of the installation. (Default: `/usr/local`)
 - `BINDIR=<PATH>`: Set the `dunst` executable's path (Default: `${PREFIX}/bin`)
 - `DATADIR=<PATH>`: Set the path for shared files. (Default: `${PREFIX}/share`)
+- `SYSCONFDIR=<PATH>`: Set the base directory for system config files. (Default: `${PREFIX}/etc/xdg`)
```
fwSmit, line 109: Is `SYSCONFDIR` unused when you set `SYSCONFFILE`?

fwSmit (commented):
This PR looks good. The use of GQueue seems like a good solution. I don't think it works as intended in it's current state. Specifically, I think only the changes from the last file will actually end up in the settings struct.
I think this PR will also be a good step towards https://github.com/dunst-project/dunst/issues/796

src/utils.c
```diff
@@ -141,7 +141,7 @@ char **string_to_array(const char *string, const char *delimiter)
 {
         char **arr = NULL;
         if (string) {
-                arr = g_strsplit(string, ",", 0);
+                arr = g_strsplit(string, delimiter, 0);
```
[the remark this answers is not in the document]
WhitePeter: Does it cause a conflict? Because I cannot do without, since `XDG_CONFIG_DIRS` gets split on `:`. Or should I pull https://github.com/dunst-project/dunst/commit/9f15d147b0089939ea4efbfef4a92e3a7bdcc23e?

docs/dunst.1.pod
```diff
@@ -122,25 +101,38 @@ missed notifications after returning to the computer.
...
-$XDG_CONFIG_HOME/dunst/dunstrc
+=back
+
+Dunst will search these directories for the following relative paths:
+
+=over 4
+
+dunst/dunstrc
+
+dunstrc (deprecated since v0.2 and ignored if the former is found)
```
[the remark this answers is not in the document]
WhitePeter: Great, that makes the code more concise, too.

src/settings.c
WhitePeter: No, it should be the *only* one in case it can be opened, because then load_settings() returns early.

README.md
WhitePeter: No, `SYSCONFDIR` is being used no matter what. I dropped `SYSCONFFILE` in the code because it did not make much sense, when walking directories and searching for relative paths anyway. I guess, one could say that users who set SYSCONFFILE explicitly might have their reasons, i.e. they do not want it to be installed as a 'live' config. `SYSCONFDIR`, however, dictates the default for `XDG_CONFIG_DIRS`.

WhitePeter:
> I don't think it works as intended in it's current state. Specifically, I think only the changes from the last file will actually end up in the settings struct.

It looked differently in my tests, if you can call them that. I have `/usr/local/etc/xdg/dunst/dunstrc`, which differs from the shipped one on only one line:
```ini
--- dunstrc     2021-09-26 23:32:21.687693760 +0200
+++ /usr/local/etc/xdg/dunst/dunstrc    2021-09-23 15:24:58.343225816 +0200
@@ -326,7 +326,7 @@

 [urgency_critical]
     background = "#900000"
-    foreground = "#ffffff"
+    foreground = "#0fffff"
     frame_color = "#ff0000"
     timeout = 0
```
and ~/.config/dunst/dunstrc:
```ini
--- /usr/local/etc/xdg/dunst/dunstrc    2021-09-23 15:24:58.343225816 +0200
+++ /home/peter/.config/dunst/dunstrc  2021-09-29 19:33:57.746505639 +0200
@@ -325,8 +325,8 @@
     #icon = /path/to/icon

 [urgency_critical]
-    background = "#900000"
-    foreground = "#0fffff"
+    background = "#400000"
+    #foreground = "#ffffff"
     frame_color = "#ff0000"
     timeout = 0
1
```
When I do `notify-send -c critical Test` I get the `background` set in my user config but the `foreground` set in the system-wide config.
> I think this PR will also be a good step towards #796

Nice, I was thinking of something like that, too. But maybe one does not need `include` for that. A proper setup of `XDG_CONFIG_DIRS` could achieve the same or similar, i.e. (staying in the given example, and in prefix `/usr/local`) put the themed dunstrc in `~/.config/manjaro/dunst/dunstrc` and run `env XDG_CONFIG_DIRS=~/.config/manjaro:/usr/local/etc/xdg`, or something like that. Settings in `~/.config/manjaro/dunst/dunstrc` will override those in `/usr/local/etc/xdg/dunst/dunstrc` and, since it is the most important one `$XDG_CONFIG_HOME/dunst/dunstrc` settings would override the manjaro-themed ones, if any. At least, that is the idea and how it *should* work. But if you say it might not, then I will have to change the code until it does.

src/settings.c
WhitePeter: I think I got it. I have found my mistake, why I got errors when not calling `set_defaults()` every time. I now have a working solution that only runs `settings_init()` once, which basically only calls `set_defaults()`.

fwSmit: > P.S.: OK, here is what my further reading of `options_parser.c` turned up: `set_defaults()` changes `struct settings allowed_settings` but [`save_settings()`](https://github.com/dunst-project/dunst/blob/e973cf8b1c86abf687ab3bd7ea4a7285dafd3018/src/option_parser.c#L542) is what actually applies the settings: [`curr_setting` gets overwritten](https://github.com/dunst-project/dunst/blob/e973cf8b1c86abf687ab3bd7ea4a7285dafd3018/src/option_parser.c#L567), if the entry has a value. So I _think_ it does not hurt, at least, but it even seems like `set_defaults()` _must_ run uncondintionally to get the desired outcome, because I get errors on subsequent calls of `apply_settings()` otherwise.

- `set_defaults()` changes the settings struct and sets them to the internal defaults (see settings_data.h, `setting.default_value`).
- `apply_settings()` changes the settings struct and sets them to the values of the loaded ini file (loaded by `load_ini_file()`). This behaviour should probably be changed. I would suggest to make an ini file struct which contains the data of a single ini file. This can then be passed to `apply_settings()`.
- `settings_init()` sets the settings struct to all 0's. Can't quite remember why, but this was needed. This should only be done once, but currently it doesn't matter if you call this function twice, since there's a safeguard (`count`).

I should probably add some documentation about these functions

src/settings.c
fwSmit: Oh right

src/utils.c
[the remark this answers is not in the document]
fwSmit: On second though I think it will fine if you don't do anything. This change will probably disappear from the commit when merging.

fwSmit (commented):
> When I do `notify-send -c critical Test` I get the `background` set in my user config but the `foreground` set in the system-wide config.

That seems good. I'll do some testing as well when you have updated your PR with my comments.


> Nice, I was thinking of something like that, too. But maybe one does not need `include` for that. A proper setup of `XDG_CONFIG_DIRS` could achieve the same or similar, i.e. (staying in the given example, and in prefix `/usr/local`) put the themed dunstrc in `~/.config/manjaro/dunst/dunstrc` and run `env XDG_CONFIG_DIRS=~/.config/manjaro:/usr/local/etc/xdg`, or something like that. Settings in `~/.config/manjaro/dunst/dunstrc` will override those in `/usr/local/etc/xdg/dunst/dunstrc` and, since it is the most important one `$XDG_CONFIG_HOME/dunst/dunstrc` settings would override the manjaro-themed ones, if any. At least, that is the idea and how it _should_ work. But if you say it might not, then I will have to change the code until it does.

Sure that would be a possibility, but there are two things that make me like `include` better:
- You'll have to set a separate XDG_CONFIG_DIRS for dunst, which is not exactly the purpose of that environment variable.
- You cannot easily change dunst's environment during runtime. `include` can be combined with #719 to dynamically change themes.

src/option_parser.c
```diff
@@ -226,20 +226,21 @@ int string_parse_bool(const void *data, const char *s, void *ret)
...
+struct section *get_or_create_section(const char *name)
 {
-        for (int i = 0; i < section_count; i++) {
-                if (STR_EQ(name, sections[i].name)) {
-                        DIE("Duplicated section in dunstrc detected.");
-                }
-        }
+        struct section *s = get_section(name);
+        if (!s) {
+                LOG_D("New section: [%s]", name);
```
WhitePeter, line 233: This was basically reimplementing `get_section()`, the only difference being `DIE()` if there already was one. The new part was split out from `add_entry()` now calls here every time to have a new section created or an existing one reused.

src/option_parser.c
```diff
@@ -268,10 +269,9 @@ struct section *get_section(const char *name)
 
 void add_entry(const char *section_name, const char *key, const char *value)
 {
-        struct section *s = get_section(section_name);
-        if (!s)
-                s = new_section(section_name);
+        struct section *s = get_or_create_section(section_name);
 
+        LOG_D("\t%s=%s", key, value);
```
WhitePeter, line 274: This, combined with the new logging message in `load_ini_file()` makes for a cheap way to see how dunst views the combined ini-files internally.

src/option_parser.c
```diff
@@ -606,7 +607,7 @@ int load_ini_file(FILE *fp)
 
                         g_free(current_section);
                         current_section = (g_strdup(start + 1));
-                        new_section(current_section);
+                        LOG_D("[%s]", current_section);
```
WhitePeter, line 610: Pretty print the section to make it look like in the ini file.

WhitePeter (commented):
OK, I have it ready for review, finally--lost myself in tweaking. ;)

Running `dunst -verbosity debug` should make it obvious, what is happening, i hope. Try
```sh
cat ~/.config/dunst/dunstrc /usr/local/etc/xdg/dunst/dunstrc | dunst -verbosity debug -conf -
```
which was most likely fatal previously, because of section duplication. This example reverses the priority of dunstrc files, because the system-wide one is read last (possibly) overriding settings in the first one. I hope this makes the new power more visible that comes with this.

P.S.: Or, thinking of my sugesstion about drop-ins in #796:
```
cat ~/.config/dunst/dunstrc ~/.config/dunst/dunstrc.d/*.conf | dunst -verbosity debug -conf -
```

src/option_parser.c
WhitePeter: Also, do not call `new_section()` here, because the later call to `add_entry()` will simply create it, if it does not exist already.

src/settings.c
```diff
@@ -37,54 +72,44 @@ static GQueue *xdg_config() {
...
-        if (count == 0)
+        static bool init_done = false;
+        if (!init_done) {
+                LOG_D("Initializing settings");
                 settings = (struct settings) {0};
-        count++;
+
+                init_done = true;
+        }
 }
```
WhitePeter, line 113: I could not make sense of the counter, a bool latch seems more appropriate.

WhitePeter:
>     * You cannot easily change dunst's environment during runtime. `include` can be combined with #719 to dynamically change themes.

I am working on implementing the drop-in idea as a demo of sorts. combined with #719 you can keep a static XDG_CONFIG_DIRS layout and just replace the file which controls the theme or whatever aspect it controls.

src/settings.c
fwSmit: Yup

fwSmit:
@WhitePeter I added a draft version of the release notes [here](https://github.com/dunst-project/dunst/commit/8126ce3bb21a93ce1855a5949e4dc6df7fe2e1d1#diff-9fbfac5aba9b4f62631e81c73b400210bf31c31da365fee4993b982f746b8e3f). Would you mind checking if I included everything that's important for maintainers trying to maintain a dunst package?

PR #933, opened 2021-10-16
Add support for config drop-ins (rudimentary includes as a bonus)

WhitePeter:
This PR introduces support for drop-ins--those files (or snippets) usually residing in a *.d directory to override (some or all settings of) the main/base config in the parent directory.

There is also some very rudimentary support for including additional files, inspired by #796. But it is more of an "anti-sale" I am trying here, since I believe that drop-ins are a sufficient alternative. They can be implemented in a cleaner way and do not have problems that might arise with including files, i.e. include loops as mention [here](https://github.com/dunst-project/dunst/pull/917#issuecomment-944391728). Things like that are simply impossible with drop-ins. So, with the right care one should be able to achieve the same (or similar) results by leveraging drop-ins. See FILES section in 'dunst(1)' for details on the drop-in support. There is, intentionally, no further documentation on including files, because I believe it is more trouble to get right than it is worth, given the rather robust alternative, which is a straight forward evolution of #917.

Technically, one would need to really preprocess all files and keep track of already included files, if you want to get include support right. At first glance, hooking this into 'load_ini_file()' does look nice, until you think about all the potential problems. From my perspective the right place to do this would probably be in 'load_settings()' where the file handling takes place now, because not only do you need to keep track of all files opened by include rules but the corresponding file containing the first include as well. That information is not available in 'load_ini_file()' unless you want to push file handling there. But if you do that there is one big gain from NOT supporting includes and solely relying on drop-ins: the [key-value file parser provided by glib](https://developer-old.gnome.org/glib/unstable/glib-Key-value-file-parser.html), which I stumbled upon while browsing the documentation, could be used to do the dirty work in 'load_ini_file()'. While there is no way to make that work with includes, it seems trivial to replace the cumbersome string parsing and drop includes altogether.

~Oh, one last thing, I forgot to mention in the commit messages: 'load_settings()' now has a safeguard to not set defaults, if called again, with #719 in mind.~ _Edit: This wasn't actually true, since `set_defaults()` ran unconditionally, had overlooked that I moved it because of a failing test, and on second thought it does not make sense to not set defaults on subsequent calls of `load_settings()`_

P.S.: Also dialed back a little on the prefixes for debug messages etc. and made them work with clang as well. ~I just realized that the commit message suggests otherwise, but only debug, error, and critical messages get this additional prefix.~

### DIFF
--- a/src/settings.c
+++ b/src/settings.c
@@ -260,45 +260,52 @@ void check_and_correct_settings(struct settings *s) {
 
 }
 
-void load_settings(char *path) {
+static void process_conf_file(const gpointer conf_fname, gpointer n_success) {
+        const gchar * const p = conf_fname;
+
+        LOG_D("Trying to open '%s'", p);
+        /* Check for "-" here, so the file handling stays in one place */
+        FILE *f = STR_EQ(p, "-") ? stdin : fopen_verbose(p);
+        if (!f)
+                return;
+
+        LOG_I("Parsing config, fd: '%d'", fileno(f));
+        struct ini *ini = load_ini_file(f);
+        LOG_D("Closing config, fd: '%d'", fileno(f));
+        fclose(f);
+
+        LOG_D("Loading settings");
+        save_settings(ini);
+
+        LOG_D("Checking/correcting settings");
+        check_and_correct_settings(&settings);
+
+        finish_ini(ini);
+        free(ini);
+
+        ++(*(int *) n_success);
+}
+
+void load_settings(const char * const path) {
         settings_init();
         LOG_D("Setting defaults");
         set_defaults();
 
-        GQueue *config_files = g_queue_new();
+        GQueue *conf_files = g_queue_new();
 
         if (path) /** If @p path [in] was supplied it will be the only one tried. */
-                g_queue_push_tail(config_files, path);
+                g_queue_push_tail(conf_files, g_strdup(path));
         else
-                get_conf_files(config_files);
+                get_conf_files(conf_files);
 
-        if (g_queue_is_empty(config_files))
+        /* Load all conf files and drop-ins, least important first. */
+        g_queue_reverse(conf_files); /* so we can iterate from least to most important */
+        int n_loaded_confs = 0;
+        g_queue_foreach(conf_files, process_conf_file, &n_loaded_confs);
+        g_queue_free_full(conf_files, g_free);
+
+        if (0 == n_loaded_confs)
                 LOG_W("No configuration file, using defaults");
-        else { /* Load all conf files and drop-ins, least important first. */
-                gchar *path;
-                while ((path = g_queue_pop_tail(config_files))) {
-                        LOG_D("Trying to open '%s'", path);
-                        FILE *f;
-                        /* Check for "-" here, so the file handling stays in one place */
-                        if (!(f = STR_EQ(path, "-") ? stdin : fopen_verbose(path)))
-                                continue;
-
-                        LOG_I("Parsing config, fd: '%d'", fileno(f));
-                        struct ini *ini = load_ini_file(f);
-                        LOG_D("Closing config, fd: '%d'", fileno(f));
-                        fclose(f);
-
-                        LOG_D("Loading settings");
-                        save_settings(ini);
-
-                        LOG_D("Checking/correcting settings");
-                        check_and_correct_settings(&settings);
-
-                        finish_ini(ini);
-                        free(ini);
-                }
-        }
-        g_queue_free(config_files);
 
         for (GSList *iter = rules; iter; iter = iter->next) {
                 struct rule *r = iter->data;
--- a/src/settings.h
+++ b/src/settings.h
@@ -161,18 +161,7 @@ struct settings {
 
 extern struct settings settings;
 
-void load_settings(char *cmdline_config_path);
-
-/** \brief Open files verbosely.
- *
- * This is a wrapper around fopen() and fexists() that does some preliminary
- * checks and sends log messages.
- *
- * @returns The result of the fopen() call.
- * @retval NULL if the fopen() call failed or the file does not exist, see
- * also fexists().
- */
-FILE * fopen_conf(char * const path);
+void load_settings(const char * const path);
 
 #endif
 /* vim: set ft=c tabstop=8 shiftwidth=8 expandtab textwidth=0: */


WhitePeter:
Commit messages and code should now be in sync. Especially ini.c was missing the fix for section names surrounded by spaces, which fell by the wayside when resolving a conflict while rebasing against upstream.

WhitePeter:
> The problem is that I need a subdirectory of `g_get_system_data_dirs`.

Yes, of course but you don't need to, and most probably should not, do that in one go. The spec is called "XDG **Base** Directory Specification" for a reason. You get the base directory and append the appropriate *relative* path(s) after. That is what `get_conf_files()` does as well inside the `for` loop.

> The old code you point to makes a temporary variable with the subdirectory path concatenated to it, but I don't want to recreate it every time.

It was only meant as an example of what `g_get_system_*_dirs()` and `g_get_uses_*_dir()` can do for you.

> You'll also have to make custom code for adding more directories (for example `~/.icons` that is still recommended by the spec).

Yes, that is *one* special case for backwards compatibility. You can just prepend it to the array and drop that part of the code, once backwards compatibility is no longer required, or whatever. All the other directories are specified in the XDG BD spec.

I have cloned your fork of Dunst and was about to prepare a PR with a PoC but, for the time being, let me try some quick and dirty pseudo code to demonstrate:
```
static array all_xdg_data_bd_arr; /* so as to only do this once */
if (all_xdg_data_bd_arr)
        goto BUILD_SEARCH_PATH_ARR;

all_xdg_data_bd_arr[0] = g_get_user_data_dir();
extend_array(all_xdg_data_bd_arr, g_get_system_data_dirs());
/*...*/

BUILD_SEARCH_PATH_ARR:
static array all_icon_theme_dirs;
if (all_icon_theme_dirs)
         goto FIND_ICON_THEME;

all_icon_theme_base_dirs[0] = g_build_filename(g_get_home_dir(), ".icons");
foreach xdg_data_dir in all_xdg_data_dirs do {
        icons_dir = g_build_filename(xdg_data_dir, "icons");
        append_to_array(all_icon_theme_base_dirs, icons_dir);
        pixmaps_dir = g_build_filename(xdg_data_dir, "pixmaps");
        append_to_array(all_icon_theme_base_dirs, pixmaps_dir);
}
/*...*/

FIND_ICON_THEME:
foreach icon_theme_base_dir in all_icon_theme_base_dirs do {
       find(icon_theme_base_dir, "index.theme"); /* find() walks all subdirectories */
}
```
(or something along those lines, haven't had time to flesh out the details and have not fully understood the spec yet, but you should get the gist of it, hopefully. The `goto` labels only serve as replacement for proper function declarations.)

This way you can reuse `all_xdg_bd_arr` for other stuff as well, only need to append the appropriate relative paths similarly to BUILD_SEARCH_PATH_ARR, or even parametrize that as well to be reusable.

> The glib library function is just annoying to deal with, so I opted not to use it.

Why is it annoying? It gets used in other places as well. I find it quite convenient, so why not use it when it is a requirement for Dunst, anyway? Even your code for #947 uses it for other stuff, i.e. GPtrArray. Cannot have it both ways. ;)

> I'm also considering contributing the icon lookup code to [mako](https://github.com/emersion/mako), so I want to stay as independent from glib as possible in that part (the list is easy enough to replace).

That is a good point, but then you would also need to rewrite *any* code that uses Glib, to make it portable to non-Glib projects. It's not the worst idea, though. Let me think a little more about it.


On a related note: is there another channel for communication? This does not seem like the right place. A mailing list would be great, or maybe IRC, even though I am not a frequent IRC user and we seem to be in different time zones--judging by the mail timestamps in my notifications you seem to be in GMT-7 and I am in GMT+2, so email seems more appropriate.

fwSmit:
I'll play around with the code some more, while finishing #947. 

> On a related note: is there another channel for communication? This does not seem like the right place. A mailing list would be great, or maybe IRC, even though I am not a frequent IRC user. ~and we seem to be in different time zones--judging by the mail timestamps in my notifications you seem to be in GMT-7 and I am in GMT+2, so email seems more appropriate.~ _Edit: The timestamps are obviously those of githubs mail server, I just realized._

I'm about in the same timezone as you, so IRC could be an option for real-time communication. I'm not very often online on there, though, so you would have to message me beforehand. Important conclusions should be posted to github, however.

fwSmit (commented):
Please drop commits 05cc9fd2df1de7a80a3a01e666dac7b4c65a9dbd and 46deb084c0f1e8a8ebf0d58d27413506d4e757e7 with `git rebase -i HEAD~9`. Make sure to save that commit somewhere in a different branch if we want to use it later.

fwsmit:
I've reconsidered the decision to read every config file and apply them one after the other. Apart from if it conforms better to the standards, let's consider the user benefits.
Benefits of applying all configs:
- One could change something in the global dunstrc and it will be applied for all users, even if they have a user-specific dunstrc.

Benefits of applying only the last config (let's ignore drop-ins for now):
- We can change the default dunstrc without breaking existing configs. This comes in for the new icon lookup, which isn't compatible with the old one. We could enable it in the default dunstrc, while still allowing other people to use the old mechanism until the next major release. Of course, this can also be used for any other incompatible change.
- Rules will be in the right order. If the system-wide dunstrc defines rule `A` before section `B` and the user-specific config defines them the other way around, rule `A` will still be applied before rule `B`.

To me the benefits of easy backward compatibility, while still providing the best defaults for new users is very nice, while I don't think there are many benefits with applying all configs. What are your thoughts?

> Rules will be in the right order. If the system-wide dunstrc defines rule `A` before section `B` and the user-specific config defines them the other way around, rule `A` will still be applied before rule `B`.

Also, we need to think about this either way, since the drop-ins have this problem as well. We could make all rules from other files independent by adding something like a filename attribute to each rule. I think this is necessary to avoid confusion from users.

fwsmit:
I've also taken a look at the doxygen output  and I do agree that the separate macros might make the doxygen output a bit clearer. But I don't think this should go over the code readability. I found myself scrolling back to the macro definitions way too often, even just while reviewing this PR. So I do stand by my point of removing at least the macros `XDG_CONFIG_DIRS_DEFAULT`, `DROP_IN_DIR` and `DROP_IN_PATTERN`.

`XDG_CONFIG_DIRS_DEFAULT` can be replaced with `SYSCONFDIR` where it's used. The comment can be moved to `SYSCONFDIR`.
The definition of `SYSCONFDIR` is fine by me, since it doesn't decrease the readability.
I don't entirely like `BASE_RC`, but it can stay for now, since it's used in two places.

The relevant comments may be moved to the function description or somewhere in the function. Indented comments inside functions are not a problem to me.

fwsmit:
Also, could you add some tests for the added functions. This comes with the added bonus of valgrind testing for memleaks.

fwsmit:
@WhitePeter have you had some time to look at this?

fwsmit:
@WhitePeter Since you don't seem to have the time to finish this, I'll go ahead an do that (keeping your commit attribution intact, of course).

fwsmit:
This PR is now in a non-compilable state, sorry for that. It doesn't really matter, though, since I am finishing it in #997. Closing this one, as it's not needed anymore.

PR #937, opened 2021-10-17
Add word wrap in again

### LINKED ISSUE #921: `word_wrap` and `ellipsize` setting stop working
After using the new window geometry setting, `word_wrap` and `ellipsize` seems not working, here is my [dunstrc](https://github.com/fitrh/dotfiles/blob/d46e9ce708f196480ea3945a18226f0e6449fea6/.config/dunst/dunstrc).

To reproduce the behavior:

- Set `width = (0, 376)`, `word_wrap = false`, and `ellipsize = end`
- run `dunstify "123456789123456789123456789123456789`

### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon v1.6.1-136-g6cfa0b3`
- Install type: `manually`
- Distro and version: ` Arch Linux 5.14.6-arch1-1`

fwsmit:
This time it's a rule. I changed ellipsize and alignment to a rule as well.
You don't need to change anything to your config, since you can also use rules in the [global] section now. But if you want, you can turn off word_wrap just word just for the notifications that need it.

@fitrh or @wael444 would you mind checking if this matches the desired behaviour?

fixes #921

### DIFF
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -370,10 +370,6 @@ If any of these exists in the format but hasn't been specified in the
 notification (e.g. no icon has been set), the placeholders will simply be
 removed from the format.
 
-=item B<alignment> (values: [left/center/right], default: left)
-
-Defines how the text should be aligned within the notification.
-
 =item B<vertical_alignment> (values: [top/center/bottom], default: center)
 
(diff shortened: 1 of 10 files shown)


fwSmit:
Great, thanks for testing

Issue #822, opened 2021-02-11
MINOR issue; icon paths need to be adjusted for different desktop environments


garberw:
icon paths need to be updated for different desktop environments. for mate desktop environment in dunstrc use ...
<pre><code>icon_path = /usr/share/icons/mate/16x16/status/:/usr/share/icons/mate/16x16/devices/:/usr/share/icons/mate/16x16/apps:/usr/share/icons/hicolor/16x16/apps</code></pre>what else needs to be added to this?

### Installation info

- Version: `v1.5.0-87-g814e620`
- Install type: `manually`
- Distro and version: ` fedora 32`

fwSmit:
Currently we leave it up to users to change the icon_path to something sensible for their desktop environment. We could improve this by adding some more icon sets to the default icon_path. I'm open to adding more suggestions for icon paths.
Also, to make sure the icon_path doesn't get too long, we could make the icon_path recursive. So you only need to add `/usr/share/icons/mate/16x16/` and get all the 16x16 icons.

garberw:
if it is recursive that would probably solve the problem in general.  the first posting is an adequate solution in my specific case.  thank you :-)

fwSmit:
I don't think we would want `/usr/share/icons/` to be the default icon path if it's recursive, since you would get inconsistently sized 
icons, depending on the theme.
Also apparently we should add the hicolor theme as a fallback according to the spec: https:// standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html

garberw:
yes it seems hicolor is closest to what we want :-)

fwSmit:
I've added a few commits, because the default setting wasn't working. Let me know if it works now :)

fwsmit:
The fix has now been merged. You can set an icon theme with `icon_theme = mate, hicolor`

Issue #157, opened 2014-03-19
Support freedesktop icon theme specification


progandy:
I'd like to have support for the xdg icon theme specification so you only have to set a theme name and not twenty directories. Maybe allow one directory to be searched before the icon theme where you can put your own icons if needed.

http://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#icon_lookup

Vladimir-csp:
Perhaps, you can find some useful code in [here](https://github.com/johanmalm/jgmenu)
jgmenu also supports reading icon theme from xsettings with fallback to GTK config files.

bebehei:
My `icon_path` configuration field is [currently over 2000 chars long](https://github.com/bebehei/df/blob/3b702c47eadfa56c4fee1253803c66a82a76a739/i3/dunst/dunstrc#L41), while it could boil down to less than 20 chars. This is insane.

Current implementations of this are either unmaintained, lack scaling support or are not a common library/package. I'll probably write a fresh one for this algorithm. It'll probably give me some experience in libraries.

Vladimir-csp:
I am keeping an eye on three thematically close independent software projects that work with icons: yours, [tint2](https://gitlab.com/o9000/tint2) and [jgmenu](https://github.com/johanmalm/jgmenu). Lightweight implementation of XDG icon theme spec would benefit many, I see collaboration potential.

bebehei:
Thanks for mentioning tint2. On the first glance, the code for the icon theme looks quite neat, too!

https://gitlab.com/o9000/tint2/blob/master/src/launcher/icon-theme-common.c

fwsmit:
The new implementation has been finished and merged. You can try it out by enabling it in the settings. See the man page for how to do that.

PR #947, opened 2021-10-24
Recursive icons V2

fwsmit:
I've rewritten the icon lookup implementation, this time making use of `index.theme` and following the [spec](https://specifications.freedesktop.org/notification-spec/notification-spec-latest.html) closely. The results are very promising with ~0.5-1.5 ms lookup times on my system, instead of ~150 ms.
This PR replaces `min_icon_size` and `max_icon_size` with `icon_size`, because it doesn't make sense in the current implementation to look for multiple size. It also replaces `icon_path` with `icon_theme`. Now you only need to specify 1 theme name instead of sometimes 10's of directories.

It's not finished quite yet, since it'll need some cleanup, but testing and feedback is welcome.

What I still want to improve:
- Allow for multiple icon themes and/or make `icon_theme` rulable.
- Maybe add some form of backward compatibility with old configs by detecting the icon theme from `icon_path` and detecting `icon_size` from `max_icon_size`. I'm currently leaning towards not doing this, though.
- Instead of the above, I could also leave the old code in and add a config switch to enable the new code. Still, I don't know if there's much use in adding backward compatibility, since most people don't look at dunst's output, so when the setting is removed, it will still break their config.
- Increase test coverage

Fixes: #822 #157

ref first try: #864


docs/dunst.5.pod
```diff
@@ -389,8 +388,43 @@ If B<icon_position> is set to off, this setting is ignored.
...
+This setting enables the new icon lookup method. This new system will eventually
+be the old icon lookup.
+
+Currently icons are looked up in the B<icon_path> and scaled according to
+B<min_icon_size> and B<max_icon_size>. Since the B<icon_path> wasn't recursive,
+one had to add a ton of paths to this list.
+This has been drastically simplified by the new lookup method. Now you only have
+to set B<icon_theme> to the name of the theme and B<icon_size> to the icon size
+you want. To enable this new behaviour, set B<enable_recursive_icon_lookup> to
+true in the I<[experimental]> section. See the respective settings for more
```
ncfavier, line 426: This seems wrong, it expects it in the `global` section

fwsmit: Experimental settings are no longer in the experimental section. This is because some of them are rules, so that wouldnt make sense

etrnal70:
Setting icon_size to larger value than the available icons didn't seems to resize it to be bigger. Is this an intended behavior ?

Narrat:
Already loving this change :)
Although there seems to be an uncovered case which causes a segfault if there is no icon found?

Currently testing with various lines in the config:
```
# Paths to default icons.
enable_recursive_icon_lookup=true
icon_theme = Papirus, hicolor, Adwaita
#icon_theme = Adwaita, Papirus
#icon_theme = Papirus, Adwaita
#icon_theme = Papirus
#icon_path = /usr/share/icons/
```
The enabled one segfaults. The second and third line are working (probably because of existing icon in Adwaita). The fourth  one (unsurprisingly) also segfaults.
My main theme Papirus inherits breeze (currently not installed as it seems) and hicolor (installed), so I assume it isn't necessary to specify them. But as I'm currently testing...

Example log:
```
WARNING: Could not find theme breeze
WARNING: Adding default theme Papirus
WARNING: Adding default theme Hicolor
WARNING: Adding default theme Adwaita
DEBUG: [src/dunst.c:run:0076] RUN
INFO: Idle status queried: 0
INFO: Wayland: Hiding window
DEBUG: [src/icon-lookup.c:find_icon_in_theme:0278] Finding icon audio-volume-low in theme Papirus

DEBUG: [src/icon-lookup.c:find_icon_in_theme:0278] Finding icon audio-volume-low in theme (null)

zsh: segmentation fault (core dumped)  dunst -verbosity debug
```

fwsmit:
> Already loving this change :) Although there seems to be an uncovered case which causes a segfault if there is no icon found?
> 
> Currently testing with various lines in the config:
> 
> ```
> # Paths to default icons.
> enable_recursive_icon_lookup=true
> icon_theme = Papirus, hicolor, Adwaita
> #icon_theme = Adwaita, Papirus
> #icon_theme = Papirus, Adwaita
> #icon_theme = Papirus
> #icon_path = /usr/share/icons/
> ```
> 
> The enabled one segfaults. The second and third line are working (probably because of existing icon in Adwaita). The fourth one (unsurprisingly) also segfaults. My main theme Papirus inherits breeze (currently not installed as it seems) and hicolor (installed), so I assume it isn't necessary to specify them. But as I'm currently testing...
> 
> Example log:
> 
> ```
> WARNING: Could not find theme breeze
> WARNING: Adding default theme Papirus
> WARNING: Adding default theme Hicolor
> WARNING: Adding default theme Adwaita
> DEBUG: [src/dunst.c:run:0076] RUN
> INFO: Idle status queried: 0
> INFO: Wayland: Hiding window
> DEBUG: [src/icon-lookup.c:find_icon_in_theme:0278] Finding icon audio-volume-low in theme Papirus
> 
> DEBUG: [src/icon-lookup.c:find_icon_in_theme:0278] Finding icon audio-volume-low in theme (null)
> 
> zsh: segmentation fault (core dumped)  dunst -verbosity debug
> ```

Thanks, I could reproduce it. It's not a problem in the icon lookup, but in the theme initialization. The code currently assumes all inherited themes exist, but that might not be the case.

Fixed in 2cfa8c9340b70b455d71f8b06e6ee1778f338c47

Narrat:
That makes sense. Thanks for the quick fix :)

PR #961, opened 2021-10-29
wayland: fix dunst not quiting on quitting compositor

fwSmit:


fwSmit:
I'll merge this PR now. The CI inexplicably failed at the codecov phase, but that will probably fix itself.

Issue #940, opened 2021-10-21
Dunst cannot be configured through command line arguments


exorcist365:
Why have you removed this, it was one of the most useful functionalities of this program????

aleksfadini:
Please bring back shortcuts and command line arguments

fwSmit:
> @fwSmit would be good to mention this in the README, don't you think? I can make a PR.

True. A PR would be appreciated :)

fwSmit:
For clarity, I'm considering re-adding the command line options in a slightly different from. They would be with double dashes (`--`). It shouldn't be too complicated, it's just something I haven't gotten to.

marcelofern:
For the record, this is a script I was using to change dunst colors based on the pywal theme. This isn't working anymore:
```sh
#!/bin/sh

# source wal colors.
source ~/.cache/wal/colors.sh

pkill --exact dunst
dunst \
    -config=~/.config/dunst/dunstrc \
    -frame_width 1 \
    -lb $color0 \
    -lf $color7 \
    -lfr $color3 \
    -cb $color0 \
    -cf $color7 \
    -cfr $color3 \
    -nb $color0 \
    -nf $color7 \
    -nfr $color3 &
```

aleksfadini:
> For the record, this is a script I was using to change dunst colors based on the pywal theme. This isn't working anymore:
> 
> ```shell
> #!/bin/sh
> 
> # source wal colors.
> source ~/.cache/wal/colors.sh
> 
> pkill --exact dunst
> dunst \
>     -config=~/.config/dunst/dunstrc \
>     -frame_width 1 \
>     -lb $color0 \
>     -lf $color7 \
>     -lfr $color3 \
>     -cb $color0 \
>     -cf $color7 \
>     -cfr $color3 \
>     -nb $color0 \
>     -nf $color7 \
>     -nfr $color3 &
> ```

Would like to know what is a workaround for this, since it stopped working.

marcelofern:
@aleksfadini my understanding is that there is no workaround. @fwSmit your feedback on this would be appreciated.

I've created a script to manually change `dunstrc` based on pywal colors, but it is by no means a better solution, it is not even a good solution at all. However, at the time I am writing this I couln't find a better replacement:

```sh
#!/bin/sh

# source wal colors.
source ~/.cache/wal/colors.sh

DUNST_FILE=~/.config/dunst/dunstrc

# update dunst based on pywal colors.
[...]

PR #963, opened 2021-10-30
docs: add a troubleshooting section

EgZvor:
This should help users that did not expect the removal of some features
discover alternatives. See https://github.com/dunst-project/dunst/issues/940 for example.


fwSmit:
Looks good. Thanks for the PR

Issue #950, opened 2021-10-25
Notification window wrongly sized


vpont:
- Version: `Dunst - A customizable and lightweight notification-daemon v1.7.0-15-g1d7bfbe`
- Install type: `manually`
- Distro and version: ` Arch Linux`

For some reason and with a combination of a few config values and the Iosevka Nerd Font results with the notification window being wrongly sized:
![Screenshot](https://i.postimg.cc/N0RZH7tr/2021-10-25-104159-2560x1440-scrot.png)

You can reproduce it by changing only the following settings:
```
width = (300, 500)
frame_width = 0
font = Iosevka Nerd Font Mono 11
```
And this test notification:
```
dunstify "This is a notification with a long long line

Test"
```

fwSmit:
Maybe @vincentbernat wants to take a look at this?

pawel-szopinski:
I also have similar issue. It does not happen to all notifications, and sometimes even the exact same popup (same content) is  displayed correctly and another time not. 

You can also notice that border widths are not identical on all sides (this also happens randomly).

My xft.dpi is 154 (which means fractional scaling of 1.6) and screen resolution is 3072x1920.

Relevant settings in dunstrc:
width = (0, 300)
height = 1800
scale = 0
frame_width = 2

Here's the screenshot:
![gscreenshot_2021-05-22-160805](https://user-images.githubusercontent.com/25465830/139922731-fed86767-2f3f-443f-bd7d-26ad42ac49a8.png)

fwsmit:
I've posted a PR that might solve this issue. Could you test if this works?

fwsmit:
I also noticed that trying to close notifications with the mouse doesn't work quite well on x11 when scaled. Do you experience the same issue?

vpont:
It looks like it's working!

![screenshot](https://i.imgur.com/4bLbiSa.png)

vpont:
> I also noticed that trying to close notifications with the mouse doesn't work quite well on x11 when scaled. Do you experience the same issue? Edit: I've also posted a fix for that. Would be nice if you could test that as well.

It was working fine before the fix for me.

fwsmit:
> It was working fine before the fix for me.

Is it still working with the fix? 

Before the fix I was able to close notifications with the mouse, but the intersection calculation was as if the notifications were unscaled. So if you have 2 notifications and click on the bottom part of the top notification, the bottom one would close.

vpont:
[...]

PR #965, opened 2021-11-03
x11: fixes for dpi scaling

fwsmit:
This may solve issue with incorrect window sizes.
It also fixes mouse position calculations. So the correct notification will be closed again when clicking on it.

Fixes: #950 #966

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -163,9 +163,9 @@ static void get_text_size(PangoLayout *l, int *w, int *h, double scale) {
         // scale the size down, because it may be rendered at higher DPI
 
         if (w)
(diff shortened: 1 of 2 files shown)


vpont:
Is it enough to fix the text dimensions? I mean, there are other elements that depend on the scale, like icon dimensions, alignments, etc which also use round()

fwsmit:
> Is it enough to fix the text dimensions? I mean, there are other elements that depend on the scale, like icon dimensions, alignments, etc which also use round()

Probably not, but I don't want to risk breaking notifications with random ceils.

PR #968, opened 2021-11-08
proof of concept: dunstctl reload

### LINKED ISSUE #719: Reload config by SIGUSR1

fwsmit:
This is just a proof of concept that reloading can be done pretty easily. It's not completely stable. I've noticed that sometimes the window disappears and doesn't come back anymore.

On reload, all rules are reapplied. When a notification's properties were already changed by a rule this won't be undone. So this might give a different result for existing notification than when re-sending the notification.

Fixes #719

### DIFF
--- a/dunstctl
+++ b/dunstctl
@@ -62,6 +62,9 @@ case "${1:-}" in
 	"context")
 		method_call "${DBUS_IFAC_DUNST}.ContextMenuCall" >/dev/null
 		;;
+	"reload")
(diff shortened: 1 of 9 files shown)


bynect:
@fwsmit if it's fine by you I want to actually implement this

fwsmit:
Yeah, go ahead 😃

PR #970, opened 2021-11-09
dbus/dunstctl: add history command (#307)

nikp123:
The current implementation tries to retrieve notifications stored in the
history buffer (which means they've been either discarded or hidden by
the user) and does so using the existing D-Bus system. However the
output of the dbus-send command is (in my opinion) super ugly and
unusable for any practical scripting application.

This is VERY much still a work in progress.

Any additional suggestions or ideas are welcome.

### DIFF
--- a/contrib/notification-history.sh
+++ b/contrib/notification-history.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+history_json=$(dunstctl history)
+
+history_length=$(echo $history_json|jq -r '.data[] | length')
+
+options=""
+
+for iter in $(seq $history_length); do
+    i=$((iter-1))
+    application_name=$(echo $history_json|jq -r .data[][$i].appname.data)
+    notification_summary=$(echo $history_json|jq -r .data[][$i].summary.data)
+    notification_timestamp=$(echo $history_json|jq -r .data[][$i].timestamp.data)
+    system_timestamp=$(cat /proc/uptime|cut -d'.' -f1)
+
+    how_long_ago=$((system_timestamp-notification_timestamp/1000000))
+
+    notification_time=$(date +%X -d "$(date) - $how_long_ago seconds")
+
+    option=$(printf '%04d - %s: "%s" (at %s)' "$iter" "$application_name" \
+        "$notification_summary" "$notification_time")
+
+    options="$options$option\n"
+done
+options="$options""Cancel"
+
+result=$(echo -e $options|rofi -dmenu -i)
+if [ "$result" = "Cancel" ]; then
+    # Exit if cancelled
+    exit 0
+fi
+if [ "$result" = "" ]; then
+    # Exit on empty strings
+    exit 0
+fi
+if [ "$?" -ne 0 ]; then
+    # Exit on non-zero return values
+    exit 0
+fi
+
+# Get the internal notification ID
+selection_index=$((${result:0:4}-1))
+notification_id=$(echo $history_json|jq -r .data[][$selection_index].id.data)
+
+# Tell dunst to revive said notification
+dunstctl history-pop-id $notification_id
+
--- a/dunstctl
+++ b/dunstctl
@@ -114,40 +114,8 @@ case "${1:-}" in
 			|| die "Dunst controlling interface not available. Is the version too old?"
 		;;
 	"history")
-		counter=0
-		json_string="{ notifications: ["
-		for i in $(dbus-send --print-reply --dest=org.freedesktop.Notifications /org/freedesktop/Notifications org.dunstproject.cmd0.NotificationListAll | grep "string" | cut -d'"' -f2)
-		do
-			case $(($counter%8)) in
-				"0")
-				json_string=$json_string"{body:\"$i\", "
-				;;
-				"1")
-				json_string=$json_string"message:\"$i\", "
-				;;
-				"2")
-				json_string=$json_string"summary:\"$i\", "
-				;;
-				"3")
-				json_string=$json_string"actions:\"$i\", "
-				;;
-				"4")
-				json_string=$json_string"app_name:\"$i\", "
-				;;
-				"5")
-				json_string=$json_string"category:\"$i\", "
-				;;
-				"6")
-				json_string=$json_string"default_action_name:\"$i\","
-				;;
-				"7")
-				json_string=$json_string"icon_path:\"$i\" }, "
-				;;
-			esac
-			((counter=counter+1))
-		done
-		json_string=$json_string"]}"
-		echo $json_string
+		busctl --user --json=pretty -l --no-pager call org.freedesktop.Notifications /org/freedesktop/Notifications org.dunstproject.cmd0 NotificationListAll 2>/dev/null \
+			|| die "Dunst is not running."
 		;;
 	"")
 		die "dunstctl: No command specified. Please consult the usage."
--- a/dunstctl
+++ b/dunstctl
@@ -120,7 +120,7 @@ case "${1:-}" in
 			|| die "Dunst is not running."
 		;;
 	"history-pop-id")
-		method_call "${DBUS_IFAC_DUNST}.NotificationPopHistory" "int32:${2:-0}" >/dev/null
+		method_call "${DBUS_IFAC_DUNST}.NotificationPopHistory" "uint32:${2:-0}" >/dev/null
 		;;
 	"")
 		die "dunstctl: No command specified. Please consult the usage."
--- a/dunstctl
+++ b/dunstctl
@@ -53,7 +53,7 @@ command -v dbus-send >/dev/null 2>/dev/null || \
 
 case "${1:-}" in
 	"action")
-		method_call "${DBUS_IFAC_DUNST}.NotificationAction" "int32:${2:-0}" >/dev/null
+		method_call "${DBUS_IFAC_DUNST}.NotificationAction" "uint32:${2:-0}" >/dev/null
 		;;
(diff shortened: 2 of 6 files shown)


fwsmit:
Maybe you could use [busctl](https://man.archlinux.org/man/busctl.1)? After some playing around I came up with this command: `busctl --user --json=pretty call org.freedesktop.Notifications /org/freedesktop/Notifications org.dunstproject.cmd0 NotificationListAll`.
Btw, don't forget to add the method name to the `introspection_xml` and `methods_fdn`. And add the new command in the `dunstctl --help` output.

src/dbus.c
```diff
@@ -283,6 +288,50 @@ static void dbus_cb_dunst_NotificationShow(GDBusConnection *connection,
...
+        // reverse chronological list
+        for(int i = queues_length_history(); i > 0; i--) {
+                GVariantBuilder *notif_builder;
+                struct notification *n;
+                n = g_list_nth_data(notification_list, i-1);
+
+                // create sub-list per notification
+                notif_builder = g_variant_builder_new(G_VARIANT_TYPE("as"));
+
+                g_variant_builder_add(notif_builder, "s", n->body);
```
fwsmit, line 314: Null strings aren't allowed here it seems. I get the output `CRITICAL: g_variant_new_string: assertion 'string != NULL' failed`

nikp123: so im deciding whether i should throw an dbus error or just return a literal empty string. probably the former

nikp123:
I like the idea. I've been also playing around the idea that instead of arrays i use a dbus dictionary instead

src/dbus.c
fwsmit: I think you should either return an empty string or a string that signals it's null like `(null)`. It's not an error when variables in  a notification are set to NULL.

fwsmit:
> I like the idea. I've been also playing around the idea that instead of arrays i use a dbus dictionary instead
> 
> ie this: https://stackoverflow.com/questions/8846671/how-to-use-a-variant-dictionary-asv-in-dbus-send#38573254

A dictionary seems good, since it would create a stable interface. You can add or remove items without breaking all scripts.

src/dbus.c
nikp123: id probably go for the empty string since a falsely crafted notification could also have "(null)"

fwsmit: They could also have `""` :p. But I'd be fine with either.

src/dbus.c
```diff
@@ -163,13 +166,15 @@ DBUS_METHOD(dunst_NotificationAction);
...
 DBUS_METHOD(dunst_NotificationShow);
+DBUS_METHOD(dunst_NotificationListAll);
 DBUS_METHOD(dunst_Ping);
 static struct dbus_method methods_dunst[] = {
         {"ContextMenuCall",        dbus_cb_dunst_ContextMenuCall},
         {"NotificationAction",     dbus_cb_dunst_NotificationAction},
         {"NotificationCloseAll",   dbus_cb_dunst_NotificationCloseAll},
         {"NotificationCloseLast",  dbus_cb_dunst_NotificationCloseLast},
         {"NotificationShow",       dbus_cb_dunst_NotificationShow},
+        {"NotificationListAll",    dbus_cb_dunst_NotificationListAll},
```
fwsmit, line 177: `NotificationListAll` is currently not a good name, since it only lists the notifications from history. You could include all notifications and add a history attribute or rename the function to something like `NotificationListHistory`.
What are you planning on using this for actually?

nikp123: Basically reviewing notifications that expired but weren't shown (or at least seen) by the user.
image
ie. scripting some way to show notifications of the past, like a sort-of notification history

nikp123: ```result=$(dunstctl history | jq -r '.data[][].summary.data'|rofi -dmenu -i)```

nikp123: and i agree the current name is kinda bad, so ill fix it for now

nikp123:
it's done

src/dbus.c
nikp123: on top of that i could implement so that it also prints the actions (in the same dbus data transfer), allowing the user to implement their own handler for old events

fwsmit: > on top of that i could implement so that it also prints the actions (in the same dbus data transfer), allowing the user to implement their own handler for old events

I don't think that's possible after the notification is closed.

fwsmit: > Basically reviewing notifications that expired but weren't shown (or at least seen) by the user.
> image
> ie. scripting some way to show notifications of the past, like a sort-of notification history

Cool. Wouldn't a "history pop id" be useful for that as well? That way you can take a look at an old notification again.

nikp123: yep, that would be a neat idea

nikp123: also should that be for all notifications or just the ones from history?

fwsmit: > also should that be for all notifications or just the ones from history? i assume the latter because of the name

Yes, the latter. We already have history-pop, but history-pop-id would bring back a specific notification from history. I already took a stab at implementing it, but we decided it wouldn't really be useful in many situations (see https://github.com/dunst-project/dunst/pull/823).

nikp123: @fwsmit also done

fwsmit:
Thanks for working on this. If you want you can add the script you're gonna use this feature with to the contrib directory

nikp123:
> Thanks for working on this. If you want you can add the script you're gonna use this feature with to the contrib directory

sure

fwsmit:
Could you add tests for `queues_history_pop_by_id` and `dbus_cb_dunst_NotificationListHistory`? We try to keep most of the code tested.

src/dbus.c
```diff
@@ -283,6 +293,104 @@ static void dbus_cb_dunst_NotificationShow(GDBusConnection *connection,
...
+
+static void dbus_cb_dunst_NotificationPopHistory(GDBusConnection *connection,
+                                           const gchar *sender,
+                                           GVariant *parameters,
+                                           GDBusMethodInvocation *invocation)
+{
+        LOG_D("CMD: Popping notification from history");
+
+        gint32 id;
+        g_variant_get(parameters, "(i)", &id);
```
fwsmit, line 385: Could you make this a uint, `"(u)"`?

dunstctl
```diff
@@ -113,6 +115,13 @@ case "${1:-}" in
 		dbus-send --print-reply=literal --dest="${DBUS_NAME}" "${DBUS_PATH}" "${DBUS_IFAC_DUNST}.Ping" >/dev/null 2>/dev/null \
 			|| die "Dunst controlling interface not available. Is the version too old?"
 		;;
+	"history")
+		busctl --user --json=pretty -l --no-pager call org.freedesktop.Notifications /org/freedesktop/Notifications org.dunstproject.cmd0 NotificationListHistory 2>/dev/null \
+			|| die "Dunst is not running."
+		;;
+	"history-pop-id")
+		method_call "${DBUS_IFAC_DUNST}.NotificationPopHistory" "int32:${2:-0}" >/dev/null
```
fwsmit, line 123: Could you make this a uint as well?

nikp123:
sure, just not immediatelly

fwsmit (commented):
I have a few more comments on the code :)

fwsmit:
> sure, just not immediately

That's okay. Take your time :+1:

nikp123:
seems that queues_update breaks after 5 messages. is it supposed to do that?

nikp123:
let me describe the breakage in more detail.

If i put into waiting more than 5 messages with ``skip_display`` on, ONLY the first five get to remain in the history after a queues_update(STATUS_NORMAL);

nikp123:
oops seems that i have messed up the prefixes when using sed, one sec

nikp123:
here you go; added some tests

unfortunately i think adding the dbus tests would create way more code than the dbus functions themselves, at which point you would most likely debug the tests themselves than the actual code

dunstctl
fwsmit: This still has to be make uint

contrib/notification-history.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
```
fwsmit, line 1: It's not posix, so please make this /bin/bash

fwsmit (changes requested):
There are still 2 small things that need to be done

fwsmit:
> 

Thanks for adding the tests! It's fine if you don't add the dbus tests.

fwsmit:
It seems like a conflict was caused by me merging another PR

nikp123:
here we go

src/dbus.c
```diff
@@ -328,12 +433,16 @@ static void dbus_cb_dunst_RuleEnable(GDBusConnection *connection,
                 target_rule->enabled = true;
         else if (state == 2)
                 target_rule->enabled = !target_rule->enabled;
+>>>>>>> upstream/master
```
fwsmit, line 436: There are still version control markers here and in other places

nikp123: oops....

nikp123: anyway fixed

fwsmit (approved):

fwsmit:
Merged it!

fwsmit:
I'm thinking that we don't need a separate command for `history-pop-id`. `history-pop` can be used with an argument instead. I'll go ahead and change that.

nikp123:
thanks

PR #971, opened 2021-11-10
Make get_notification_at work with Y offset

mattfbacon:
Currently this doesn't work, and the get_notification_at code assumes that the notification is at the top of the screen. This is not correct considering the user can arbitrarily position the notification.

TODO: support bottom-left and bottom-right origin.


mattfbacon:
Nevermind, it was an issue on my end.

PR #981, opened 2021-11-21
Feature: Enable or disable rules on the fly

lluises:
# Feature: Enable or disable rules on the fly
This feature allows to enable or disable rules through dunstctl whenever you please.
Its main use case for me is muting certain notifications for a period of time, or changing the color of notifications depending on which "mode" I want. Instead of changing the configuration and reloading dunst (which is tedious), with this feature I can easily script it and apply it to new notifications.

# Example:
In configuration:
```
[red]
    background = "#ff0000"
    summary = "*"
    enable = false
```

Then you might enable it whenever you want with
```
dunstctl rule red enable
```

Or disable it when you're tired
```
dunstctl rule red disable
```

Rules are enabled by default unless you explicitly disable them.

# Comment
I created it for myself, but maybe someone else can take advantage of it, so I'm sharing the code. Its a simple  implementation, it doesn't require many lines of code, it hardly changes the current code, and it shouldn't cause any problems with existing configurations, nor with people who will not use the feature.

I'm kinda new to Github so if there is something I'm doing wrong, don't hesitate to teach me ^.^

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -175,8 +175,8 @@ static struct dbus_method methods_dunst[] = {
         {"NotificationCloseAll",   dbus_cb_dunst_NotificationCloseAll},
         {"NotificationCloseLast",  dbus_cb_dunst_NotificationCloseLast},
         {"NotificationShow",       dbus_cb_dunst_NotificationShow},
-        {"RuleEnable",             dbus_cb_dunst_RuleEnable},
         {"Ping",                   dbus_cb_dunst_Ping},
+        {"RuleEnable",             dbus_cb_dunst_RuleEnable},
 };
 
 void dbus_cb_dunst_methods(GDBusConnection *connection,
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -80,6 +80,10 @@ static const char *introspection_xml =
     "        <method name=\"NotificationCloseLast\" />"
     "        <method name=\"NotificationCloseAll\"  />"
     "        <method name=\"NotificationShow\"      />"
+    "        <method name=\"RuleEnable\">"
+    "            <arg name=\"name\"     type=\"s\"/>"
+    "            <arg name=\"state\"    type=\"i\"/>"
+    "        </method>"
     "        <method name=\"Ping\"                  />"
 
     "        <property name=\"paused\" type=\"b\" access=\"readwrite\">"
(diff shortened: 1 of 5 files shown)


src/dbus.c
```diff
@@ -163,13 +167,15 @@ DBUS_METHOD(dunst_NotificationAction);
...
 DBUS_METHOD(dunst_NotificationShow);
+DBUS_METHOD(dunst_RuleEnable);
 DBUS_METHOD(dunst_Ping);
 static struct dbus_method methods_dunst[] = {
         {"ContextMenuCall",        dbus_cb_dunst_ContextMenuCall},
         {"NotificationAction",     dbus_cb_dunst_NotificationAction},
         {"NotificationCloseAll",   dbus_cb_dunst_NotificationCloseAll},
         {"NotificationCloseLast",  dbus_cb_dunst_NotificationCloseLast},
         {"NotificationShow",       dbus_cb_dunst_NotificationShow},
+        {"RuleEnable",             dbus_cb_dunst_RuleEnable},
```
fwsmit, line 178: This list should be sorted apparently (there is a test about it that failed). You can test on your own system by running `make test`.

src/settings_data.h
```diff
@@ -586,6 +587,17 @@ static const struct setting allowed_settings[] = {
                 .parser_data = NULL,
                 .rule_offset = offsetof(struct rule, set_icon_size),
         },
+        {
+                .name = "enable",
```
fwsmit, line 591: I would name it "enabled"

fwsmit (changes requested):
This is a great feature. I did post an experimental patch for adding a reload feature (#968), but I think this will still be useful after that for easy edits. I do think that `dunstctl reload` should not reset the `enabled` state of the rules, but that can be easily done.
I have a few small comments, but other than that it looks good! Thanks for sharing your code.

src/dbus.c
lluises: Ups, I didn't noticed, thank you. Fixed with https://github.com/dunst-project/dunst/pull/981/commits/f4e445b7f6c4df6e907a4f9bdf597ca24ca696e2.

src/settings_data.h
lluises: Done with https://github.com/dunst-project/dunst/pull/981/commits/f4e445b7f6c4df6e907a4f9bdf597ca24ca696e2, should I rename it elsewhere?

fwsmit: Could you rename it in the rule struct and it's usages as well?

src/dbus.c
fwsmit: Great

src/settings_data.h
lluises: Thats easy. Done in https://github.com/dunst-project/dunst/pull/981/commits/1440a0d94ad3950fefb92b454babf7e2de9ffe6e .

fwsmit: Thanks!

fwsmit (approved):

fwsmit:
It all looks good, thanks! I've merged it!

fwsmit:
Thanks for sharing your code :)

lluises:
Its a pleasure :D

Issue #972, opened 2021-11-10
Icon position


RustemB:
How to put icon at the corner of notification? not at the center of left/right side only

fwsmit:
If you want that, you should send a PR :)

RustemB:
i'm bad in C :disappointed:

fwsmit:
It should be pretty simple to implement. You can take a look at how the other positionings are implemented

tfaughnan:
Is is not already achievable through a combination of both `icon_position` and `vertical_alignment`?

fwsmit:
Don't think so, but you can try

bynect:
did the pr fix this?

PR #985, opened 2021-11-24
Add support for "top" option to icon_position rule and hide_text rule

m-bartlett:
### EDIT: Features for other pull requests

A list of possible features recommended as other PRs in below conversation to keep track:

- Make height a notification rule
- Progress bar alignment (made centered within this PR)
- Add explicit `vertical_padding` setting, make `padding` act as a default for both `horizontal_padding` and `vertical_padding` if either is unset
  https://github.com/dunst-project/dunst/pull/985#discussion_r786239475
---

This adds a setting `center_icon_when_no_text` which centers the notification icon if the notification's text content is empty or only space characters.

The motivation for this feature is to support simple graphical notifications using the progress bar for volume, brightness, or any other such notification where the icon _is_ the notification.

I had attempted to achieve this sort of notification using symbolic fonts (I am using a [nerd-font](https://www.nerdfonts.com)) however `dunst` seems to position font characters pixel-wise which causes font faces (despite being monospaced) to not be placed consistently. Notice in this  animation of volume notifications using purely symbolic font characters that the "icon" moves slightly by a few pixels along the x-axis when comparing the "low-volume" symbol to the rest of the symbols:
<p align="center">off-center-monospace-icons</p>

Admittedly the difference is slight but I believe having more freedom over the icons is a benefit, as the icon images can enforce being a consistent dimension.

Here is an animation of the same volume notification redone using my proposed code and centered icons instead of symbolic characters from fonts:
<p align="center">centered-icon
<br/><i>notice the base of the speaker icon does not move between each frame</i>
</p>

So long as the body text is exclusively space characters, longer strings of spaces can be used to increase the width of the notification if the user prefers:
<p align="center">wider-from-spaces</p>

And as the setting name implies, if there is non-whitespace notification text content, the icon proceeds to respect the user's **`icon_position`** preference:
<p align="center">uncentered-icon</p>

For reference in my `dunstrc` I am using the following rule to achieve this notification:
```ini
[volume]
    category="volume"   # Match criteria
    format = %b
    alignment = center
    history_ignore = yes
    fullscreen = show
```

and this sample `notify-send` command to create an "empty" notification body with an icon:
```
notify-send \
  -c volume \
  -h string:synchronous:volume \
  -h int:value:57 \
  -i /usr/share/icons/Papirus/48x48/status/notification-audio-volume-high.svg \
  'ignored-summary' '<span></span>'
```

---

I tried searching through this repository but I didn't find any code style guidelines or contribution guidelines. If those exist and I've overlooked them, please feel free to refer me to them. Otherwise please let me know if there is anything I should change about this PR to improve consistency.


fwsmit:
Since you're making a rule for that notification anyways, it may be better to add an option `center` to `icon_position` instead. In the future this could also be expanded to more positions, as was requested in https://github.com/dunst-project/dunst/issues/972.

m-bartlett:
For clarity, do you imagine that `icon_position: center` might yield something like this given the notification content from above (assuming the notification text is configured to be center-justified)?

<p align="center">center icon</p>

I don't think this would be too hard to implement, although I personally prefer separating the feature and having the icons with text remain to the left or right. A compromise is potentially having this behavior captured in other extra option values like `icon_position: center-left` or `icon_position: center-right`, i.e the icon is centered with no text or left/right with text. They're not intuitive option names to me, but they could be described in detail in the man page, or could be made to be more verbose (e.g. `icon_position: right-centered-when-no-text`). What's your opinion between these approaches?

fwsmit:
> For clarity, do you imagine that `icon_position: center` might yield something like this given the notification content from above (assuming the notification text is configured to be center-justified)?
> 
> center icon

Yes, I was thinking something like that. Now that you show that, it occurs to me that the name `top` might be more fitting than `center`.

> I don't think this would be too hard to implement, although I personally prefer separating the feature and having the icons with text remain to the left or right. A compromise is potentially having this behavior captured in other extra option values like `icon_position: center-left` or `icon_position: center-right`, i.e the icon is centered with no text or left/right with text. They're not intuitive option names to me, but they could be described in detail in the man page, or could be made to be more verbose (e.g. `icon_position: right-centered-when-no-text`). What's your opinion between these approaches?

Maybe your use case wasn't entirely clear to me, but I thought you wanted all your volume notifications to have no text and other kinds of notifications with text need to have their icon on the left. If that is the case you can achieve this the following way:

`dunstrc`
```ini
[other]
   icon_position = left
   format = "%s\n%b"

[volume]
    category="volume"   # Match criteria
    format = %b
    icon_position = top
    history_ignore = yes
    fullscreen = show
```

Rules are flexible enough that we don't need the added functionality of `center_icon_when_no_text`.

When you want to exactly mimic the behaviour of `center_icon_when_no_text`, e.g. automatically center the icon when no text is present in the notification you can do this by using a script like this instead of dunstify:

```sh
# didn't test this, but something like this might work
if [ -z $1]
then
   shift 1
   dunstify "empty-summary" $@
else
  dunstify $@
fi
```

`dunstrc`
```ini
[other]
   icon_position = left
   format = "%s\n%b"

[empty]
    summary = "empty-summary"   # Match criteria
    format = %b
    icon_position = top
    history_ignore = yes
    fullscreen = show
```

Does either solutions satisfy your use case?

m-bartlett:
The config you shared and what you suggest with setting `icon_position` for each notification rule would certainly satisfy my desired use case much more cleanly with the `top` option implemented.

However I am not observing that `icon_position` can be overridden in the rule for a specific notification match, I'm testing with the latest commit on master. It doesn't seem to be listed as [one of the properties that can be overridden in the comment in the sample rc](https://github.com/dunst-project/dunst/blob/master/dunstrc#L336). I have `icon_position = left` set globally (perhaps that is the issue?) but trying to modify it to `right` or `off` in the rule still results in the notification showing the icon to the left.

i.e. my config is
```ini
[global]
...
    icon_position = left
...
[volume]
    category="volume"
    format = %b
    alignment = center
    history_ignore = yes
    fullscreen = show
    icon_position = right  # <---
```

After reloading `dunst`, testing with:
```
notify-send -c volume -h string:synchronous:volume 'test summary' 'test body' -i /usr/share/icons/Vibrancy-Colors/status/96/audio-volume-high-panel.png
```
yields:

image


Do I have this configured incorrectly? Or am I misunderstanding anything?

fwsmit:
> The config you shared and what you suggest with setting `icon_position` for each notification rule would certainly satisfy my desired use case much more cleanly with the `top` option implemented.

Great, let's do that then.
 
> However I am not observing that `icon_position` can be overridden in the rule for a specific notification match, I'm testing with the latest commit on master. It doesn't seem to be listed as [one of the properties that can be overridden in the comment in the sample rc](https://github.com/dunst-project/dunst/blob/master/dunstrc#L336). I have `icon_position = left` set globally (perhaps that is the issue?) but trying to modify it to `right` or `off` in the rule still results in the notification showing the icon to the left.
> 
> i.e. my config is
> 
> ```ini
> [global]
> ...
>     icon_position = left
> ...
> [volume]
>     category="volume"
>     format = %b
>     alignment = center
>     history_ignore = yes
>     fullscreen = show
>     icon_position = right  # <---
> ```
> 
> After reloading `dunst`, testing with:
> 
> ```
> notify-send -c volume -h string:synchronous:volume 'test summary' 'test body' -i /usr/share/icons/Vibrancy-Colors/status/96/audio-volume-high-panel.png
> ```
> 
> yields:
> 
> image
> 
> Do I have this configured incorrectly? Or am I misunderstanding anything? Or if that feature also needs to be implemented, I could potentially provide that as a commit here as well as the `top` icon placement.

I didn't notice this before, but `icon_position` currently is only allowed as a global setting as you can see [right here](https://github.com/dunst-project/dunst/blob/e971028218c932eec30708ad8a686d2047a73f78/src/settings_data.h#L1049-L1058). So first `icon_position` needs to be made a rule. That isn't too hard, but the right steps need to be taken for that. I've made a checklist for myself that I'll share with you:

### How to add/change a rule

- Add variable to `struct rules` in `rules.h` (make sure to read the comment at the top)
- Add variable to to `struct notification` in `notification.h`
- Apply the rule in `rule_apply` in `rules.c`
- Change the listing in `settings_data.h` (make sure to move it to the other rule listings for clarity)
- Add the default rule value in `settings_data.h` (usually -1 or NULL)
- Set default value in notification.c (`notification_create`). This is where the
  real default is set.
- Free the variable in `notification.c` if dynamically allocated.
- Free the variable in `rules.c` if dynamically allocated.
- Remove the setting from the global settings struct in `settings.h`.
- Actually use the new setting.
- Update the documentation
- Test that it works

An example of making something a rule can be found in edc6f5a8c7a51a56b591cfa72618a43adc7b8d11

m-bartlett:
>So first `icon_position` needs to be made a rule. That isn't too hard, but the right steps need to be taken for that. I've made a checklist for myself that I'll share with you:
> 
> ### How to add/change a rule
> * Add variable to `struct rules` in `rules.h` (make sure to read the comment at the top)
> * Add variable to to `struct notification` in `notification.h`
> * Apply the rule in `rule_apply` in `rules.c`
> * Change the listing in `settings_data.h` (make sure to move it to the other rule listings for clarity)
> * Add the default rule value in `settings_data.h` (usually -1 or NULL)
> * Set default value in notification.c (`notification_create`). This is where the
>   real default is set.
> * Free the variable in `notification.c` if dynamically allocated.
> * Free the variable in `rules.c` if dynamically allocated.
> * Remove the setting from the global settings struct in `settings.h`.
> * Actually use the new setting.
> * Update the documentation
> * Test that it works
> 
> An example of making something a rule can be found in [edc6f5a](https://github.com/dunst-project/dunst/commit/edc6f5a8c7a51a56b591cfa72618a43adc7b8d11)

Thanks for the guidance, I'll work towards making `icon_position` a mutable rule property, as well as implementing the `top` option. This may take me a few weeks due to the upcoming holidays occupying much of my free time, but once I feel I have a candidate set of changes I will force push over the source branch used in this PR and @ you for a review.

fwsmit:
Nice! There's also this issue about more icon positions (#972), if you want to.

fwsmit:
Hey @m-bartlett have you had time to work on this?

m-bartlett:
Not quite yet but it has been on my todo list, just putting some finishing touches on different project

m-bartlett:
Made a little bit of headway, I've added `icon_position` as a notification rule, added "top" to the icon_position enum/docstrings, and I've got the basics for the layout changes needed to position the icon at the top. Currently working around modifying the notification height/width, and text dimensions/placement appropriately when the icon_position is set to "top". No pushes to this branch yet.

I am rolling with the assumption that the global setting "height" should still be respected absolutely even with the icon being on top generally adding a lot more height. If you use a relatively large icon it causes the notification to exhaust its height before ever putting the text. It might be hard to restrict the height of "normal" notifications while supporting a few custom rules that allow a top-icon since the height boundary will be filled in differently. Let me know if you think this should be different, i.e. height should only restrict the text's rendered height and ignore the icon.

An example of what I mean:

With height 300 icon: left
image

With height 300 icon: top
image

m-bartlett:
@fwsmit I've got the core of these features implemented as far as my testing has proven. Still no pushes to this PR's branch yet as I want to run everything by you first before pushing

Things needing your decision:

- Progress bars have a min/max width setting but they are aligned left if the notification is wider than the progress bar. I changed this behavior to center the progress bar within the notification window (see my screenshot below). As far as I'm able to tell, there isn't a configuration option for this behavior (yet). This could probably be added as a configurable setting but let's save that for a different PR if necessary. I can easily undo this change if you think that's best.

- I'm finding with the vertical stacked layout implied by having the icon on "top", that the notification text (i.e. summary+body) always has a height, even if these fields are "empty" (i.e. just whitespace characters, as dbus doesn't seem to support a truly empty notification). In practice I find this to cause a noticeably large empty gap in a "textless" notification which is the use-case I'm trying to support. Consider this screenshot from my working branch:

  <p align="center">top-icon-varying-body-lengths</p>

  My ideal in the top case with "no text" is that this padding would not be added to the layout. 
  I'm aware from the man pages that `format=""` is reserved for silencing certain matched notifications so that's not an option for configuring a notification with no text. However, what are you thoughts on adding another rule that would indicate not to render any text? Maybe something like like `textless=yes` or `no_text=yes`.
  I propose this because there is no straightforward way of checking if a notification is "empty" as A) the notification cannot actually really be empty and B) Pango will always render these not-actually-empty strings with a height. I could add similar logic from my original PR (i.e. if the body is only space characters then it is empty) but I'm skeptical that that's a good approach.

fwsmit:
> I am rolling with the assumption that the global setting "height" should still be respected absolutely even with the icon being on top generally adding a lot more height. 

Yes that seems right.

> If you use a relatively large icon it causes the notification to exhaust its height before ever putting the text. It might be hard to restrict the height of "normal" notifications while supporting a few custom rules that allow a top-icon since the height boundary will be filled in differently. Let me know if you think this should be different, i.e. height should only restrict the text's rendered height and ignore the icon.

The default max height could probably be a bit higher, but for these cases you can always set the max height higher with a rule.


> I've got the core of these features implemented as far as my testing has proven. Still no pushes to this PR's branch yet as I want to run everything by you first before pushing

Feel free to push unfinished code, you can always force-push later.



> Things needing your decision:
> 
>     * Progress bars have a min/max width setting but they are aligned left if the notification is wider than the progress bar. I changed this behavior to center the progress bar within the notification window (see my screenshot below). As far as I'm able to tell, there isn't a configuration option for this behavior (yet). This could probably be added as a configurable setting but let's save that for a different PR if necessary. I can easily undo this change if you think that's best.

A new progress bar position setting best for another PR. But for now, the centering behaviour seems fine.

>       My ideal in the top case with "no text" is that this padding would not be added to the layout.
>       I'm aware from the man pages that `format=""` is reserved for silencing certain matched notifications so that's not an option for configuring a notification with no text. However, what are you thoughts on adding another rule that would indicate not to render any text? Maybe something like like `textless=yes` or `no_text=yes`.

`format=""` does exactly what you want. You can match the desired notifications and remove the text from there. What would a `textless=yes` option add?

fwsmit:
This feature cannot really be tested via automated testing, but it would be nice to easily test it in the future, so could you add a few tests to `test/functional-tests/test.sh`?

m-bartlett:
> The default max height could probably be a bit higher, but for these cases you can always set the max height higher with a rule.

This would be sensible, however neither height or width are supported as rules currently, just global settings. Should I go ahead and make these rules?

> format="" does exactly what you want. You can match the desired notifications and remove the text from there. What would a textless=yes option add?

`format=""` suppresses the notification, i.e. it doesn't show the notification at all. This feature is somewhat confusingly documented, but for example in settings_data.h:

```
.name = "always_run_script",
                .section = "global",
                .description = "Always run rule-defined scripts, even if the notification is suppressed with format = \"\".",

I'm suggesting the `no_text` rule as a means to inform the notification rending to skip rendering  text, but still show the rest of the notification (like the icon and progress bar).
For example, in this notification the red box shows where whitespace text is still being rendered:
![2022-01-12_11-26](https://user-images.githubusercontent.com/85039141/149200015-dfb92341-2b5a-443f-b80d-a896aa5f91d3.png)

Whereas with a rule like `no_text=yes` the text rendering would be skipped and would not occupy height:
![2022-01-12_11-19-2](https://user-images.githubusercontent.com/85039141/149200104-b1bf7c0d-dcc4-47d5-8169-1af70a34c906.png)


> This feature cannot really be tested via automated testing, but it would be nice to easily test it in the future, so could you add a few tests to test/functional-tests/test.sh?

Yep I will add some test cases there

**Comment by fwsmit** [comment]
> This would be sensible, however neither height or width are supported as rules currently, just global settings. Should I go ahead and make these rules?


> `format=""` suppresses the notification, i.e. it doesn't show the notification at all. This feature is somewhat confusingly documented, but for example in man page for dunstrc:

I was wrong on both of these, thanks for correcting. I it would be sensible to add this option, but I would call it `hide_text`.

As for the height, yes, this should definitely be made a rule, but I would prefer if you did that in another PR. The width cannot really be made a rule, since we are limited to a single notification window. Until that's fixed, I think it's best to stick to one width.


> ```
> always_run_script (values: [true/false] default: [true]
>            Always run rule-defined scripts, even if the
>            notification is suppressed with format = "". See
>            SCRIPTING.
> ```
> 
> However later on in the same man page under the markup section:
> 
> > If the format is set to an empty string, the notification will **not** be suppressed.

This documentation should definitely be improved. Also, I think that letting `format=""` be a way of skipping is a bit confusing, but for backward compatibility we should let it in for a bit.

**Comment by m-bartlett** [comment]
Okay, so for next actionables:
* Make `height` (i.e. max height) a rule
* Add `hide_text` as a rule
* Don't modify the behavior of `format`

Is this correct?

**Comment by fwsmit** [comment]
> Okay, so for next actionables:
> 
>     * Make `height` (i.e. max height) a rule (separate PR)
> 
>     * Add `hide_text` as a rule
> 
>     * Don't modify the behavior of `format`
> 
> 
> Is this correct?

Yep, but I would do the `height` changes in a separate PR.

**Comment by m-bartlett** [comment]
Okay I will work on adding the `hide_text` rule as well as some test cases. Once it appears to be working as intended I will push to this branch.

File: src/draw.c [hunk 1]
```diff
@@ -658,7 +701,7 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, dou
                 unsigned int frame_width = settings.progress_bar_frame_width,
                              progress_width = MIN(width - 2 * settings.h_padding, settings.progress_bar_max_width),
                              progress_height = settings.progress_bar_height - frame_width,
-                             frame_x = settings.h_padding,
+                             frame_x = width/2 - progress_width/2,
```
m-bartlett, line 704: This is centering the progress bar instead of left-aligning

src/settings_data.h
```diff
@@ -1049,7 +1074,7 @@ static const struct setting allowed_settings[] = {
         {
                 .name = "icon_position",
                 .section = "global",
-                .description = "Align icons left/right/off",
+                .description = "Default icon alignment (left/right/top/off)",
```
m-bartlett, line 1077: In your checklist for changing a setting to a rule one of the bullets is
> Remove the setting from the global settings struct in settings.h`.

I have opted to keep `icon_position` as a global setting so that users can explicitly set a default value for all notifications. Consequently, if `icon_position` is unset in both cases, it will use the global setting's default.

Implemented here in the notification struct initialization:

https://github.com/dunst-project/dunst/blob/cfada4261f8883e849411c1a87959ac1d416df5b/src/notification.c#L421

test/functional-tests/test.sh
```diff
@@ -188,6 +188,27 @@ function progress_bar {
...
+
+function hide_text {
+    start_dunst dunstrc.hide_text
+    ../../dunstify -a "dunst tester" -I '../data/icons/valid.png' -u n "text not hidden" "You should be able to read me!\nThe next notifications should not have any text."
+    ../../dunstify -a "dunst tester" -u l "text hidden" "If you can read me then hide_text is not working."
+    ../../dunstify -a "dunst tester" -I '../data/icons/valid.png' -u l "text hidden + icon" "If you can read me then hide_text is not working."
+    ../../dunstify -a "dunst tester" -h int:value:$((RANDOM%100)) -I '../data/icons/valid.png' -u l "text hidden + icon + progress bar" "If you can read me then hide_text is not working."
+    keypress
+}
+
```
m-bartlett, line 211: Here are the test notifications I get for reference:
<span> icon_left icon_top icon_right icon_off hide_test </span>

m-bartlett:
@fwsmit I believe this is ready for your review, let me know if there's anything I should change

test/functional-tests/test.sh
fwsmit: Looks good. I see see that you're using the icons in the data directory. That icon is just a small text file, so it looks like corrupted data. I would use the default icon instead by leaving out the `-I` argument. It isn't a big problem that the icon is not located inside the dunst repository.

src/draw.c
```diff
@@ -176,6 +176,15 @@ static int get_text_icon_padding()
         }
 }
 
+static int get_vertical_text_icon_padding()
+{
+        if (settings.text_icon_padding) {
+                return settings.text_icon_padding;
```
fwsmit, line 182: I believe the text_icon_padding should only be used when the icon is at the top. If the icon icon is not top, you should just use normal vertical padding.
Same issue in `get_horizontal_text_icon_padding()`.

This makes me notice that the code already had the issue that when the icon is off, the `text_icon_padding` is still used. Could you change that as well?

src/settings_data.h
fwsmit: This should not be needed, since all rules also work in the global section (except for the filtering rules, since that would make the global section not global anymore).

fwsmit (commented):
Overall the changes look good. I'll take a closer look at the changes in draw.c later, but I don't expect to find any big problems. I've left a few comments here and there.

src/settings_data.h
m-bartlett: I see, I wasn't aware it worked that way. In that case I'll remove the global setting and its documentation

src/draw.c
m-bartlett: Do you have an opinion on adding a `vertical_padding` setting? I would imagine the `padding` setting would then just serve as a default if either vertical or horizontal are unset

m-bartlett: > I believe the text_icon_padding should only be used when the icon is at the top. If the icon icon is not top, you should just use normal vertical padding.
Same issue in get_horizontal_text_icon_padding().

I attempted to address this in the latest commits

> This makes me notice that the code already had the issue that when the icon is off, the `text_icon_padding` is still used. Could you change that as well?

I wasn't able to reproduce this. Are you able to point out which lines this is happening on?

fwsmit: > Do you have an opinion on adding a vertical_padding setting? I would imagine the padding setting would then just serve as a default if either vertical or horizontal are unset

The padding variable _is_ the vertical padding currently. It just has an unfortunate name. I think it would not be too breaking to make it the general padding setting and create a separate vertical padding setting.

fwsmit: >   I wasn't able to reproduce this. Are you able to point out which lines this is happening on?

I might have been wrong on this one, I didn't test if it would happen. It's probably handled well already by just ignoring the icon padding when no icon is present.

m-bartlett: >The padding variable is the vertical padding currently. It just has an unfortunate name. I think it would not be too breaking to make it the general padding setting and create a separate vertical padding setting.

Would replacing accesses of `settings.h_padding` and `settings.padding` with these be a reasonable approach?

```C
static int get_horizontal_padding()
{
        if (settings.h_padding) {
                return settings.h_padding;
        } else {
                return settings.padding;
        }
}

static int get_vertical_padding()
{
        if (settings.v_padding) {
                return settings.v_padding;
        } else {
                return settings.padding;
        }
}
```

Alternatively I could make it so `settings.h_padding` or `settings.v_padding` are set to `settings.padding` if they are unset when the settings object gets created, so no check needs to be performed and everything can just use `settings.{h/v}_padding`

fwsmit: I think the first approach is best, since that would open the possibility of making the padding settings rules later. I would recommend you do this addition in a separate PR, though, to keep this one from getting too big.

src/draw.c
```diff
@@ -167,24 +167,6 @@ static struct color layout_get_sepcolor(struct colored_layout *cl,
         }
 }
 
-static int get_horizontal_text_icon_padding()
```
fwsmit: I think you should keep these functions. `text_icon_padding` may be 0, signalling that normal padding should be used.

m-bartlett: I see, I may have misinterpreted what you meant. Should I undo https://github.com/dunst-project/dunst/pull/985/commits/4bdf2aba4179e74c7470c5ee5eb639bb5e0c1854 where I removed the functions and changed the height calculations to just use `text_icon_padding`?

src/draw.c
```diff
@@ -399,17 +414,22 @@ static GSList *create_layouts(cairo_t *c)
...
-        get_text_size(cl->l, NULL, &h, scale);
+        get_text_size(cl->l, NULL, &h_text, scale);
         if (cl->icon)
                 h_icon = get_icon_height(cl->icon, scale);
         if (have_progress_bar(cl)){
                 h_progress_bar = settings.progress_bar_height + settings.padding;
         }
 
-        int res = MAX(h, h_icon) + h_progress_bar;
+        int res = MAX(h_text, h_icon) + h_progress_bar;
```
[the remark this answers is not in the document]
m-bartlett: Do you mean like this?
```C
int res;
if (cl->n->icon_position == ICON_TOP && cl->n->icon) {
        res = h_icon + h_text + h_progress_bar + vertical_padding;
} else {
        res = MAX(h_text, h_icon) + h_progress_bar;
}

return res;
```

src/draw.c
fwsmit: Yep, that's good. And then you should expand the check to choose between the `text_icon_padding` and horizontal/vertical vertical padding number. This check needs the icon position and if an icon exists. For the horizontal for example that would look something like:
```
static int get_horizontal_text_icon_padding(struct notifiction n)
{
        if (settings.text_icon_padding && n->icon && (n->icon_position == ICON_LEFT || n->icon_position == ICON_RIGHT)) {
                return settings.text_icon_padding;
        } else {
                return settings.h_padding;
        }
}
```

You may also check for this another way, but the functionality should be the same.

src/draw.c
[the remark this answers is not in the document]
fwsmit: Yes! If you like, you may also put the return statement inside the if-else statement. You don't need a res variable then.

docs/dunst.5.pod
```diff
@@ -364,7 +364,7 @@ Hide the count of stacked duplicate notifications.
 Show an indicator if a notification contains actions and/or open-able URLs. See
 ACTIONS below for further details.
 
-=item B<icon_position> (values: [left/right/off], default: off)
+=item B<icon_position> (values: [left/right/top/off], default: off)
```
fwsmit, line 367: It seems like this documentation was wrong. The default icon position was "left" before. Could you update this and set the default icon position to left in `notification.c` (`notification_create()`).

test/functional-tests/test.sh
```diff
@@ -188,6 +188,29 @@ function progress_bar {
     keypress
 }
 
+function icon_position {
+    for position in left top right off; do
+        tmp_dunstrc dunstrc.icon_position "icon_position = $position"    # use default icon theme
+        start_dunst dunstrc.tmp
+        for urgency in l n c; do
+            ../../dunstify -a "dunst tester" -u $urgency "icon_position = $position"
+        done
```
fwsmit, line 197: Maybe add a few tests for testing padding in combination with icon position. So these same tests, but then with all the paddings set to different values.

m-bartlett:
@fwsmit I believe I've made all requested changes, let me know what you think.

I consolidated the icon_position tests into one batch of notifications, and then added iterating different padding configurations for the icon_position test.

Here is each batch of icon_position tests rendered with the 4 different padding configurations now found in the icon_position test as rendered on my screen:

<span> no-padding horizontal-padding icon-padding vertical-padding </span>

fwsmit (approved):
Thank you, looks perfect. The functional tests are also very nice.

PR #997, opened 2021-12-06
Add support for config drop-ins

fwsmit:
This PR introduces support for drop-ins--those files (or snippets) usually residing in a *.d directory to override (some or all settings of) the main/base config in the parent directory.


fwsmit:
Merged it.

Kranzes:
It seems that drop-ins don't work when using --config to point to the dunstrc? (the drop-ins are inside dunstrc.d which is next to the dunstrc file).

fwsmit:
Yeah, that could be. If you have a use case for that to be different, please open an issue about it

PR #998, opened 2021-12-08
Fix dunstctl action dbus error

mark2185:
Calling `dunstctl action` after the `1.7.2` release would result in an error
```
Error org.freedesktop.DBus.Error.InvalidArgs: Type of message, “(u)”, does not match expected type “(i)”
Failed to communicate with dunst, is it running? Or maybe the version is outdated. You can try 'dunstctl debug' as a next debugging step.
```

This fixes it.

### DIFF
--- a/dunstctl
+++ b/dunstctl
@@ -56,7 +56,7 @@ command -v dbus-send >/dev/null 2>/dev/null || \
 
 case "${1:-}" in
 	"action")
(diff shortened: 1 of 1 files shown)


fwsmit:
I think I fixed this error in master already. Would you mind checking that?

mark2185:
Huh, indeed.

Sorry for the noise, then!

fwsmit:
No problem, thanks anyways. I released a new version with the fix.

Issue #992, opened 2021-12-01
[experimental icon lookup] Improve config options


fwsmit:
This only applies to the experimental icon lookup. This should be fixed before the experimental icon lookup is enabled by default.

Currently `icon_size` is only used as the default icon size for looking up icons in a theme. It does not affect the scaling of said icon or other icons.

There are 3 ways an icon can be passed to dunst:

### Via icon name
The icon is searched in the selected theme. The icon is only found if the `index.theme` says the icon is suitable for `icon_size`. The icon should then be scaled to that size (this is necessary when the icon is scalable).

### Via a path
Now dunst cannot choose what size icon is passed to it. In the old implementation the icon was scaled between `min_icon_size` and `max_icon_size`. If this is kept the same, there are 3 settings for icon sizes. We might just scale everything to `icon_size`, but that will perhaps result in some badly scaled icons. It will be a good starting point, though.

### Raw icon data via dbus
Configuration can be done the same as for icons passed via a path.

#### TODO
- [x] Scale icon to `icon_size` when it is imported from icon name.
- [x] Test if icons passed via path get scaled correctly.

imsamuka:
> Test if icons passed via path get scaled correctly.

This can be tested already? :confused:

fwsmit:
> > Test if icons passed via path get scaled correctly.
> 
> This can be tested already? :confused: 

It should work as intended now. All icons are scaled to `icon_size` iirc. 


> > We might just scale everything to icon_size, but that will perhaps result in some badly scaled icons.
> 
> Maybe we could create a new global option `scale_non_svg`, that disable and enable scaling for other formats. That could be used on a application/match basis too. This way we can remove `min_icon_size` and `max_icon_size`, right?

Yeah that's a good idea.

imsamuka:
Raw icon data is supposed to scale as well when using `scale_non_svg`, right? The notifications from `mpDris2` doesn't respect `icon_size` yet, only `max_icon_size`.

Changing `icon_size` on matches/applications is working well.

I took a while to figure out why a icon wasn't scaling up, but i got it: the file `.../96x96/apps/enjoy-music-player.svg` exists and is used, but it is actually a 48x48 :face_exhaling:. Since it's a SVG, dunst should scale these up too, i think?

fwsmit:
> Raw icon data is supposed to scale as well when using `scale_non_svg`, right? The notifications from `mpDris2` doesn't respect `icon_size` yet, only `max_icon_size`.

[...]

PR #1003, opened 2021-12-25
Always scale icons to `icon_size`

fwsmit:
This PR fixes two cases where the resulting icon size would not be the same as `icon_size`.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -584,16 +584,6 @@ static struct notification *dbus_message_to_notification(const gchar *sender, GV
                 g_variant_unref(dict_value);
         }
 
-        dict_value = g_variant_lookup_value(hints, "image-data", G_VARIANT_TYPE("(iiibiiay)"));
-        if (!dict_value)
-                dict_value = g_variant_lookup_value(hints, "image_data", G_VARIANT_TYPE("(iiibiiay)"));
-        if (!dict_value)
-                dict_value = g_variant_lookup_value(hints, "icon_data", G_VARIANT_TYPE("(iiibiiay)"));
-        if (dict_value) {
-                notification_icon_replace_data(n, dict_value);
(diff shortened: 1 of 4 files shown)


fwsmit:
Hi @owl4ce,

It's not entirely clear to me what's the desired behaviour. In #992 I mention all the possible ways an icon can be passed to dunst. Could you tell me how you would like each of them to behave? Note that the idea was to remove min_icon_size and max_icon_size to simplify the configuration. I didn't want to have multiple icon size config variables. Instead I can add new configuration variables like the suggested `scale_non_svg` to disable scaling for non-scalable icons.

So for 
- Pass via icon name
- Pass via a path
- Pass raw icon data via dbus

Currently they should all scale to `icon_size`.

ncfavier:
> Currently they should all scale to `icon_size`.

This is not the case. On 1.8.0, with the following minimal config:

```
[global]
enable_recursive_icon_lookup=yes
icon_size=128
min_icon_size=128
max_icon_size=500
```

raw icons are scaled down to something like 40x40, while normal icons are 128x128. Doesn't happen if I remove `enable_recursive_icon_lookup`.

fwsmit:
> > Currently they should all scale to `icon_size`.
> 
> This is not the case. On 1.8.0, with the following minimal config:
> 
> ```
> [global]
> enable_recursive_icon_lookup=yes
> icon_size=128
> min_icon_size=128
> max_icon_size=500
> ```
> 
> raw icons are scaled down to something like 40x40, while normal icons are 128x128. Doesn't happen if I remove `enable_recursive_icon_lookup`.

Yep, that's a bug. They always get scaled to 32x32 to be specific. See #1043

ncfavier:
Ah so I guess the DPI scaling makes them 40x40

PR #1004, opened 2021-12-27
Fix default `format` value in documentation

Profpatsch:
It was last changed in https://github.com/dunst-project/dunst/commit/a7b65b2e7b4919039d687232c3cd10f533c256b0 but the documentation was never updated.


fwsmit:
Thanks for the fix

Issue #1013, opened 2022-01-12
Support regex matching in rules


levouh:
Is there a better way to ignore things based on `summary` for a given `appname` (yes I am aware that I should use `desktop_entry` instead) than:

```ini
[kdec_pushover_ignore]
    appname = "KDE Connect"
    summary = "Pushover"
    format = ""

[kdec_gmail_ignore]
    appname = "KDE Connect"
    summary = "Gmail"
    format = ""

[kdec_cal_ignore]
    appname = "KDE Connect"
    summary = "Calendar"
    format = ""

[kdec_ov_ignore]
    appname = "KDE Connect"
    summary = "Okta Verify"
    format = ""
```

For example, using a script based on the `appname` that just modifies the `format` like:

```ini
[kdec_ignore]
    appname = "KDE Connect"
    summary = "*"
    script = ignore.sh
```

where the script can use "better" pattern matching (basically just a bunch of `|` statements) and "return" `format = ""` (to ignore the notification) manually?

fwsmit:
You cannot use a script to do this, but you should be able to do it the way you suggest inside of dunst. Dunst uses [`fnmatch(3)`](https://man.archlinux.org/man/fnmatch.3) for pattern matching a list of patterns. The pattern: `summary = @(one|two|three)"`  matches any string that contains "one", "two" or "three".

This should probably get documented better.

levouh:
Definitely does, thanks.

levouh:
@fwsmit Given that my notifications are coming through as:

```
appname=KDE Connect
summary=Gmail
```

which is captured using a script to just log every notification (skipping something like running `dunst` with `-print` mostly just to check the rule is happening), my rule is:
```
[kdec_ignore]
    appname = "KDE Connect"
    summary = "@(Pushover|Gmail|Calendar)"
    format = ""
```

and things are not being ignored as I would expect. Is the format there correct? I will modify the "logging" to check trailing whitespace, but I would expect these to be ignored?

I run `dunst` as a user service, and am restarting it to reload the configuration via `systemctl --user restart dunst`. I have also verified that once I kill it, there are no other instances running (and it is loading my other configuration so I know the configuration file is being read).

fwsmit:
Whoops, read the fnmatch documentation incorrectly. Dunst calls fnmatch without flags, but you need the FNM_EXTMATCH flag to have the behavior. It's a gnu extension, so I'll have to look to see how we can enable it on certain platforms.

[...]

PR #1017, opened 2022-01-15
rules: implement regex matching

fwsmit:
This enables more advanced regular expression syntax. It can be enabled with `enable_posix_regex`.

See the man page for more information.

Fixes: #1013
Fixes: #645
Fixes: #658

Related: #659

### DIFF
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -58,6 +58,22 @@ keyboard focus.
 
 =back
 
+=item B<enable_posix_regex> (default: false)
+
+When set to true (recommended), you can use POSIX regular expressions for
+filtering rules. It uses the POSIX Extended Regular Expression syntax:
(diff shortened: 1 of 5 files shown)


docs/dunst.5.pod
```diff
@@ -58,6 +58,18 @@ keyboard focus.
 
 =back
 
+=item B<enable_posix_regex> (default: false)
+
+When set to true, you can use POSIX regular expressions for filtering rules. It
+uses the POSIX Extended Regular Expression syntax:
+https://en.m.wikibooks.org/wiki/Regular_Expressions/POSIX-Extended_Regular_Expressions.
+
+If this is set to false, dunst will us fnmatch(3) for matching strings.
```
levouh, line 67: _nit_: _use_

Might be worth noting that `dunst` passes no (?) extra flags to `fnmatch` as well?

fwsmit: Thanks, I'll add both your suggestions.

PR #1022, opened 2022-01-23
fix: unable to override timeout

### LINKED ISSUE #1020: Can't override timeout if program specified it; can't override urgency
### Issue description

I'm trying to override the timeout for some notifications. It does not work when the application specified a timeout. I also discovered that I can not override the urgency.

* `notify-send -a "program" "summary" "body"` — timeout override works, urgency override does not work.
* `notify-send -a "program" -t 10000 "summary" "body"` — timeout override does not work, urgency override does not work.

#### Example notification

This is the output I get with `notify-send -a "program" -t 10000 "summary" "body"` The urgency seems to be set correctly, but it is displayed as a normal notidication.

```
{
        appname: 'program'
        summary: 'summary'
        body: 'body'
        icon: 'dialog-warning'
        raw_icon set: false
        icon_id: '(null)'
        desktop_entry: ''
        category:
        timeout: 10000
        urgency: CRITICAL
        transient: 0
        formatted: '<b>summary</b>
body'
        fg: #ffffff
        bg: #00ff00
        highlight: #1745d1
        frame: #888888
        fullscreen: show
        format: <b>%s</b>
%b
        progress: -1
        stack_tag:
        id: 2
        actions: {}
        script_count: 0
}
```

### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon 1.7.3 (2021-12-08)`
- Install type: `Gentoo repo`

<details>

```ini
[urgency_normal]
    background  = "#00ff00"
    timeout = 10

[urgency_critical]
    background  = "#ff0000"
    timeout = 0

[override_test]
    appname = "program"
    timeout = 2
    urgency = critical
```
</details>

rpbranco:
Fixes: #1020

### DIFF
--- a/test/dbus.c
+++ b/test/dbus.c
@@ -654,10 +654,18 @@ TEST test_timeout_overflow(void)
         n_dbus = dbus_notification_new();
         n_dbus->app_name = "dunstteststack";
         n_dbus->app_icon = "NONE";
-        n_dbus->summary = "test_hint_urgency";
+        n_dbus->summary = "test_timeout_overflow";
         n_dbus->body = "Summary of it";
         n_dbus->expire_timeout = 2147484;
         gint64 expected_timeout = G_GINT64_CONSTANT(2147484000);
+
+        // Disable urgency_normal rule in order to avoid timeout override
+        struct rule *urgency_normal_rule = get_rule("urgency_normal");
+        bool previous_urgency_normal_rule_state = false;
+        if (urgency_normal_rule) {
+                previous_urgency_normal_rule_state = urgency_normal_rule->enabled;
+                urgency_normal_rule->enabled = false;
+        }
 
         guint id;
         ASSERT(dbus_notification_fire(n_dbus, &id));
@@ -665,6 +673,10 @@ TEST test_timeout_overflow(void)
 
         n = queues_debug_find_notification_by_id(id);
         ASSERT_EQ_FMT(expected_timeout, n->timeout, "%" G_GINT64_FORMAT);
+
+        if (urgency_normal_rule) {
+                urgency_normal_rule->enabled = previous_urgency_normal_rule_state;
+        }
 
         dbus_notification_free(n_dbus);
 
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -590,6 +590,9 @@ static struct notification *dbus_message_to_notification(const gchar *sender, GV
                 g_variant_unref(dict_value);
         }
 
+        if (timeout >= 0)
+                n->timeout = ((gint64)timeout) * 1000;
+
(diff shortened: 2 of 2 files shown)


tastytea:
Thanks! I can can confirm the timeout override works with this patch.

But when I override the urgency to be critical only the icon changes to the one used for critical notifications, the colors are the same as for normal ones.

rpbranco:
No worries! I hope it was helpful :smile:. 

Regarding your problem with the notification colour, the [documentation](https://dunst-project.org/documentation/) states:

> `urgency`
> This sets the notification urgency.
> 
> IMPORTANT NOTE: This currently DOES NOT re-apply the attributes from the urgency_* sections. The changed urgency will only be visible in rules defined later. Use msg_urgency to match it.

This leads me to believe that this is not a bug but intended behaviour.
So, in your example, you can have the proper notification colours if you swap the order of the sections.

Example:
```
[override_test]
    appname = "program"
    timeout = 2
    urgency = critical

[urgency_normal]
    background  = "#00ff00"
    timeout = 10

[urgency_critical]
    background  = "#ff0000"
    timeout = 0
```

tastytea:
> So, in your example, you can have the proper notification colours if you swap the order of the sections.

Ah right, that works. 😊

fwsmit:
Thanks for the fix. I can confirm that it's indeed intended behaviour that the rule order matters.

test/dbus.c
```diff
@@ -654,18 +654,30 @@ TEST test_timeout_overflow(void)
         n_dbus = dbus_notification_new();
         n_dbus->app_name = "dunstteststack";
         n_dbus->app_icon = "NONE";
-        n_dbus->summary = "test_hint_urgency";
+        n_dbus->summary = "test_timeout_overflow";
         n_dbus->body = "Summary of it";
         n_dbus->expire_timeout = 2147484;
         gint64 expected_timeout = G_GINT64_CONSTANT(2147484000);
 
+        // Disable urgency_normal rule in order to avoid timeout override
```
fwsmit, line 662: This shouldn't be needed right? The dbus timeout should overwrite the rule.

rpbranco: > This shouldn't be needed right?

I did not know that the dbus timeout should have priority over the ones defined in the rules.

The change I made sets the notification timeout before applying the rules.
This gives priority to the timeouts defined in rules and not the one in the dbus.
That is why I had to change the test.

> The dbus timeout should overwrite the rule.

This makes it impossible for users to change a dbus timeout right?

fwsmit:
As I already said earlier, the current behaviour is intended. If there is anything you want to be changed, please discuss it in #1020. I'll close this MR for now.

PR #1027, opened 2022-01-30
Allow negative offsets (fixes issue #1026)

### LINKED ISSUE #1026: Regression: Negative offset
I used to use negative offset to cut 6 pixels off dunst notifications so they don't have rounded corners (thanks to picom) at the edge of the screen because that looks ugly. It was also getting rid of the right border edge which is good for the looks too. This is the geometry I used to use and the notifications would pop up along the right edge of my screen:

        geometry = "350x5--6+200"

I just updated dunst and noticed that I can't set a negative offset anymore since the `offset` parameter only accepts non-negative numbers. So whether I write `offset = -6x200` or `offset = 6x200`, I get the same result which is ugly.

Is there any chance to add that feature back? I'll try to patch it in myself, but I'm a bit rusty at C.

kurogetsusai:
Not sure if you want that feature in dunst or not, but this PR fixes regression introduced in 1a456a56 which made it impossible to use negative offsets (issue #1026). I personally think it's useful. Tested on X11 only. I don't have Wayland around, sorry.

fixes #1026

### DIFF
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -136,8 +136,7 @@ Origin can be one of:
 =item B<offset> format: (horizontal, vertical)
 
 Respectively the horizontal and vertical offset in pixels from the corner
-of the screen specified by B<origin>. Both values should always be positive or
-zero.
+of the screen specified by B<origin>.
 
 Examples:
         origin = top-right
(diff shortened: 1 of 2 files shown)


fwsmit:
I've tested it on wayland and it works as well. Might be good to say in the docs that a negative offset will lead to the notification being off screen.

kurogetsusai:
Thanks for testing it on Wayland! I'll update the documentation as soon as I get to my laptop. Sorry for the delay.

kurogetsusai:
I updated the docs. Let me know if there's anything else you'd like me to change.

fwsmit:
Thanks for the update, I'll go ahead and merge it

PR #1030, opened 2022-02-10
Add ignore_close action to rules

mhcerri:
Add a new action to the rules that will prevent the client that create
the notification from closing it.

One practical example for this new option is Mattermost, which
automatically closes its notifications after a few seconds. With this
option, it's still possible to have persistent notifications even when
the requesting application tries to dismiss it.


mhcerri:
Although it's not the same, I will test if `ignore_dbusclose` covers my use case.

Issue #1035, opened 2022-02-16
Add `override_timeout` setting, which overrides the timeout set by dbus.


fwsmit:
Some programs set a timeout via dbus which is not always desirable. Add a setting that overrides this timeout. Note that the normal `timeout` setting shouldn't be changed.

Ref: #1020

bynect:
can this be closed?

fwsmit:
Yes

PR #1038, opened 2022-02-19
Rule attribute to override timeout set by dbus

rpbranco:
This adds a new (modifying) rule attribute `override_dbus_timeout`, as requested in #1035.

The timeout priority is now:
1. Rule `override_dbus_timeout`
2. Dbus timeout
3. Rule `timeout`
  (4. Internal defaults)


src/dbus.c
```diff
@@ -604,6 +604,9 @@ static struct notification *dbus_message_to_notification(const gchar *sender, GV
                 g_variant_unref(dict_value);
         }
 
+        // Set the dbus timeout
+        n->dbus_timeout = timeout < 0 ? timeout : ((gint64)timeout) * 1000;
```
fwsmit, line 608: Instead of storing a separate dbus_timeout variable, I'd just override the timeout.

src/rules.h
```diff
@@ -29,6 +29,7 @@ struct rule {
 
         /* modifying */
         gint64 timeout; // this has to be the first modifying rule
+        gint64 override_dbus_timeout;
```
fwsmit, line 32: You can then make this an `override_timeout` variable (set by the `override_dbus_timeout` setting). That can be used in `notification_init` after the rule processing to override the `timeout`.

fwsmit:
Thanks for improving your PR. I think this is the right approach. The only comment I have is that the logic can be made a bit simpler. See the inline comments

rpbranco:
Thank you for the fast reply! Regarding your suggestions, and before implementing them, how would you feel if I were to implement a rule attribute to allow filtering based on the dbus timeout?

I ask this before making any changes because, in order to implement this, it would be necessary to know the dbus timeout when applying the rules.

fwsmit:
> Thank you for the fast reply! Regarding your suggestions, and before implementing them, how would you feel if I were to implement a rule attribute to allow filtering based on the dbus timeout?
> 
> I ask this before making any changes because, in order to implement this, it would be necessary to know the dbus timeout when applying the rules.

If that's the case, then yes don't remove it. Matching on dbus timeout would be great, although maybe not very useful on it's own. What's your use case? Maybe it's more useful in combination with #1040 to allow for `match_dbus_timeout =! -1` or something.

src/dbus.c
fwsmit:
```suggestion
if (timeout >= 0)
        n->dbus_timeout =((gint64)timeout) * 1000;
```

If you want to use the dbus_timeout in the future, I'd still suggestion doing this.
Also, I think it's better to modify n->timeout in notification.c then.

rpbranco:
I made the suggested changes.

Regarding my use case for the `match_dbus_timeout`, I have an application that sets the dbus timeout of some notifications to `0`. I would like to be able to match only these notifications.

Sorry for the delayed answer!

fwsmit:
> I made the suggested changes.
> 
> Regarding my use case for the `match_dbus_timeout`, I have an application that sets the dbus timeout of some notifications to `0`. I would like to be able to match only these notifications.
> 
> Sorry for the delayed answer!

Great, thank you. The changes look good, so I'll merge it

PR #1039, opened 2022-02-20
Scale offset according to DPI

livanh:
Currently the notification size changes according to the "scale" option, but the offset does not. I believe it makes more sense to scale them together

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -810,12 +810,12 @@ void calc_window_pos(const struct screen_info *scr, int width, int height, int *
                 case ORIGIN_TOP_LEFT:
                 case ORIGIN_LEFT_CENTER:
                 case ORIGIN_BOTTOM_LEFT:
-                        *ret_x = scr->x + settings.offset.x;
+                        *ret_x = scr->x + round(settings.offset.x * draw_get_scale());
                         break;
                 case ORIGIN_TOP_RIGHT:
                 case ORIGIN_RIGHT_CENTER:
                 case ORIGIN_BOTTOM_RIGHT:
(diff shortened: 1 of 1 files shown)


fwsmit:
Thanks for the improvement. I agree that it's good to scale more things. Have you tested this on X11 or wayland?

livanh:
I tested it on X11. I'll test it on wayland as well if needed

fwsmit:
It does not have an effect on wayland. calc_window_pos is only used in X11. The current behavior in wayland is:

- When I set the scale to 1, the notification is offset X distance from the edge of the display.
- When I set the scale to 2, the notification is offset 2X distance from the edge of the display.
- etc.

Is this already the desired behaviour? What's the behaviour on X11? Can you maybe explain a bit why?

livanh:
Yes, the behaviour on wayland is already what I wanted to achieve. On X11, an offset of N was always interpreted as "N pixels", regardless of the scale. It would have to be "scale x N pixels" to be consistent with wayland.

As a concrete example, let's say I have a panel at the bottom of the screen that is 30 pixels high, and I want the notifications to appear just above it, without overlapping. I can set origin = bottom and offset = 0x30 and it works as intended. Now, if I configure everything to scale to 2X, the panel is 60 pixels high (correct), the notifications are twice as big (correct), but the offset is still 30 pixels (incorrect, because now the notifications cover the top half of the panel). With this change, the offset will also be scaled to 60 pixels.

I hope the explanation is clear. If you need it, I can improve it or provide screenshots.

fwsmit:
Thanks for the explanation @livanh. I get the need for this and it seems like a logical change (espcially so, since it's already like this on wayland). I'll go ahead an merge it!

Issue #1011, opened 2022-01-07
Margin / spacing / gaps between notifications?


Octelly:
Is it possible to add some spacing between notifications?
I tried producing the desired effect with ``separator_color = "#00000000"``, but that doesn't quite achieve what I want.
![image](https://user-images.githubusercontent.com/29976879/148532990-a70aa024-1422-4b17-a278-8b9a13097a65.png)
``wired-notify`` supports this via a gap config setting if I recall correctly. It is also a popular feature among different tiling window managers such as AwesomeWM (``beautiful.useless_gap = 5``) or QTile (layout property ``margin=5``).

fwsmit:
Needs #1018 before this can be implemented correctly.

paddyw2:
I have implemented this feature for myself quickly, you can view the commit/diff here: https://github.com/paddyw2/dunst/commit/436613f0975a5686bdc028f6b46bb61247af036b 

It allows me to set these values in the config file:

```
[global]
    gaps = true
    gap_size = 10
```

Happy to clean it up and submit a PR, but it sounds like there's some code base refactoring planned first. So in the meantime you might want to try out the edits above and just build and install locally.

fwsmit:
> 



> I have implemented this feature for myself quickly, you can view the commit/diff here: [paddyw2@436613f](https://github.com/paddyw2/dunst/commit/436613f0975a5686bdc028f6b46bb61247af036b)
> 
> It allows me to set these values in the config file:
> 
> ```
> [global]
>     gaps = true
>     gap_size = 10
> ```
> 
> Happy to clean it up and submit a PR, but it sounds like there's some code base refactoring planned first. So in the meantime you might want to try out the edits above and just build and install locally.

Cool. I'd be happy to accept it as a PR. It's a pretty small patch and when we implement multiple windows, it can be replaced in a backward compatible way. So I see no issues with adding it to master :)

fwsmit:
What's good to think about for the PR is mouse clicking. I suspect that clicks in between the notifications are getting picked up by dunst. On top of that, the notification height calculation might be off, leading to the wrong notification being closed on click. This calculation should be fixed and it's probably best to ignore clicks that are in between notifications, since that can only really be fixed by multiple windows AFAIK.

bynect:
I think this has been solved with the gaps option

PR #1053, opened 2022-03-20
Add support for notification gaps

paddyw2:
## Summary

Provide gap support by adding option to draw blank spaces between notifications.

Related issues:
 - https://github.com/dunst-project/dunst/issues/1011

To support this feature, three general changes were required.

The first increases the size of the total image surface used by cairo so that extra space required by gaps does not crop notifications. To calculate this, the gap size plus size of top and bottom border frames are summed for each non top or bottom notification and added to the total height. This is because normally only the top and bottom frame widths are shown, but with gaps they are shown for each notification.

The second increments the y offset by an additional value of the gap size and extra frame widths. When gaps are enabled, each notification is treated as both first and last to reuse top and bottom frame borders draw logic.

Lastly, the click handling logic is updated to correctly associate y coordinates with the new notification positions when gaps are enabled.

To use:

```
[global]
    gaps = true
    gap_size = 10
```

## Limitations/Potential Improvements

As mentioned in https://github.com/dunst-project/dunst/issues/1011 the gap areas are still drawn as "dead space" by cairo so any clicks on these areas do not pass down to applications below as a user may expect. Therefore this is only really suitable for small gap sizes where clicking between gaps to access applications below is not required. Additionally, like rounded corners and transparency, a compositor such as `picom` is required otherwise the gaps will show up as black space.

I tried adding some unit tests in `test/draw.c` to target the logic within `calculate_dimensions` and `layout_render` but open to feedback on better ways to do this. Same goes for a functional test/demo under `test/functional-tests/test.sh` that uses `test/functional-tests/dunstrc.gaps` where I also tried to added the option to test out click actions.

I'm new to this project so let me know if there's anything I can change or improve on :)


src/draw.c
```diff
@@ -501,7 +512,7 @@ void draw_rounded_rect(cairo_t *c, int x, int y, int width, int height, int corn
 
         cairo_new_sub_path(c);
 
-        if (last) {
+        if (last || settings.gaps) {
```
fwsmit, line 515: I think it's better to pass `first = true` and `last = true` to this function when gaps is enabled. This is true in the way that it's the last in a block of notifications.
The same counts for the other functions that get passed a first and last variable.

fwsmit (changes requested):
Thanks for putting in the work to submit this and also for creating tests to keep it functional in the future. I've added a few comments mostly to simplify the implementation. If they turn out to be more complicated, let me know and I'll take a look at it.

Could you add the new settings to `docs/dunst.5.pod` to include them in the man page?

Also, the tests are currently failing, because of a memory leak (might be in the test itself). Could you take a look at that as well. You can run the memleak test locally by running `make test-valgrind`.

src/draw.c
```diff
@@ -319,6 +321,15 @@ static struct dimensions calculate_dimensions(GSList *layouts)
                 dim.w = max_width;
         }
 
+        if (settings.gaps) {
+                // roughly 2x frame borders and 1x gap size extra height required
```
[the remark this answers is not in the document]
paddyw2: With the move to using `first` and `last` flags the logic has changed slightly, so I've removed the comment as I think the code is clear enough.

test/draw.c
```diff
@@ -71,6 +141,160 @@ TEST test_layout_from_notification_no_icon(void)
...
+
+        g_slist_free_full(layouts, free_colored_layout);
+        g_slist_free_full(notifications, free_dummy_notification);
+        cairo_surface_destroy(image_surface);
+        PASS();
+}
+
+TEST test_layout_render_gaps(void)
+{
+        set_small_max_height();
```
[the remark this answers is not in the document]
paddyw2: I added logic to restore any changed settings values within each test. While it's a bit repetitive I think keeping the restore logic at the same level as the settings changes helps readability and allows each test to change what it wants. If there is a simple way of completely cloning/restoring the default settings struct that might be better.

paddyw2:
Thanks for the feedback, it definitely helped clean it up. The valgrind errors are due to the tests and are caused by `pangocairo` internals. I had initially added `pango_cairo_font_map_set_default` call which appeared to resolve the issue on my distro but after the workflow ran I see it still persisted on others. I was able to recreate and (apparently) resolve also on Ubuntu by adding a call to `FcFini` from `fontconfig` (see https://github.com/paddyw2/dunst/blob/d7dbed5f3b94b8aa1e3ae6de33ca73e2d011b179/test/helpers.c#L70) but will have to see how that holds up when tested against others.

In addition to `dunst.5.pod` I also added some docs to `dunstrc`, which I'm guessing should also be updated to highlight defaults. While doing so it occurred to me that I could potentially just use a single settings value `gap_size` with a value of zero indicating gaps are disabled. Upside is less settings, but downside is the code may be _slightly_ less readable (`if (settings.gaps) {...}` vs. `if (settings.gap_size) {...}` or `if (settings.gap_size > 0) {...}`). It would be an easy change so let me know if you have any thoughts on that or anything else.

paddyw2:
Looks like the leak still occurs in at least `debian:stretch`. I was able to reproduce in a local debian stretch container but couldn't find anything else to try clean it up. My guess here is that it is maybe something to do with the fontconfig library age in each distro. If that's the case would adding a valgrind suppression be best? Something like:

```
{
   fontconfig-pango-leaks
   Memcheck:Leak
   fun:malloc
   ...
   fun:FcConfigSubstituteWithPat
   ...
   fun:pango_layout_get_pixel_extents
}
```

fwsmit:
Looks like the issue is on debian's part, so go ahead and add it to the suppressions.

test/helpers.c
```diff
@@ -67,4 +67,14 @@ struct notification *test_notification_with_icon(const char *name, gint64 timeou
         return n;
 }
 
+void pango_memory_cleanup()
```
fwsmit, line 70: Is this still needed with the suppression? I prefer not adding a fontconfig dependency for a test.

fwsmit:
> While doing so it occurred to me that I could potentially just use a single settings value `gap_size` with a value of zero indicating gaps are disabled.

Yeah seems like a good change.

paddyw2:
I added a suppression entry to target both `malloc` and `realloc` leaks coming from the `fontconfig` via `pango` and it seems to work on debian, ubuntu and arch. Let me know if you have any suggestions there. I also removed the `gaps` boolean setting and updated the tests and docs to match.

As a side note, only the notification height is clickable and not the frame/border, which I think mirrors the previous logic in `get_notification_at` from `src/input.c`. This can be seen with the functional "gaps" test where I used an exaggerated frame border size to highlight this. However, as `mouse_x` is not currently used in `input_handle_click`, any click along the x-axis including the frame/border will still register. Not a problem, just thought I'd mention this for completeness.

docs/dunst.5.pod
```diff
@@ -240,16 +240,12 @@ is equivalent to
 Defines width in pixels of frame around the notification window. Set to 0 to
 disable.
 
-=item B<gaps> (values: [true/false], default: false)
+=item B<gap_size> (default: 0)
 
-If set to true, display a gap of size B<gap_size> between notifications.
+Size of gap to display between notifications.
```
fwsmit, line 245: Maybe mention that when gap is 0, separator lines are used and when gap is bigger, a frame is used

fwsmit:
> As a side note, only the notification height is clickable and not the frame/border, which I think mirrors the previous logic in `get_notification_at` from `src/input.c`. This can be seen with the functional "gaps" test where I used an exaggerated frame border size to highlight this. However, as `mouse_x` is not currently used in `input_handle_click`, any click along the x-axis including the frame/border will still register. Not a problem, just thought I'd mention this for completeness.

Well spotted. The reason that frames are currently ignored is mostly because the separator lines don't really belong to either notification and frames were just an extension of that. On second thought, I think it's best to have the frames and separator lines be clickable as well. Half of the separator line can be for the top notification and half can be for the bottom. If you want you can implement that, otherwise I'll make an issue for that.

> I added a suppression entry to target both `malloc` and `realloc` leaks coming from the `fontconfig` via `pango` and it seems to work on debian, ubuntu and arch. Let me know if you have any suggestions there. I also removed the `gaps` boolean setting and updated the tests and docs to match.

Great!

paddyw2:
I updated the documentation and took a shot at updating the click handling logic for both gaps and separator lines, so let me know what you think. I added some tests in `test/input.c` but it may be easiest to see via the `gaps` and `separator_click` functional tests.

fwsmit:
Thank you for your excellent work. Everything seems to work as intended. I'll go ahead and merge this!

fwsmit (approved):

mrusme:
Any information on when this PR will make it into a release? :-)

mrusme:
I can see [a commit](https://github.com/dunst-project/dunst/commit/8f03f03914d7deee12058ba82670781dda6f8942#) for 1.9.0, dated for `2022-06-21`, yet no tag/release. :(

sollymay:
> I can see [a commit](https://github.com/dunst-project/dunst/commit/8f03f03914d7deee12058ba82670781dda6f8942#) for 1.9.0, dated for `2022-06-21`, yet no tag/release. :(

Was also curious about this one myself

fwsmit:
It is now released!

sollymay:
hey @fwsmit was this released for Fedora dnf yet? I updated the repos but still see 1.8.1-2 as the latest version:

``Last metadata expiration check: 0:00:58 ago on Mon 27 Jun 2022 11:33:54 PM EDT.
Package dunst-1.8.1-2.fc36.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!```

fwsmit:
That's up to the fedora maintainer.

PR #1060, opened 2022-03-30
Fixes dunstctl history command

stremlenye:
Removes `-l` option from the `busctl` invocation since it's only relevant to the `list` outputs.
That fixes the command to fail on invocation with an internal error
```
busctl: invalid option -- 'l'
```
See [buctl(1)](https://www.freedesktop.org/software/systemd/man/busctl.html)

### DIFF
--- a/dunstctl
+++ b/dunstctl
@@ -138,7 +138,7 @@ case "${1:-}" in
 			|| die "Dunst controlling interface not available. Is the version too old?"
 		;;
 	"history")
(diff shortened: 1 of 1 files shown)


fwsmit:
Weird that it didn't give that error earlier

bebehei (approved):

fwsmit:
For me the dunstctl history command works, but it seems like removing the -l can't hurt, so I'll merge this PR. Thank you!

PR #1062, opened 2022-04-07
follow=keyboard: Fix regression where we don't fall back to mouse

cdown:
Commit e8fc45da6371 ("follow=keyboard: Fall back to follow=mouse instead
of XDefaultScreen()", PR#708) introduced the behaviour that we fall
back to follow=mouse if there is no focused client in order to
accommodate for window managers where sitting on the root window is
normalised, like dwm.

Commmit ebcd20d8d040 ("Fix process of gettign the active monitor",
PR#809) added support for multiple X screens on one display. However, it
broke the functionality introduced in the previous PR, because
explicitly focusing the root window results in XGetInputFocus()
returning the root window, not PointerRoot.

Fix this by explicitly checking if the focused window is the root
window.

### DIFF
--- a/src/x11/screen.c
+++ b/src/x11/screen.c
@@ -389,12 +389,12 @@ sc_cleanup:
  */
 static Window get_focused_window(void)
 {
-        Window focused;
+        Window focused, root = RootWindow(xctx.dpy, DefaultScreen(xctx.dpy));
         int ignored;
 
         XGetInputFocus(xctx.dpy, &focused, &ignored);
 
-        if (focused == None || focused == PointerRoot)
+        if (focused == None || focused == PointerRoot || focused == root)
                 focused = 0;
         return focused;
 }


fwsmit:
Thank you for your contribution. It'd be nice if @mcz could check if this doesn't regress their fixes. Otherwise, I'll take a look next week to review the code better.

mcz:
Doesn't look like it would break my patch.
I think that technically it would be correct to check if the focused window is *any* root window (each screen has its own) by checking against the ```RootWindow()``` for every screen that exist on the opened display (from ```0``` to ```ScreenCount() - 1```). Not sure if that would be worth it since true multi-screen setups are so rare (with xinerama and xrandr being dominant). In my opinion doing so would only make sense as part of a big change separating multi-monitor (xinerama/xrandr) from multi-screen logic, which once again would a lot of work benefiting basically no one.

fwsmit:
Thanks for reviewing mcz and thanks for the patch cdown. I'll merge this now.

PR #1070, opened 2022-04-26
Icon: handle scaling when recursive lookup is enabled

vincentbernat:
This is a naive attempt done by selecting a bigger icon. Maybe I am
missing something.

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -337,7 +337,7 @@ void notification_icon_replace_path(struct notification *n, const char *new_icon
         g_clear_pointer(&n->icon_id, g_free);
 
         g_free(n->icon_path);
-        n->icon_path = get_path_from_icon_name(new_icon, n->icon_size);
+        n->icon_path = get_path_from_icon_name(new_icon, n->icon_size * draw_get_scale());
         if (n->icon_path) {
                 GdkPixbuf *pixbuf = get_pixbuf_from_file(n->icon_path,
                                 n->icon_size, draw_get_scale());


vincentbernat:
I was using my own more complex patchset until recently. However, after upgrading to 1.8.1, I noticed this could be done this way. What do you think?

fwsmit:
This is a bit more complicated I'm afraid. For some themes this might work, but for some themes and scales this might not.

I'll give an example:
Scale 2, icon size 48 -> searching for icon of size 96.
This will probably work, since most themes have icons of this size.

Scale 1.5, icon size 48 -> searching for icon size 72.
This might be problematic, since some themes won't have icons of this size.

The way the searching works is that it only searches for exact matches, as defined by index.theme. So if the index.theme say the 48x48 icons will only work for 48x48, the search will not return that icon when searching for 49x49. There is no fallback, so no icon would be found for that size. Many themes do include scalable icons which can be used for those odd sizes, but we shouldn't rely on that.

What could work is looking for the scaled size and if that doesn't return anything, use the unscaled size.

To be more correct, we should actually also make use of the special `x2` icons that some themes have. These are less detailed than the non-x2 icons, since they are meant for hi-dpi displays. So a `48x48@x2` icon has 96 pixels, but the detail of a 48 pixel icon. But since dunst also supports fractional scaling you will have to account for that.

So if you want, you can try the approach of looking for a scaled icon and if that cannot be found an unscaled one. I would implement that as a function similar to `get_path_from_icon_name` that can take a scale value. It can probably be done more efficiently than searching twice by changing `get_path_from_icon_name`.

vincentbernat:
I had another more complex proposition in #864 (https://github.com/dunst-project/dunst/commit/dc7e456f034e106bd1ab5f1900b2c58f913ba9fc) which was working fine for me. I was going to reimplement it on top of current master but I am lacking time for that. At the time, you already mentioned the 48x48@2x example. I was going to say I have no such icons, but just checking now and I see Papirus is using that.

Should I close this PR or do you want to keep it as a reminder? I won't have time to work on it as this quick workaround makes it work for me.

fwsmit:
I can take a look at the icon scaling soon. Could you make an issue with your needs for the icon scaling explained?

Issue #1076, opened 2022-05-17
Allow customization of SVG fill color


UtkarshVerma:
### Issue description
I use solid SVG icons for my notifications like those from FontAwesome. I wish to have the SVG colored according to the `foreground` colour defined in `dunstrc` as well.

![image](https://user-images.githubusercontent.com/31820255/168771175-6902b916-df97-4f57-84f9-655846092268.png)

Currently, the icon is displayed in black while the text is coloured.

### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon 1.8.1 (2022-03-02)`
- Install type: `package`
- Window manager / Desktop environment: `dwm`

fwsmit:
Do you know how this could be achieved?

UtkarshVerma:
I don't know how dunst works, but the way I was thinking of implementing this was editing all my SVG icons so that they point to the color `currentColor` instead of a specific value.

Therefore, `<svg fill="currentColor" ...>...</svg>` would do the job. This doesn't work because the SVG render doesn't take this `currentColor` into account. If we could have the SVG renderer define `currentColor` equal to the foreground color, that would be enough.

fwsmit:
Yeah I'm not sure if the svg renderer we use supports setting colours.

UtkarshVerma:
> Yeah I'm not sure if the svg renderer we use supports setting colours.

What if we handle SVGs differently? We can parse SVGs as plain text, do a simple string substitution for "currentColor" -> "foreground", and then pass this modified string to the renderer?

HashirGJ8842:
What is the status of this feature? Can i look into it?

fwsmit:
There was a PR, #1081, but it wasn't finished. Feel free to look into in and perhaps continue the PR. You can also start a new one if you like

HashirGJ8842:
Okay, thanks 👍

bynect:
Note: I liked the implementation of #1081, but I think that it is better to implement a more general solution like `svg_style` for including a stylesheet string into the svg (for more adhoc styles)

PR #1081, opened 2022-06-06
Implement SVG coloring

UtkarshVerma:
This PR aims to solve #1076. It does so by doing a simple string substitution of 'currentColor' with the raised notification's foreground colour.

A better way would have been to let the SVG renderer (librsvg) take care of this but it [doesn't support](https://gitlab.gnome.org/GNOME/librsvg/-/blob/main/FEATURES.md) specifying a custom 'currentColor' through the API AFAIK.

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -52,6 +52,7 @@ LDFLAGS_DEBUG  :=
 pkg_config_packs := gio-2.0 \
                     gdk-pixbuf-2.0 \
                     "glib-2.0 >= 2.44" \
+                    librsvg-2.0 \
                     pangocairo \
                     x11 \
                     xinerama \
--- a/src/icon.c
+++ b/src/icon.c
@@ -3,6 +3,7 @@
 #include <assert.h>
 #include <cairo.h>
 #include <gdk-pixbuf/gdk-pixbuf.h>
+#include <librsvg/rsvg.h>
 #include <stdbool.h>
 #include <string.h>
 #include <math.h>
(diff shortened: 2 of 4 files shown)


fwsmit:
It seems like librsvg supports setting current color: https://gitlab.gnome.org/GNOME/librsvg/-/issues/379

UtkarshVerma:
> It seems like librsvg supports setting current color: https://gitlab.gnome.org/GNOME/librsvg/-/issues/379

I saw that issue, but couldn't find any function in the API which allowed doing this.

progandy:
> > It seems like librsvg supports setting current color: https://gitlab.gnome.org/GNOME/librsvg/-/issues/379
> 
> I saw that issue, but couldn't find any function in the API which allowed doing this.

I'd guess it should be something like 
```rsvg_handle_set_stylesheet(..., "svg {color: $forgegroundcolor;}", ...)```

fwsmit:
The suggestion by progandy seems good. @UtkarshVerma, could you take a look to see if it's possible to implement it?

fwsmit:
@UtkarshVerma It's been a while. Would you like to finish this PR?

UtkarshVerma:
@fwsmit Unfortunately, I'm too caught up with work. It would be great if someone else picked this PR up.

fwsmit:
Allright, let me know if anyone wants to pick this up

UtkarshVerma:
@fwsmit I have implemented this feature. Please let me know what you think. I need your suggestions on the following:
- Can we make the SVG stylesheet user-configurable? We can make it default to `svg: { color: %s; }`, but if the user has icon-themes which don't depend on `currentColor`, then they can override it to more powerful rules like `path { fill: %s !important; }`.
- I think the SVG is being scaled differently than the PNG icons. It would be great if you could chime in on this.

fwsmit:
Thanks for implementing!

> * Can we make the SVG stylesheet user-configurable? We can make it default to `svg: { color: %s; }`, but if the user has icon-themes which don't depend on `currentColor`, then they can override it to more powerful rules like `path { fill: %s !important; }`.

If that's neccesary, then yes. But I'd suggest to first make the simplest version of styling and later allow for more complicated configuration (while not breaking the simpler one). I think it's most valuable to have a simple way to configure the colours and only require more complicated configuration for more complicated cases.

> * I think the SVG is being scaled differently than the PNG icons. It would be great if you could chime in on this.

I did some work on the icon lookup. I explained most of it while implementing. Take a look at #992 for example.

It depends on how you pass the icon to dunst. If it's via an icon theme, the minimum and maximum size specified by the theme are used. There were bugs where SVG icons were too big, but that should be fixed with #1003 (note that the configuration is now once again using `min_icon_size` and `max_icon_size` and not `icon_size`.

If you have any specific questions, do let me know, as it's quite complicated.

fwsmit:
We'll need to add librsvg to the docker image for the tests to work. Also, are you able to add some tests to test these changes?

UtkarshVerma:
SVG scaling was an issue in the older code. I've fixed it for now by switching to cairo-based rendering. Time is still a constraint for me, so I would really appreciate if someone could pick up this PR and iron out the few remaining details.

fwsmit:
> SVG scaling was an issue in the older code.

Could you elaborate on what issue there was with SVG scaling?


> Time is still a constraint for me, so I would really appreciate if someone could pick up this PR and iron out the few remaining details.

I'm not active here anymore either. So I've added the help wanted tag.

fwsmit:
What still needs to be done? I can think of the following:

- Add librsvg to the CI images (https://github.com/dunst-project/docker-images). I'm willing to help with that.
- Add tests
- Do manual testing

If that's done, I will review the code more thoroughly. Haven't found any issues with a quick scan.

UtkarshVerma:
You've covered pretty much everything. The other thing which is remaining is adding a configuration option for specifying the SVG stylesheet.

bynect:
Hello, I encountered the problem of styling icons, so I am willing to pick this pr up.
If you could please specify the unfinished tasks I'll try taking care of them.

fwsmit:

PR #1107, opened 2022-09-05
zsh completion script for dunstctl

elig0n:
includes commands completion and command options completion such as: rules (requires gAWK), history IDs (requires jq) and various toggle options


contrib/_dunstctl
```diff
@@ -0,0 +1,98 @@
+#compdef _dunstctl dunstctl
+# Depends on: gAWK (rule), jq (history-pop)
+
+local curcontext="$curcontext" ret=1
+local -a state line subs
+
+if [[ -v XDG_CONFIG_HOME || -n "$XDG_CONFIG_HOME" ]]; then
+  local DUNSTRC="$XDG_CONFIG_HOME/dunst/dunstrc"
```
bebehei, line 8: Just use this single line as a replacement for the whole if-clause. This has automatic POSIX parameter substitution.

```suggestion
  local DUNSTRC="${XDG_CONFIG_HOME:-$HOME/.config}/dunst/dunstrc"
```

Alternatively, if you want to keep the if clause for readability, please rewrite the condition. One should not do double pipes inside a test.

One alternative would be:

```zsh
if [[ -v XDG_CONFIG_HOME ]] || [[ -n "$XDG_CONFIG_HOME" ]]; then
```

Then not about the style: what should the `-v` check actually fulfil? `-n` should be enough and also enforce a non-empty variable. What would you do with an empty variable?

bebehei (commented):
Nice to see a PR regarding zsh completion

fwsmit:
Looks great. Thanks!

PR #1108, opened 2022-09-06
zsh completion script for dunst

elig0n:


fwsmit:
Thanks for creating this. Could the file name maybe be a bit more descriptive? A comment inside the file would also be helpful. This way people can know what it's about.

fwsmit:
The same counts for your other PR

PR #1119, opened 2022-10-28
Added rounded corners option for progress bar

ghost:
This PR adds the option `progress_bar_corner_radius` to enable rounded corners on the progress bar. If its value is non-zero, `draw_rounded_rect` is used in place of `draw_rect` to render the bar. Please let me know if anything needs to be changed!

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -497,12 +497,12 @@ static int frame_internal_radius (int r, int w, int h)
  * The top corners will get rounded by `corner_radius`, if `first` is set.
  * Respectably the same for `last` with the bottom corners.
  */
-void draw_rounded_rect(cairo_t *c, int x, int y, int width, int height, int corner_radius, double scale, bool first, bool last)
+void draw_rounded_rect(cairo_t *c, float x, float y, int width, int height, int corner_radius, double scale, bool first, bool last)
 {
         width = round(width * scale);
         height = round(height * scale);
-        x = round(x * scale);
-        y = round(y * scale);
+        x *= scale;
+        y *= scale;
         corner_radius = round(corner_radius * scale);
 
         const double degrees = M_PI / 180.0;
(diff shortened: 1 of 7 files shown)


src/settings.h
```diff
@@ -144,9 +144,10 @@ struct settings {
         enum mouse_action *mouse_middle_click;
         enum mouse_action *mouse_right_click;
         int progress_bar_height;
-        int progress_bar_min_width;
+        int progress_bar_min_width;
```
fwsmit, line 147: Unnecesary spacing added

src/draw.c
```diff
@@ -742,26 +742,48 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, dou
 
                 // draw progress bar
                 // Note: the bar could be drawn a bit smaller, because the frame is drawn on top
-                // left side
+                // left side (fill)
                 cairo_set_source_rgba(c, cl->highlight.r, cl->highlight.g, cl->highlight.b, cl->highlight.a);
-                draw_rect(c, x_bar_1, frame_y, progress_width_1, progress_height, scale);
+                if(settings.progress_bar_corner_radius > 0) {
```
fwsmit, line 747: I think the checks for the corner radius are unnecesary. In the other places we use draw_rounded_rect we don't check for that either. For optmization purposes, the check could be added in `draw_rounded_rect` itself if we want.
So I'd suggest removing the check here at least.

src/draw.c
```diff
@@ -742,26 +742,48 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, dou
...
-                cairo_rectangle(c,
-                                (frame_x + half_frame_width) * scale,
-                                (frame_y + half_frame_width) * scale,
-                                (progress_width - frame_width) * scale,
-                                progress_height * scale);
                 cairo_set_line_width(c, frame_width * scale);
-                cairo_stroke(c);
+                if(settings.progress_bar_corner_radius > 0) {
+                        draw_rounded_rect(c,
+                                        (frame_x + half_frame_width),
```
fwsmit, line 769: The brackets are not needed anymore

src/draw.c
```diff
@@ -742,26 +742,48 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, dou
...
+                                        (frame_y + half_frame_width),
+                                        (progress_width - frame_width),
+                                        progress_height,
+                                        settings.progress_bar_corner_radius,
+                                        scale, true, true);
+                        cairo_stroke(c);
+                } else {
+                        // TODO draw_rect instead of cairo_rectangle resulted
```
fwsmit, line 777: This TODO explains it a bit, but for the drawing of the rectangle, you need a non-integer position. The position gives the center of the line. So if you want a horizontal line of 1 pixel width at height = 100 to not be blurry, you need to give a y-coordinate of 100.5.
I think to solve this, you will need to change the position arguments of draw_rounded_rect to floats.

src/settings.c
```diff
@@ -201,6 +201,9 @@ void check_and_correct_settings(struct settings *s) {
                 if (s->progress_bar_min_width > s->width.max) {
                         LOG_W("Progress bar min width is greater than the max width of the notification.");
                 }
+                if (s->progress_bar_corner_radius > 100) {
```
fwsmit, line 204: The radius should be clamped to half the progress bar height, since that's the maximum that allows for a smooth circle.
This check currently also doesn't change the value itself.

fwsmit (commented):
Thank you for making this PR. It looks pretty good. I've added a few comments. The main thing is that you need to check if the lines of the progress bar are not blurry. This is because of the way cairo draws lines.

src/draw.c
vlfldr: Thanks for all of the detailed feedback. I've pushed a commit which addresses every comment except this one. How should the subpixel position be calculated? `x = (x * scale)` rounded to the nearest .5? Do we add or subtract .5 if starting from an integer? Thank you

src/draw.c
fwsmit: I don't think you've addressed this one

src/draw.c
fwsmit: Yeah that's a tough one. If you never want blurry lines, you would have to do some logic like that. Since we don't officially support fractional scaling, you might leave that as a TODO (make a comment in that case). With whole number scaling it should work like this.
If you want to support fractional scaling, I think the calculation should be done inside `draw_rounded_rect`. It needs the additional information of the line width. Then the calculation is as follows:
`x = round(x*scale) + half_frame_width`

src/settings.c
```diff
@@ -201,6 +201,12 @@ void check_and_correct_settings(struct settings *s) {
                 if (s->progress_bar_min_width > s->width.max) {
                         LOG_W("Progress bar min width is greater than the max width of the notification.");
                 }
+                int progress_bar_max_corner_radius = (s->progress_bar_height / 2);
+                if (s->progress_bar_corner_radius > progress_bar_max_corner_radius) {
+                        settings.progress_bar_corner_radius = progress_bar_max_corner_radius;
+                        LOG_W(  progress_bar_max_corner_radius,
```
fwsmit, line 207: I think these arguments have to be the other way around

vlfldr:
@fwsmit Sorry for the delay. Just want to clarify, when you say "With whole number scaling it should work like this" in a comment above do you mean it should work as-is with the parameters changed to floats, or it should work by rounding to the nearest .5?

I am going to hold off on passing the frame width to draw_rounded_rect to support fractional scaling for now as it would involved either overloading or touching more of the codebase than this PR intended to.

fwsmit:
> @fwsmit Sorry for the delay. Just want to clarify, when you say "With whole number scaling it should work like this" in a comment above do you mean it should work as-is with the parameters changed to floats, or it should work by rounding to the nearest .5?

It should work as  is. You can best check this by setting the progress bar width to 1 and seeing if the progress bar border is a nice 1-pixel line and not a blurry multi-pixel line.

I've tested it and it works great.

> I am going to hold off on passing the frame width to draw_rounded_rect to support fractional scaling for now as it would involved either overloading or touching more of the codebase than this PR intended to.

Okay that's fine

fwsmit:
After you address the open comment it's good

fwsmit:
Thanks! This is a great addition to dunst. I'll go ahead and merge it

PR #1124, opened 2022-11-04
notification-history.sh: Rewrite for speed

notevenaperson:
My main intention is minimize calls to jq since (on my thinkpad) each takes 50-100 ms, which can add up to a noticeable delay (~400ms for each notification) until the window first appears.

Seeing past notifications is an important part of how I use notifications, so I polished the script.

### DIFF
--- a/contrib/notification-history.sh
+++ b/contrib/notification-history.sh
@@ -1,47 +1,55 @@
-#!/bin/bash
+#!/bin/sh
+tests() {
+    #check if these get displayed right
+    notify-send "Dinner is ready"
+    notify-send "The\nTrick"
+    notify-send "The Trick
+#2"
+}
 
-history_json=$(dunstctl history)
+history_json="$(dunstctl history)"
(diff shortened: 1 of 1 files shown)


notevenaperson:
With 26 notifications in dunst history:
```sh
time
``` 
```
./(old)notification-history.sh  6.56s user 0.43s system 99% cpu 7.057 total
./(new)notification-history.sh  0.37s user 0.11s system 45% cpu 1.066 total
```

contrib/notification-history.sh
```diff
@@ -1,47 +1,55 @@
...
-history_length=$(echo $history_json|jq -r '.data[] | length')
-
-options=""
-
-for iter in $(seq $history_length); do
-    i=$((iter-1))
-    application_name=$(echo $history_json|jq -r .data[][$i].appname.data)
-    notification_summary=$(echo $history_json|jq -r .data[][$i].summary.data)
-    notification_timestamp=$(echo $history_json|jq -r .data[][$i].timestamp.data)
+#!/bin/sh
```
bebehei, line 1: KIM: You've changed shells here. This might not be the way things go. You might not have seen a problem, if on your system `/bin/sh` is linked to bash.

notevenaperson: No, I wrote it deliberately only using POSIX features. I had and have `/bin/sh` linked to `dash`.

fwsmit:
Good optimization. Looks good, so I'll merge it.

Issue #765, opened 2020-09-30
DBus PropertiesChanged signal not being emitted for `paused ` property


ammgws:
I want to monitor changes to `dunst`'s pause state in real time using D-Bus. According to the output of `busctl --user introspect org.freedesktop.Notifications /org/freedesktop/Notifications org.dunstproject.cmd0`, the `paused` property should emit a PropertiesChanged signal (assuming that's what `emits-change` in the below output means), however there appears to be no signal generated when monitoring the bus and running `dunstctl set-paused toggle`. Am I going mad or is there actually not meant to be a signal emitted when it's changed?

Introspection of the interface showing the `paused` property:
```
> busctl --user introspect org.freedesktop.Notifications /org/freedesktop/Notifications org.dunstproject.cmd0
NAME                   TYPE      SIGNATURE RESULT/VALUE FLAGS
.ContextMenuCall       method    -         -            -
.NotificationAction    method    i         -            -
.NotificationCloseAll  method    -         -            -
.NotificationCloseLast method    -         -            -
.NotificationShow      method    -         -            -
.Ping                  method    -         -            -
.paused                property  b         false        emits-change writable
```

Monitoring the bus while spamming `dunstctl set-paused toggle`:
```
>busctl --user monitor org.freedesktop.Notifications
Monitoring bus message stream.
‣ Type=method_call  Endian=l  Flags=0  Version=1 Cookie=2
  Sender=:1.361  Destination=org.freedesktop.Notifications  Path=/org/freedesktop/Notifications  Interface=org.freedesktop.DBus.Properties  Member=Get
  UniqueName=:1.361
  MESSAGE "ss" {
          STRING "org.dunstproject.cmd0";
          STRING "paused";
  };

‣ Type=method_return  Endian=l  Flags=1  Version=1 Cookie=46  ReplyCookie=2
  Sender=:1.303  Destination=:1.361
  UniqueName=:1.303
  MESSAGE "v" {
          VARIANT "b" {
                  BOOLEAN true;
          };
  };

‣ Type=method_call  Endian=l  Flags=0  Version=1 Cookie=2
  Sender=:1.362  Destination=org.freedesktop.Notifications  Path=/org/freedesktop/Notifications  Interface=org.freedesktop.DBus.Properties  Member=Set
  UniqueName=:1.362
  MESSAGE "ssv" {
          STRING "org.dunstproject.cmd0";
          STRING "paused";
          VARIANT "b" {
                  BOOLEAN false;
          };
[...]

PR #1127, opened 2022-11-16
DBus PropertiesChanged signal not being emitted for *Length PropertiesChanged

bim9262:
Similarly to what @ammgws noted in #765, for `paused`, the various Length PropertiesChanged were not emitting PropertiesChanged when notifications were opened/closed. This change makes sure that these values are always kept up to date.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -694,6 +694,73 @@ static struct notification *dbus_message_to_notification(const gchar *sender, GV
         return n;
 }
 
+void signal_length_propertieschanged()
+{
+
+        static unsigned int last_displayed = 0;
+        static unsigned int last_history = 0;
+        static unsigned int last_waiting = 0;
+
+        if (!dbus_conn)
+                return;
+
+        GVariantBuilder *builder = g_variant_builder_new(G_VARIANT_TYPE_VARDICT);
+        GVariantBuilder *invalidated_builder = g_variant_builder_new(G_VARIANT_TYPE_STRING_ARRAY);
+
+        unsigned int displayed = queues_length_displayed();
+        unsigned int history = queues_length_history();
+        unsigned int waiting = queues_length_waiting();
+        bool properties_changed = false;
+
+        if (last_displayed != displayed) {
+                g_variant_builder_add(builder,
+                                      "{sv}",
+                                      "displayedLength", g_variant_new_uint32(displayed));
+                last_displayed = displayed;
+                properties_changed = true;
+        }
+
+        if (last_history != history) {
+                g_variant_builder_add(builder,
+                                      "{sv}",
+                                      "historyLength", g_variant_new_uint32(history));
+                last_history = history;
+                properties_changed = true;
+        }
+        if (last_waiting != waiting) {
+                g_variant_builder_add(builder,
+                                      "{sv}",
+                                      "waitingLength", g_variant_new_uint32(waiting));
+                last_waiting = waiting;
+                properties_changed = true;
+        }
+
+        if (properties_changed) {
+                GVariant *body = g_variant_new("(sa{sv}as)",
+                                               DUNST_IFAC,
+                                               builder,
+                                               invalidated_builder);
+
+                GError *err = NULL;
+
+                g_dbus_connection_emit_signal(dbus_conn,
+                                              NULL,
+                                              FDN_PATH,
+                                              PROPERTIES_IFAC,
+                                              "PropertiesChanged",
+                                              body,
+                                              &err);
+                                
+                if (err) {
+                        LOG_W("Unable to emit signal: %s", err->message);
+                        g_error_free(err);
+                }
+        }
+
+        g_clear_pointer(&builder, g_variant_builder_unref);
+        g_clear_pointer(&invalidated_builder, g_variant_builder_unref);
+}
+
 static void dbus_cb_Notify(
                 GDBusConnection *connection,
                 const gchar *sender,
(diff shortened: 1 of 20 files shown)


fwsmit:
Thanks for submitting your PR. It seems there are errors with building this. Could you take a look at that?
To make sure this feature keeps working in the future, could you also add a test for it in `test/dbus.c`? There are already similar tests, so it shouldn't be too hard to do.

bim9262:
Sure, I think I got the error fixed, I'll look into adding a test now.

bim9262:
I've added in a test and everything now seems good on my end!

test/dbus.c
```diff
@@ -740,6 +815,45 @@ TEST test_signal_actioninvoked(void)
...
+        n_dbus->app_icon = "NONE2";
+        n_dbus->summary = "Headline for New";
+        n_dbus->body = "Text";
+        g_hash_table_insert(n_dbus->actions, g_strdup("actionkey"), g_strdup("Print this text"));
+
+        guint id;
+        ASSERT(dbus_notification_fire(n_dbus, &id));
+        ASSERT(id != 0);
+
+        signal_length_propertieschanged();
```
fwsmit, line 836: The point of this test is to check if the signal gets sent automatically, right? So you should use something like `queues_update` and check if it sends the signal.

src/dbus.c
```diff
@@ -656,6 +656,63 @@ static struct notification *dbus_message_to_notification(const gchar *sender, GV
...
+
+        GVariantBuilder *builder = g_variant_builder_new(G_VARIANT_TYPE_ARRAY);
+        GVariantBuilder *invalidated_builder = g_variant_builder_new(G_VARIANT_TYPE("as"));
+
+        unsigned int displayed = queues_length_displayed();
+        unsigned int history = queues_length_history();
+        unsigned int waiting = queues_length_waiting();
+        bool properties_changed = false;
+
+        if (last_displayed != displayed)
```
fwsmit, line 677: Do we want to signal only the variables that changed? It might make it harder for programs to use. It feels a bit weird to have a variable amount of changes in a single signal. Maybe signalling every variable separately would be better?

fwsmit:
There are still some memory leaks (you probably don't clean up your test or something). You can test for memory leaks locally with `make test-valgrind` (make sure to install valgrind).

test/dbus.c
fwsmit: Yes, I think `queues_update` is the best place to put it.

src/dbus.c
fwsmit: Allright, then it's fine. Could you maybe share the code you are using to test it?

fwsmit: Oh I see, you linked a PR.

src/dbus.c
```diff
@@ -856,13 +916,16 @@ gboolean dbus_cb_dunst_Properties_Set(GDBusConnection *connection,
...
+                                              PROPERTIES_IFAC,
                                               "PropertiesChanged",
                                               g_variant_new("(sa{sv}as)",
                                                             interface_name,
                                                             builder,
                                                             invalidated_builder),
                                               NULL);
+
+                g_clear_pointer(&builder, g_variant_builder_unref);
+                g_clear_pointer(&invalidated_builder, g_variant_builder_unref);
```
bim9262, line 928: I found a few memory leaks.

fwsmit: Thanks, that leak existed for more than 2 years! Although it only leaked memory every time you paused dunst.

fwsmit:
Maybe it'd be useful to add test coverage for `dbus_cb_dunst_Properties_Get` as well. It's not covered in master yet.

The test is currently failing, so could you take a look at that?

bim9262:
I am still having trouble with one test `test_notification_at`.

```
==99978== Invalid read of size 8                                                                                                                             
==99978==    at 0x4EAA576: g_dbus_connection_emit_signal (gdbusconnection.c:5698)
==99978==    by 0x12288D: signal_length_propertieschanged (dbus.c:749)        
==99978==    by 0x12CA8F: test_notification_at (input.c:102)                                                                                                 
==99978==    by 0x12CA8F: suite_input (input.c:154)                           
==99978==    by 0x13EA0B: greatest_run_suite (test.c:33)                                                                                                     
==99978==    by 0x13EA0B: greatest_run_suite (test.c:33)              
==99978==    by 0x120BAA: main (test.c:71)     
```

I spent a few hours trying to figure out what's causing the error. There are lots of other tests that call `queues_update`  which then calls `signal_length_propertieschanged`. I noticed that this is the only test that calls `queues_init`, but not `queues_teardown`.

However adding in the matching `queues_teardown` results in the following error:

```
test: test/../src/notification.c:271: notification_unref: Assertion `n->priv->refcount > 0' failed.
bash: line 2: 101920 Aborted                 (core dumped) ./test/test -v
     101921 Done                    | ./test/greenest.awk
make: *** [Makefile:76: test] Error 134
```

fwsmit:
I've investigated the issue and it's caused by improper cleanup of the dbus tests. The connection is not set to NULL after running the dbus tests. I also made sure to not double free the notification, but not sure if this is neccesary. This diff should solve the issue:

```
diff --git a/src/dbus.c b/src/dbus.c
index fcd7ffd..13b8121 100644
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -25,7 +25,7 @@
 
 #define PROPERTIES_IFAC "org.freedesktop.DBus.Properties"
 
-GDBusConnection *dbus_conn;
+GDBusConnection *dbus_conn = NULL;
 
 static GDBusNodeInfo *introspection_data = NULL;
 
@@ -1187,6 +1187,7 @@ void dbus_teardown(int owner_id)
         g_clear_pointer(&introspection_data, g_dbus_node_info_unref);
 
         g_bus_unown_name(owner_id);
+        dbus_conn = NULL;
 }
 
 /* vim: set ft=c tabstop=8 shiftwidth=8 expandtab textwidth=0: */
diff --git a/test/input.c b/test/input.c
index ae8ec14..21deae9 100644
--- a/test/input.c
+++ b/test/input.c
@@ -145,7 +145,8 @@ TEST test_notification_at(void)
         ASSERT(result != NULL);
         ASSERT(result == bottom_notification);
 
-        g_slist_free_full(notifications, free_dummy_notification);
+        queues_teardown();
+        g_slist_free(notifications);
         PASS();
 }
```

bim9262:
Thank you so much @fwsmit, that patch seems to have sorted out the issues I was seeing in valgrind.

src/dbus.c
```diff
@@ -694,6 +694,76 @@ static struct notification *dbus_message_to_notification(const gchar *sender, GV
...
+        if (last_displayed != displayed)
+        {
+                g_variant_builder_add(builder,
+                                      "{sv}",
+                                      "displayedLength", g_variant_new_uint32(displayed));
+                last_displayed = displayed;
+                properties_changed = true;
+        }
+
+        if (last_history != history)
```
bebehei, line 724: Please use consistent style for this. IIRC it should be this:

```suggestion
        if (last_history != history){
```

bim9262: It seems like there are a handful of inconsistent formatting (extra line break or no space between the `)` and `{`).
I'll fix them in a separate commit.

bim9262:
@fwsmit, is this ready to merge in or is there something else I need to add or change?

fwsmit:
I wanted to look at the test coverage before merging. Did you add a test for `dbus_cb_dunst_Properties_Get`?

test/dbus.c
```diff
@@ -296,6 +371,56 @@ TEST test_invalid_notification(void)
         PASS();
 }
 
+TEST test_dbus_cb_dunst_Properties_Set(void)
```
bim9262, line 374: I didn't add a test for `dbus_cb_dunst_Properties_Get`, just `dbus_cb_dunst_Properties_Set`.

bim9262:
I just added a small test for `dbus_cb_dunst_Properties_Get`. How exhaustive would you like to make the test?

fwsmit:
> I just added a small test for `dbus_cb_dunst_Properties_Get`. How exhaustive would you like to make the test?

Just one test is already better than none. Thanks for your contribution! I'll merge it

Issue #1122, opened 2022-11-02
libXScrnSaver and libnotify disabled at build time


apprehensions:
IMO, libXScrnSaver and libnotify should be disabled at build time via `IDLE=0` and `DUNSTIFY=0`, for dunstify it can be added to Makefile rule dunstify to check for libnotify. but for libxscrnsaver i haven't got very far.

wael444:
@fwsmit why is libXScrnSaver needed? i was hoping for an answer.

bebehei:
AFAIK is libXScrnSaver used for X11 when to detect if we're idling.

wael444:
yeah that must be a optional feature at build time... but couldnt figure it out due to a segmentation fault.

bebehei:
If I remember this correctly - can be changed by @fwsmit in the meantime - the X11-Codeparts are not yet optional, so neither libXScrnSaver. But take this with a grain of salt, it might be FUD.

fwsmit:
The X11 code part is not yet optional. You'll have to #ifdef it out, just like I've done with the wayland code. I would go straight for all the X11 code, not just libXscreensaver

wael444:
i'd still want libXScrnSaver disabled at build time, since it's practically a useless dependency when the feature its used for isn't used.

fwsmit:
So I understand correctly you are using X11? 
You can also just ifdef out the X11 Screensaver part of you want. But I don't see many people wanting to only get rid of the Screensaver part, so I don't think that should be merged upstream

wael444:
> You can also just ifdef out the X11 Screensaver part of you want.

technically thats the same as removing it, which for me ended up in a segmentation fault.

fwsmit:
Yeah, you'll also have to make sure it's not used anywhere

bynect:
I think that this is no longer relevant thanks to #1290.

apprehensions:
This issue was made at a different stage when I was growing up, when I cared about minimalism and dependencies in my system. At that time I hadn't known X was designed for a different time and had protocols/extensions ontop of Xorg, which is why we have XScrnSaver.

Personally, I understand the need for idle detection and I can definitely find use to it, but not everyone will find these features useful - especially those who care about dependencies. I asked for a re-opening so I can close the issue properly as I am the OP.

bynect:
[...]

PR #1129, opened 2022-11-21
make libnotify & dunstify optional

wael444:
solves #1122 's libnotify, but not the libXScrnSaver problem

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ DEPS := ${SRC:.c=.d} ${TEST_SRC:.c=.d}
 
 
 .PHONY: all debug
-all: doc dunst dunstify service
+all: doc dunst service
 
 debug: CFLAGS   += ${CPPFLAGS_DEBUG} ${CFLAGS_DEBUG}
 debug: LDFLAGS  += ${LDFLAGS_DEBUG}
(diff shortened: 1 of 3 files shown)


bebehei:
Could you rename this variable flag to `DUNSTIFY`? For me this makes much more sense, since we're enabling/disabling dunstify with that.

Makefile
```diff
@@ -206,8 +209,11 @@ install-service-systemd: service-systemd
 	install -Dm644 dunst.systemd.service ${DESTDIR}${SERVICEDIR_SYSTEMD}/dunst.service
 endif
 
+ifeq (1,${LIBNOTIFY})
```
fwsmit, line 212: I think this should be
```suggestion
ifneq (0,${LIBNOTIFY})
```
Otherwise this results in false when no variable is defined. It should default to true, as stated in the README.

fwsmit:
It doesn't seem to compile

Makefile
wael444: I was on a system without libnotify, so this wasn't tested. thank you!

config.mk
```diff
@@ -30,6 +30,12 @@ VALGRIND ?= valgrind
 # You can also use "make WAYLAND=0" to build without wayland
 # WAYLAND ?= 0
 
+# Disable dependency on libnotify. this will remove the ability
+# to easily send notifications using dunstify or notify-send.
+# Do this if you have your own utility to send notifications.
+# Other applications are fine.
+# LIBNOTIFY ?=0
```
fwsmit, line 37: Forgot to change the naem here

wael444: oih n yg ydogaywe7guwryg6ygbjkmsdhagby r7eyhgubdfakahsbyfdbh jkdfgygbyhweraugjk6fdyh jzxcvlg6yrweigjhketubujfdzkl;g6xbycvoikbuyysgbivjshyikxvcyhjm kweriu6ygbklzxc6byyseruigoy6b cvjmk.kgb6yvhrsljkb6 vyhgkl6yberfhn jgbrweujmgfb
cb

fwsmit:
Thanks, looks good!

PR #1131, opened 2022-11-23
Add history-clear command

fapfaff:
Added command to remove all notifications from history.

### DIFF
--- a/src/queues.c
+++ b/src/queues.c
@@ -347,7 +347,10 @@ void queues_history_clear(void)
         if (g_queue_is_empty(history))
                 return;
 
-        g_queue_clear_full(history, (GDestroyNotify)notification_unref);
+        while (!g_queue_is_empty(history)) {
+                struct notification *n = g_queue_pop_head(history);
+                notification_unref(n);
+        }
 }
 
 /* see queues.h */
--- a/src/queues.c
+++ b/src/queues.c
@@ -344,13 +344,8 @@ void queues_notification_close(struct notification *n, enum reason reason)
 /* see queues.h */
 void queues_history_clear(void)
 {
-        if (g_queue_is_empty(history))
-                return;
-
-        while (!g_queue_is_empty(history)) {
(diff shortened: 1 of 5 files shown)


fwsmit:
Just curious, what do you use this for?

fapfaff:
For an eww widget displaying my notification history

fapfaff:
Do you have an idea why it doesn't build?
Shouldn't g_queue_clear_full() be included by glib.h?

Sorry for asking you my noob questions, but it's working locally without any problems and I couldn't find anything that helps me.

fwsmit:
I'm guessing the version of glib for these distro's is too old. You need at least version 2.6 to get this function. I've ran the archlinux build again and it works. This seems to confirm my suspicion. vTake a look at https://docs.gtk.org/glib/struct.Queue.html for the version requirements of different functions.

You can work around this by using other queue functions in glib.

fapfaff:
Thanks, I have replaced it. Should work now.

src/queues.c
```diff
@@ -347,7 +347,10 @@ void queues_history_clear(void)
         if (g_queue_is_empty(history))
```
fwsmit, line 347: This is unnecesary

src/queues.c
```diff
@@ -347,7 +347,10 @@ void queues_history_clear(void)
...
                 return;
 
-        g_queue_clear_full(history, (GDestroyNotify)notification_unref);
+        while (!g_queue_is_empty(history)) {
+                struct notification *n = g_queue_pop_head(history);
+                notification_unref(n);
+        }
```
fwsmit, line 353: You can actually use the source of glib to copy the function here. See https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gqueue.c#L166

```
g_queue_foreach (queue, (GFunc) free_func, NULL);
g_queue_clear (queue);
```

fwsmit (changes requested):

fwsmit:
LGTM, thanks!

PR #1139, opened 2023-02-06
Add option to set icon image corner radius

Back777space:
I added an option to set the icon image corner radius.

The variable name is `icon_corner_radius` and it would just have to be added to the `dunstrc` file.
It looks like this when used and it seems to work fine for me: 
image

Example:
image


fwsmit:
Wanted to comment on the dunstrc, but I see you've already fixed it :p
Thanks for your contribution!

Back777space:
> Wanted to comment on the dunstrc, but I see you've already fixed it :p
> Thanks for your contribution! 

Yeah it was a quick fix, happy to help :)

Issue #1102, opened 2022-08-20
High CPU usage for updating notification timestamps


mustaqimM:
### Issue description
With `idle_threshold` set, old notifications (ones that say "X min old") increase the CPU usage:
 
![High cpu usage on old notifications](https://user-images.githubusercontent.com/24436622/185745676-1be61b72-7152-4cbc-a12c-3bb5221ba36d.png)

I believe it's related to https://github.com/dunst-project/dunst/issues/169

### Installation info

- Version: `1.9.0 (2022-06-27)`
- Install type: `system package`
- Window manager / Desktop environment: `swaywm`

<details>
<summary>Minimal dunstrc</summary>

```ini
# -*- eval: (rainbow-mode); -*-
# See dunst(5) for all configuration options

[global]
    ### Display ###

    # Which monitor should the notifications be displayed on.
    monitor = 0

    # Display notification on focused monitor.  Possible modes are:
    #   mouse: follow mouse pointer
    #   keyboard: follow window with keyboard focus
    #   none: don't follow anything
    #
    # "keyboard" needs a window manager that exports the
    # _NET_ACTIVE_WINDOW property.
    # This should be the case for almost all modern window managers.
    #
    # If this option is set to mouse or keyboard, the monitor option
    # will be ignored.
    follow = mouse

    ### Geometry ###

    # dynamic width from 0 to 300
    # width = (0, 300)
    # constant width of 300
    width = 250

    # The maximum height of a single notification, excluding the frame.
    height = 300

    # Position the notification in the top right corner
    origin = top-right

    # Offset from the origin
    offset = 10x40

    # Scale factor. It is auto-detected if value is 0.
    scale = 0

    # Maximum number of notification (0 means no limit)
    notification_limit = 0

    ### Progress bar ###

    # Turn on the progess bar. It appears when a progress hint is passed with
    # for example dunstify -h int:value:12
    progress_bar = true

    # Set the progress bar height. This includes the frame, so make sure
    # it's at least twice as big as the frame width.
    progress_bar_height = 10

    # Set the frame width of the progress bar
    progress_bar_frame_width = 1

    # Set the minimum width for the progress bar
    progress_bar_min_width = 150

    # Set the maximum width for the progress bar
    progress_bar_max_width = 300


    # Show how many messages are currently hidden (because of
    # notification_limit).
    indicate_hidden = yes

    # The transparency of the window.  Range: [0; 100].
    # This option will only work if a compositing window manager is
[...]

Issue #1078, opened 2022-05-21
Dunst sometimes sleeps too short


fwsmit:
To reproduce, create a few notifications. Let them time out and then retrieve them from history. Dunst will then sleep sleep like the notification will time out, even when you've set the history to not time out.

No notifications are closed too early or too late.

Not reproducible with 40ec2d50268f49a5ecbf98059079b8252ae9b407
Reproducible with 3bba0b01f924a4af9c2a0b13125446153f3d4125

fwsmit:
This is most likely caused by #1073

bynect:
was this solved in the end?

fwsmit:
Not sure. Maybe @tobast knows that?

tobast:
I'll have to check, I believe I fixed it but I am not entirely sure anymore.

PR #1140, opened 2023-02-09
Floor timestamp to the second when displaying age

tobast:
This fixes the issue discussed in #1102; however, profiling shows that this is not the only performance problem when many notifications are active at once — which happens if eg. the computer is left unattended for a long time with `idle_threshold > 0`.


fwsmit:
It seems like this is more a workaround for the bug. Dunst should not redraw the notification when the age is a millisecond different. Is it possible to fix it at the point where age value is checked?

tobast:
> Thanks for taking a look at the issue! It seems like this is more a workaround for the bug. Dunst should not redraw the notification when the age is a millisecond different. Is it possible to fix it at the point where age value is checked?

I have not gone into the details of what happens during a refresh, but the main problem here seems that if there are eg. 200 notifications, Dunst will be awoken in average once every 5ms. Whether the refresh stack is fast or not, 5ms is a very short timespan; if it takes eg. 2ms of boilerplate and pretty much nothing of actual refresh, this still makes for 40% CPU usage (2ms work every 5ms). Thus, it seems to me that it is crucial not to wake Dunst this often.

> It would also be nice to have a test to prevent this or similar errors in the future.

Agreed. I don't see easily how that would be a unit test, though.

> > profiling shows that this is not the only performance problem when many notifications are active at once
> 
> Where are some other areas for improvement?

When profiling (with this patch applied) using

```bash
$ perf record -g ./dunst
```

and stress-testing with

```
$  for i in $(seq 1 200); do notify-send "Test $i" "Test"; done
```

while `notification_limit` is not set in the configuration, and `show_age_threshold` is set to a low value (eg. `5`), `perf report` shows that

```
  Children      Self  Command         Shared Object                   Symbol                                                                                                                                                                                                     
-   69.11%     0.05%  dunst           dunst                           [.] draw                                                                                                                                                                                                  ◆
   - 69.06% draw                                                                                                                                                                                                                                                                ▒
      - 39.08% layout_render                                                                                                                                                                                                                                                    ▒
         - 31.90% render_content                                                                                                                                                                                                                                                ▒
            - 12.68% layout_get_height                                                                                                                                                                                                                                          ▒
               + 12.60% get_text_size                                                                                                                                                                                                                                           ▒
            + 11.48% pango_cairo_show_layout                                                                                                                                                                                                                                    ▒
            + 4.99% layout_setup                                                                                                                                                                                                                                                ▒
            + 0.91% g_log                                                                                                                                                                                                                                                       ▒
            + 0.58% cairo_fill                                                                                                                                                                                                                                                  ▒
         + 4.51% x_get_scale                                                                                                                                                                                                                                                    ▒
         + 1.98% render_background                                                                                                                                                                                                                                              ▒
      - 21.39% calculate_dimensions                                                                                                                                                                                                                                             ▒
         - 20.10% calculate_notification_dimensions                                                                                                                                                                                                                             ▒
            + 15.41% get_text_size                                                                                                                                                                                                                                              ▒
            + 4.57% layout_setup                                                                                                                                                                                                                                                ▒
         + 1.04% g_log                                                                                                                                                                                                                                                          ▒
      + 5.41% create_layouts                                                                                                                                                                                                                                                    ▒
      + 1.69% g_slist_free_full                                                                                                                                                                                                                                                 ▒
      + 0.91% cairo_image_surface_create
```

I have not looked into the details, but from a quick glance, it seems that a lot of time is spent figuring the text size --- here, there are merely 200 notifications. For 200 notifications, it still takes about 150ms to redraw them each second. I'm not sure where the room for improvement mostly is, but I expect that all of this is redrawn only when the text changes, that is, every second.

Either part of these computations can be cached from one redraw to the next (but the age part makes it hard, I guess), or maybe it would be a good idea to simply stop drawing when the size of the notification stack exceeds the size of the screen? (I expect this is what the `notification_limit` setting is about, though).

fwsmit:
What you're saying makes sense. Dunst should only wake up once a second to update the timestamps. I would suggest every time the real time gets to the next second. This can be fixed in queues_get_next_datachange in queues.c. This is also the function for which you can write a test.

The draw code could be optimized, but I don't think it needs to, since we (should) only draw about once a second at most.

fwsmit:
I would also say that the notification limit should help against it. Setting it to some reasonable value by default should also help

tobast:
> What you're saying makes sense. Dunst should only wake up once a second to update the timestamps. I would suggest every time the real time gets to the next second. This can be fixed in queues_get_next_datachange in queues.c. This is also the function for which you can write a test.

In practice, this is what this PR's code does: by flooring the notifications' timestamps (for display purposes), this ensures that the next wake-up for this purpose is when the real time is at an exact second. It also does not do so when there is no notification age to update. If you prefer, I could rewrite this to do explicitly what you suggest (check that there is one notification needing update, and if so, wait until the next exact second), but I find the code quite elegant this way.

I could indeed write a test in any case, though.

> The draw code could be optimized, but I don't think it needs to, since we (should) only draw about once a second at most.

This problem is independent, but this is not exactly true: it does trigger only once a second, but with ~200 notifications, the (single) update takes about 150ms, which is already a 15% CPU usage for something that should not, I believe, be compute-intensive.

Setting a reasonable `notification_limit` value does help tremendously, though.

src/notification.c
```diff
@@ -671,8 +671,11 @@ void notification_update_text_to_render(struct notification *n)
         }
 
         /* print age */
+        /* Note that this floors the notification's timestamp to the second.
+         * This ensures that all notifications ages will change at once, and
+         * that only one update will occur each second for this purpose. */
         gint64 hours, minutes, seconds;
-        gint64 t_delta = time_monotonic_now() - n->timestamp;
+        gint64 t_delta = time_monotonic_now() - (n->timestamp - n->timestamp % S2US(1));
```
fwsmit, line 678: I don't think this one is neccesary. Please move the comment to queues_get_next_datachange, since that's where the magic happens.

src/queues.c
```diff
@@ -569,7 +569,9 @@ gint64 queues_get_next_datachange(gint64 time)
                 }
 
                 if (settings.show_age_threshold >= 0) {
-                        gint64 age = time - n->timestamp;
+                        /* When displaying age, floor the timestamp to the
+                         * second. See notification.c for details.*/
+                        gint64 age = time - (n->timestamp - (n->timestamp % S2US(1)));
```
fwsmit, line 574: I'd suggest to move the flooring to inside the if statement. That's a little more explicit and makes sure you aren't too early.
```c
gint64 next_second = time - time % S2US(1);
sleep = MIN(sleep, (S2US(1) - (next_second % S2US(1))));
```

fwsmit (changes requested):

fwsmit:
> In practice, this is what this PR's code does: by flooring the notifications' timestamps (for display purposes), this ensures that the next wake-up for this purpose is when the real time is at an exact second. It also does not do so when there is no notification age to update. If you prefer, I could rewrite this to do explicitly what you suggest (check that there is one notification needing update, and if so, wait until the next exact second), but I find the code quite elegant this way.

I have suggested a way that's a little more explicit and hopefully easier to read. 

> I could indeed write a test in any case, though.

:+1:

> This problem is independent, but this is not exactly true: it does trigger only once a second, but with ~200 notifications, the (single) update takes about 150ms, which is already a 15% CPU usage for something that should not, I believe, be compute-intensive.
> 
> Setting a reasonable `notification_limit` value does help tremendously, though.

Yes, could you also do that?

src/queues.c
```diff
@@ -569,7 +569,9 @@ gint64 queues_get_next_datachange(gint64 time)
...
 
                         // sleep exactly until the next shift of the second happens
```
fwsmit, line 576: This comment is not really accurate anymore

src/notification.c
tobast: I actually believe that it is still necessary -- although the comment should be moved. If not, if a notification was issued at timestamp 0.5 and `show_age_threshold` is 0, and some other notification requires update (eg. expires) at timestamp 2.6, this will result in a strange outcome for the user:

* at t=0.5, the notification is shown with age=0s
* at t=1 (age update), the age is still shown as 0s
* at t=2 (age update), the age shows 1s — 0s was displayed for 1.5s, but this is not a huge problem as this must be bootstraped anyway
* at t=2.6, the age shows 2s — 1s was displayed for 0.6s, which will look weird to the user
* at t=3s, still 2s
* at t=4s, 3s — 2s displayed for 1.4s

This can be repeated with any number of unrelated updates, and can result in a weird display. I believe the best way to avoid this is to keep this patch; this way, the first second may be stretched longer than 1s, but all subsequent updates will be regular.

src/queues.c
tobast: Done in the current revision.

tobast:
Whoops, the current revision doesn't pass one of the tests. I'm checking it.

tobast:
> > I could indeed write a test in any case, though.
> 
> +1

Done

> > Setting a reasonable `notification_limit` value does help tremendously, though.
> 
> Yes, could you also do that?

Done

It also seems to me that [dunst.c:95](https://github.com/dunst-project/dunst/blob/master/src/dunst.c#L95) 
```c
gint64 sleep = queues_get_next_datachange(now);
```
is actually wrong: computing the sleep delay from `now`, which is obtained a bit higher with
```c
gint64 now = time_monotonic_now();
```

is problematic, because between those two lines, all the loop logic happens, some of which might be taking time. It seems reasonable to use the same value for `now` throughout the computations (at least to have consistent behaviour), but I believe it would be more reasonable to query `time_monotonic_now()` a second time for computing the sleep delay? If you agree, I can also fix.

tobast:
[bc257cf](https://github.com/dunst-project/dunst/pull/1140/commits/bc257cf6ef4e7708ff85d86fe3ed9ddd1ce3505d) also ensures that `run()` only has a single active GLib timeout. In multiple cases, while stress-testing, I noticed that multiple timers were active. This can happen if a timeout is set, `run` is woken up by an incoming notification, and the new computed timeout is before the old one, but it seems not to be the only case — though I don't understand how. In this case, a timeout is added, but the old remains active, and when it ticks, it adds again its own timeout (`reason == 0`). This goes on forever, and causes `dunst` to wake up multiple times when a single time is needed.

fwsmit:
> 



> It also seems to me that [dunst.c:95](https://github.com/dunst-project/dunst/blob/master/src/dunst.c#L95)
> 
> ```c
> gint64 sleep = queues_get_next_datachange(now);
> ```
> 
> is actually wrong: computing the sleep delay from `now`, which is obtained a bit higher with
> 
> ```c
> gint64 now = time_monotonic_now();
> ```
> 
> is problematic, because between those two lines, all the loop logic happens, some of which might be taking time. It seems reasonable to use the same value for `now` throughout the computations (at least to have consistent behaviour), but I believe it would be more reasonable to query `time_monotonic_now()` a second time for computing the sleep delay? If this seems correct to you, I can also fix this.

I see I've changed that in https://github.com/dunst-project/dunst/pull/1073, more specifically in 1fc9cb678bfa44142454cdf38ad209a79fa64a94. I'll have to think a bit more about why I did that. It comes done to the question if queues_update and queues_get_next_datachange need the same time. 

> [bc257cf](https://github.com/dunst-project/dunst/pull/1140/commits/bc257cf6ef4e7708ff85d86fe3ed9ddd1ce3505d) also ensures that `run()` only has a single active GLib timeout. In multiple cases, while stress-testing, I noticed that multiple timers were active. This can happen if a timeout is set, `run` is woken up by an incoming notification, and the new computed timeout is before the old one, but it seems not to be the only case — though I don't understand how. In this case, a timeout is added, but the old remains active, and when it ticks, it adds again its own timeout (`reason == 0`). This goes on forever, and causes `dunst` to wake up multiple times when a single time is needed.

Perhaps this is related: #1078

tobast:
> I think it might be more sane to calculate the time at which the sleep should end, instead of the duration of the sleep. This way slow code doesn't impact the sleep duration (keep in mind the timer is still not guaranteed to be accurate).

This is implemented, with tests updated, in the latest revision.

src/notification.c
fwsmit: Okay, that makes sense. It sounds like reasonable behavior.

fwsmit:
Thanks for taking the time to debug this and find a good solution. I'll merge this now

biopsin:
This PR broke disabling idle_threshold when setting = -1 as spesified in dunstrc

tobast:
I'll have a look into this as soon as I get time.

PR #1141, opened 2023-02-14
rules: Add conditional branch statement

likunyur:
Judge err_buf pointer is not empty before executing the content, which increases the robustness of the program.

### DIFF
--- a/src/rules.c
+++ b/src/rules.c
@@ -169,9 +169,11 @@ static inline bool rule_field_matches_string(const char *value, const char *patt
                 if (err) {
                         size_t err_size = regerror(err, &regex, NULL, 0);
                         char *err_buf = malloc(err_size);
-                        regerror(err, &regex, err_buf, err_size);
-                        LOG_W("%s: \"%s\"", err_buf, pattern);
-                        free(err_buf);
+                        if (err_buf != NULL) {
+                                regerror(err, &regex, err_buf, err_size);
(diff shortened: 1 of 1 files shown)


fwsmit:
I think there are many more cases of not checking the return value of malloc. So it is likely the program will crash soon after in a low memory situation.

likunyur:
If these need to be checked, I could take time to analyze the code and do some optimization work.

bebehei:
The whole code has the assumption, that `g_malloc` will always return non-NULL values.

Those occurences of `malloc` should actually be changed to `g_malloc`.

I find these two occurences, which have to get fixed:

```text
[bebe:~/code/dunst] master+* ± git grep -P '(?<!g_)malloc'
.valgrind.suppressions:   fun:malloc
.valgrind.suppressions:   fun:malloc
.valgrind.suppressions:   fun:malloc
.valgrind.suppressions:   fun:malloc
.valgrind.suppressions:   fun:malloc
src/icon-lookup.c:#include <malloc.h>
src/rules.c:                        char *err_buf = malloc(err_size);
test/icon-lookup.c:        struct rule *rule = malloc(sizeof(struct rule));
[bebe:~/code/dunst] master+* ± 
```

bebehei:
I've opened PR #1144 for this and also a regression test, that this won't happen again in future.

Sorry for not merging your PR, but thank you for pointing out the problem!

PR #1145, opened 2023-02-16
Make use of USEC conversion standardized

bebehei:


fwsmit:
Looks good. Thank you, bebehei

PR #1148, opened 2023-02-25
Fix #1147 by checking NULL pointers

### LINKED ISSUE #1147: Segmentation fault on Wayland (`ctx.current_buffer` is NULL)
Hey, I'm unsure of what triggers the bug exactly, but it seems that the dunst daemon crashes randomly on SwayWM. All notifications are lost.

I may be wrong, but I think it happens when swaylock locks the session.

I tried to reproduce using the following one liner, but it didn't trigger the bug:

```
swayidle -w timeout 5 swaylock timeout 10 swaymsg "output * power off" resume swaymsg "output * power on" before-sleep swaylock
```

Here's the backtrace from the coredump:

image

So apparently this is due to the following function:

https://github.com/dunst-project/dunst/blob/7bd8e6b87f06702d8106dc46fb4ed5ffca588bf3/src/wayland/wl.c#L817-L823

It seems to confirm this is somehow related to swaylock, or the outputs being turned off by swaymsg. I'm not familiar with Wayland, but maybe when the outputs are off there's no "surface to draw on" so no `current_buffer` ? I don't know just a guess.

The segfault is caused by `ctx.current_buffer` being null:

image

I'm running Dunst 1.9.0 from the ArchLinux repositories.

ShellCode33:
Please review carefully, I'm not familiar at all with dunst nor Wayland. This might not be the right way to fix #1147.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -870,7 +870,13 @@ void draw(void)
 {
         assert(queues_length_displayed() > 0);
 
-        GSList *layouts = create_layouts(output->win_get_context(win));
+        cairo_t *c = output->win_get_context(win);
+
+        if(c == NULL) {
+                return;
+        }
+
+        GSList *layouts = create_layouts(c);
 
         struct dimensions dim = calculate_dimensions(layouts);
         LOG_D("Window dimensions %ix%i", dim.w, dim.h);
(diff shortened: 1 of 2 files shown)


fwsmit:
The change seems sensible. Have you tested if this fixes the bug?

ShellCode33:
The bug doesn't seem to occur anymore

fwsmit:
Thanks! The patch looks good, so I'll merge it

Issue #509, opened 2018-04-02
Custom indicators for URL and Actions


Sweenu:
I'd love to have the opportunity to change my notifications' colors or frame if they have actions or URLs.
Or at least, be able to change `(A)` and `(U)` to something else.

ismig:
I find the (AU) fairly distracting. It's so large it dwarfs the appname of my notifications, which is deliberately small and grey.
Small icons would be less distracting and easier to interpret -- maybe the link icon and something else for actions (a hammer? lol). Or just a custom format string.

Kryszak:
Simply being able to set indicators to some other string stylable with HTML would be great!

tsipinakis:
I have a faint memory of discussing this with @bebehei but I can't seem to find any logs of it.

How about adding conditionals to the `format` setting for this?
For example the default format becomes something like `%?action?(A)? %?url?(U)? %?both?(AU)?`

Though I don't like that we have to use `both` as well to keep the behaviour consistent. This also raises the question of how to add this without breaking indicators for all existing configs.

bebehei:
@tsipinakis Yeah, we had been discussing previously about this. But I'm not sure, where. I know I had been starting to implement similar stuff with a date field (IIRC I closed this PR because I didn't find an acceptable solution).

monkz:
For my part, i would like to customize them, to add the count of actions that is bound to a notification - Like A3. And for U replace it with a emoji.

elamperti:
Hi everyone, I've drafted [a PR](https://github.com/dunst-project/dunst/pull/1150) with changes to have custom indicators. I tried it with emoji and it works (and looks quite nice!). The indicators can be styled in the format string using Pango markup.

I'd love to hear your feedback about the implementation.

---

@tspinakis from what I saw in the code, having conditionals in the format string would be probably difficult to implement (and/or would end up requiring an ugly syntax), so I'd discard that for now.

@monkz this implementation doesn't show the number of actions or URLs, it could be added but I don't know if it would be generally useful. What would be a use case that favors knowing the number of actions at first glance?

monkz:
@elamperti This allows me to determine if I need the context menu or just a click on the notification.

PR #1150, opened 2023-03-10
Allow using custom indicators for actions and URLs

elamperti:
This PR introduces two new settings (`action_indicator` and `url_indicator`) which could be set to any string, and then be placed wherever the user finds it useful in the `format` string (using `%A` or `%U`).

There's also the possibility to use `%D` in the format to show how many duplicates a given notification has.

I've tested this locally using emoji instead of the classic "(A)" and "(U)". It works, so it should be usable with nerd fonts and similar icon fonts too.

ℹ️  As it is now, this PR is a draft to gather feedback. Things that are still to be done or defined:

- [ ] Tests
- [ ] Should we use a single format token (e.g. `%X`) to show all possible indicators in it? (actions, URLs, duplicates)
- [ ] How could this deprecate the old implementation? (or how should they live together to avoid duplicate indicators?)
- [ ] Get feedback from other users (specially those who participated in #509)


fwsmit:
Thanks for submitting this PR

> Should we use a single format token (e.g. %X) to show all possible indicators in it? (actions, URLs, duplicates)

I don't think this is necessary. If you have a use case, then I wouldn't be against it.

> How could this deprecate the old implementation? (or how should they live together to avoid duplicate indicators?)

`show_indicators` is not really useful anymore with this implementation, so it should be marked deprecated. Then we could remove it in the next major version (which will probably not come any time soon). In the mean time, `show_indicators` should just prepend `(%A%U)` or something similar to the notification text, before it's being processed by your new code. If you do that right, nothing should change for existing users, but they are able to change the indicator after that.

`hide_duplicate_count=false` should also just append `(%D)` or similar to the notification text.

fwsmit:
Hi, have you looked at my above comment? Let's figure out together what to do to finish this PR

elamperti:
Hi, sorry for leaving this PR hanging. I'll get back to it this week. Thank you!

fwsmit:
Hey, thanks for your update. Let me know when you get to it!

bynect:
It would be nice to add a format parameter for the number of actions

fwsmit:
Yeah, that should definitely be possible

fwsmit:

elamperti:
Sorry for dropping the ball. I've just opened a new PR with the missing changes + added the option to show action count using `%C`.

PR #1155, opened 2023-03-23
Sort types and ascending and descending strategy

### LINKED ISSUE #206: Do not move notifications around while I'm reading them
dunst pushes notifications around when either new ones appear or old ones disappear – depending on whether `y` in `geometry` is a positive or negative value.

As soon as a notification is being shown and I'm potentially reading it, it should stay at its position until it disappears. At least optionally.

This would be a very nice addition for people who are as bad as I at reading notifications that suddenly move once or twice across my screen just to disappear before I've read to the end.

### LINKED ISSUE #845: Feature Request: Configurable Notification Sorting Order (Ascending / Descending / Random?)
This request is related to the following bug report:
#838 Disabled Sorting Order Is Not Honored While Dunst Is Paused

Lets call the new global option being made available to control this behavior: "sort_order" (_or go with whatever you prefer_).  Once implemented, generating notifications as follows:
`notify-send 1`
`notify-send 2`
`notify-send 3`
`notify-send 4`
`notify-send -u low 1`
`notify-send -u low 2`
`notify-send -u low 3`
`notify-send -u low 4`
`notify-send -u critical 1`
`notify-send -u critical 2`
`notify-send -u critical 3`
`notify-send -u critical 4`

Would produce these notification results (_when using the default "true" value for the "sort", or more aptly named: "sort_urgency", option; a "false" value would simply make Dunst not group the notifications by urgency at all_):

* when **sort_order** in the dunstrc configuration is set to **ascending**:
1 (_critical_)
2 (_critical_)
3 (_critical_)
4 (_critical_)
1
2
3
4
1 (_low_)
2 (_low_)
3 (_low_)
4 (_low_)

* when **sort_order** in the dunstrc configuration is set to **descending**:
4 (_low_)
3 (_low_)
2 (_low_)
1 (_low_)
4
3
2
1
4 (_critical_)
3 (_critical_)
2 (_critical_)
1 (_critical_)

* when **sort_order** in the dunstrc configuration is set to **random** (_results may actually vary; I'm also not strongly attached to getting this particular setting choice, so I'd be fine with you deciding that it's too much of a gimmick and not implementing it_):
1 (_low_)
4 (_low_)
3 (_low_)
2 (_low_)
1
4
2
3
3 (_critical_)
1 (_critical_)
2 (_critical_)
4 (_critical_)

duarm:
A simple implementation of the idea discussed on #845, fixes #845. The new descending/ascending types also fixes #206 if you don't hit the notification_limit and put anything on the waiting queue.

~~The idea is to deprecate the `sort` option.~~
~~Each new sort type would have a descending and ascending option. The other new config option is `sort_ascending` , which controls the order of the fallback id **sort.**~~

The idea is to repurpose the `sort` option to specify how you want to sort your notifications.


fwsmit:
I don't think deprecating the `sort` option is neccesary. You can just add more options to `sort` while making sure the old behavior stays the same. For that it's probably easiest to add all the boolean options to `sort_type_enum_data`

fwsmit:
> I don't think we have many notifications to justify perfomance concerns while sorting, but having multiple notifications_cmp functions might make the code more manageable in the long term.

Don't think there are performance concerns either. We could always optimize later.
 
> My plan was to implement the updated time sort method I wanted, but sorting by notification->start doesn't appear to update the position of the notification when the dup_count increases. Directions welcome.

I don't know. Maybe the notification->start value doesn't get updated correctly when duplicates arrive. You can try printing the value and see if it changes.

duarm:
I'm interested into a multiple sort approach tho, I think sorting by update then sorting by urgency would be pretty useful, since I could keep the most recent at the relative top when switching songs, but a critical notification would still be at the absolute top.

fwsmit:
Hi, what's the status on this PR? Do you think I should review it in its current state, or are you still working on it?

duarm:
> Hi, what's the status on this PR? Do you think I should review it in its current state, or are you still working on it?

Hello, it's ready to review, there are just a few changes and it works, the only thing I'm not sure is about the sort I'm doing in line 193 in queue.c, pretty sure there's a better way of doing that.

There are also clear avenues to improve, like having a different notification_cmp function for each sort type, if you think it's better this way, I can do it rather quickly.

Another nice feature would multi sorts, like I explained on my last comment, but that can be another PR.

src/queues.c
```diff
@@ -190,8 +190,13 @@ int queues_notification_insert(struct notification *n)
         if (!inserted && STR_FULL(n->stack_tag) && queues_stack_by_tag(n))
                 inserted = true;
 
-        if (!inserted && settings.stack_duplicates && queues_stack_duplicate(n))
+        if (!inserted && settings.stack_duplicates && queues_stack_duplicate(n)){
+                if(settings.sort == SORT_TYPE_UPDATE){
```
duarm, line 194: this feels like a hack

dunstrc
```diff
@@ -111,8 +111,15 @@
...
 
-    # Sort messages by urgency.
+    # Sort type.
+    # possible values are:
+    #  * id: sort by id
+    #  * urgency_ascending: sort by urgency (low>normal>critical)
+    #  * urgency_descending: sort by urgency (critical>normal>low)
+    #  * update: sort by update (most recent always at the top)
     sort = yes
+    # Sort fallback ascending, otherwise descending.
```
duarm, line 121: Maybe a better description? What this does is that if two notifications are the same, we sort them by id as a fallback, this controls if this fallback sort is ascending or descending

fwsmit:
Hey, sorry for waiting so long to review this. Could you add the documentation for the new configuration keys so I can check if the code does what you intend it to do (and if the configuration options are sensible).

dunstrc
fwsmit: Yeah you could explain it a bit more thorough in here. But is it neccesary to add a new configuration key for this? Isn't there some more elegant solution for this? Something like integrating the ascending value in the `sort` key, since you already specify `urgency_ascending` and -`descending`. Is there a situation you want to specify `urgency_ascending` and `sort_ascending=false`?

src/queues.c
fwsmit: I don't think the `waiting` queue needs to be sorted. Does the timestamp of the duplicate notification get updated before sorting? If so this should work fine.

dunstrc
duarm: > Is there a situation you want to specify urgency_ascending and sort_ascending=false

If your notifications are anchored to the bottom-right for example, you might want have the sort_ascending to false, so the most recent appears by id appears at the bottom.

> Isn't there some more elegant solution for this?

I think there is, if we implement the multi-sort I mentioned before, we could have something like

sort = urgency_ascending, update, id_descending

duarm:
> Hey, sorry for waiting so long to review this. Could you add the documentation in `dunst.5.pod` for the new configuration keys so I can check if the code does what you intend it to do (and if the configuration options are sensible)?

No problem, I'm not in a hurry

dunstrc
fwsmit: I think multi-sort is a little bit complicated as well. Would it be sufficient to base the ID sorting on the anchor (top or bottom)?

duarm: We can, I'll remove the option then, I struggling to find time to work on this, maybe this weekend

src/queues.c
duarm: I'm not sorting the waiting queue anymore

dunstrc
duarm: Removed the option, now if the origin is set to BOTTOM and the sort type is UPDATE, they appear at the bottom automatically, to preserve the previous behaviour.

dunstrc
```diff
@@ -111,8 +111,15 @@
     #  * anything else will be interpreted as a X color.
     separator_color = frame
 
-    # Sort messages by urgency.
+    # Sort type.
+    # possible values are:
+    #  * id: sort by id
+    #  * urgency_ascending: sort by urgency (low>normal>critical)
```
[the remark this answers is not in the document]
duarm: Well, my reasoning is that CRITICAL is more urgent than LOW, so if you're sorting by URGENCY, critical is greater, so descending starts from CRITICAL.

duarm:
Added the docs to dunst.5.pod as requested, and resolved the conversations.

dunstrc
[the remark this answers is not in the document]
duarm: The description left by S-u-b on #845 also uses the same logic https://github.com/dunst-project/dunst/issues/845#issuecomment-811564933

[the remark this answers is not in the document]
duarm: I see, I changed the wording on the docs, the intention behind the > was "then", not "greather than"

docs/dunst.5.pod
```diff
@@ -288,9 +288,19 @@ Any other value is interpreted as a color, see COLORS
...
+critical first, then normal, then low.
+
+If set to false or id, sort notifications by id.
+
+If set to urgency_ascending, notifications are sorted by urgency, 
+low first, then normal, then critical.
+
+If set to update, notifications are sorted by their update_time. 
+So the most recent is always at the top. This means that if you set sort to update, 
+and stack_duplicates to true, the duplicate will always be at the top.
```
fwsmit, line 304:
```suggestion

When the notification window is at the bottom of the screen, this order is automatically reversed.
```

fwsmit:
I added a small bit of documentation, but other than that it looks good. Could check that bit of documentation and add it? Then I will merge this PR :)

fwsmit:
Sorry for taking a while, but I'll merge this now! Thanks a lot for your work on this

PR #1157, opened 2023-04-01
Fixes some typos on dunst(1)

pliski:
In particular fixes the use of backticks for code formatting.

### DIFF
--- a/docs/dunst.1.pod
+++ b/docs/dunst.1.pod
@@ -33,7 +33,7 @@ Use alternative config file.
 This disables the search for other config files.
 If it cannot be opened, dunst will issue a warning and fall back on its internal
 defaults.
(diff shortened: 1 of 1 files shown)


docs/dunst.1.pod
```diff
@@ -59,7 +59,7 @@ Display a notification on startup.
 
 =head1 CONFIGURATION
 
-A default configuration file is included (usually ##SYSCONFDIR##/dunst/dunstrc)
```
bebehei, line 62: Why do you replace this here with a variable and not let it be the build time constant?

fwsmit: @pliski could you take a look at this comment?

fwsmit:
Closing this, because the reason for this change is unclear. If you can explain it, feel free to reopen

PR #1158, opened 2023-04-04
Get_next_datachange: use `start` to compute timeout

tobast:
Related to issue #1102. Tested hastily, seems to fix the high CPU load.

### DIFF
--- a/src/queues.c
+++ b/src/queues.c
@@ -559,7 +559,7 @@ gint64 queues_get_next_datachange(gint64 time)
         for (GList *iter = g_queue_peek_head_link(displayed); iter;
                         iter = iter->next) {
                 struct notification *n = iter->data;
-                gint64 timeout_ts = n->timestamp + n->timeout;
+                gint64 timeout_ts = n->start + n->timeout;
 
                 if (n->timeout > 0 && n->locked == 0) {
                         if (timeout_ts > time)
--- a/src/queues.c
+++ b/src/queues.c
@@ -503,7 +503,7 @@ void queues_update(struct dunst_status status, gint64 time)
                         continue;
                 }
 
-                n->start = time_monotonic_now();
+                n->start = time;
                 notification_run_script(n);
 
                 if (n->skip_display && !n->redisplayed) {
@@ -538,7 +538,7 @@ void queues_update(struct dunst_status status, gint64 time)
                         if (i_waiting && notification_cmp(i_displayed->data, i_waiting->data) > 0) {
                                 struct notification *todisp = i_waiting->data;
 
-                                todisp->start = time_monotonic_now();
+                                todisp->start = time;
                                 notification_run_script(todisp);
 
                                 queues_swap_notifications(displayed, i_displayed, waiting, i_waiting);


tobast:
One test does not pass: `test_datachange_ttl`. The reason is that this test uses a timestamp that is not `time_monotonic_now` for simplicity, but queues.c:506 calls `time_monotonic_now`, and messes up `n->start` wrt. the timestamp as seen by the test.

tobast:
> One test does not pass: `test_datachange_ttl`. The reason is that this test uses a timestamp that is not `time_monotonic_now` for simplicity, but queues.c:506 calls `time_monotonic_now`, and messes up `n->start` wrt. the timestamp as seen by the test.

Update: it seems fine to use the `time` variable at those locations, so I changed it. The test suite now fully passes.

tobast:
Indeed, I was too hasty. I'm checking why.

tobast:
@fwsmit The test is fixed in the latest revision

src/queues.c
```diff
@@ -555,11 +555,19 @@ gint64 queues_get_next_datachange(gint64 time)
...
         gint64 wakeup_time = G_MAXINT64;
         gint64 next_second = time + S2US(1) - (time % S2US(1));
+        if(next_second - time < 1000) { // In less than 1ms
+                /* Sometimes, the actual sleep will be a few microseconds
+                   shorter than requested. If we were sleeping until next
+                   second, this new run will be just before the turn of second,
+                   and next_second will be less than 5ms away: in this case,
+                   just sleep 1s more. */
+                next_second += S2US(1);
+        }
```
fwsmit, line 565: Why this change? It seems that changing n->timestamp to n->start should fix the issue. With this change you would be skipping a turn of second, so the notifications aren't updated correctly.

tobast: The sleep method used does not guarantee that it will be exact. Sometimes, the sleep will be slightly shorter than expected, and thus the next turn of second will be a few microseconds away only. In this case, that happens once every few dozen sleeps, the next datachange will be reported in less than 1ms away, resulting in a useless call to `run`.

fwsmit: Well, if the change is less than 1ms away, dunst will still not update the seconds counter on the notification, so the next run would be required to do that. Unless you can make it so that the seconds counter is updated even when you are a few ms before the change, I would not do this.

tobast: Fair point. It seems to me that:

1. either we do not care for the occasional useless call to `run`;
1. or the clean solution would be to have two thresholds, eg.:
    * sleep until turn of second + 1ms, which will mean that the vast majority of sleeps will end just after the turn of second (this could be [done here](https://github.com/dunst-project/dunst/blob/836fa0f0289e871f8ba371b5caa6bdbbd99fc59f/src/queues.c#L557));
    * yet update notification age if we are after the turn of second (this is already [done here](https://github.com/dunst-project/dunst/blob/836fa0f0289e871f8ba371b5caa6bdbbd99fc59f/src/notification.c#L676)).

Which option do you prefer? 2 looks cleaner to me, but adds a bit of complexity.

fwsmit: I would suggest another option that works as follows:

- The sleep duration should be until the exact turn of second
- If the next turn of second is less than a few ms away, go to the next second (as is done now, but I would make it more than 1 ms, say 10ms)
- When updating the notification, if the next second is less than the same 10 ms away, update to the next second already. That should be as easy as:

```
gint64 t_delta = time_monotonic_now() - ((n->timestamp+10000) - (n->timestamp+10000) % S2US(1));
```

The above code could maybe be simplified, but adding 10 ms to the timestamp should be the same as checking 10ms later.

tobast: Lgtm. I'll implement that as soon as I get the chance.

fwsmit: Thanks!

tobast: Done in this revision.

fwsmit: The commit message says 5ms, but the code says 10ms. But other than that it seems good :+1:

src/notification.c
```diff
@@ -673,7 +673,8 @@ void notification_update_text_to_render(struct notification *n)
         /* print age */
         gint64 hours, minutes, seconds;
         // Timestamp is floored to the second for display purposes -- see queues.c
-        gint64 t_delta = time_monotonic_now() - (n->timestamp - n->timestamp % S2US(1));
+        gint64 thresholded_timestamp = n->timestamp + TURN_OF_SECOND_THRESHOLD_US;
```
fwsmit, line 676: I think you should subtract here. It should seem as if the notification has been sent TURN_OF_SECOND_THRESHOLD_US earlier, so that the turn of second is a little earlier

tobast: Sorry, I notice I forgot to say that the last revision should have fixed this.

fwsmit:
With the understanding of https://github.com/dunst-project/dunst/issues/1102#issuecomment-1511022163, I want to suggest to split this PR up in 2 parts. One part being fixing the mixup between n->start and n->timestamp. The other part being the feature to be a bit more lenient in the required sleep duration.
I want to release the first part in the next point release to fix the cpu usage issues and the other part can wait till a next version.

Are you willing to split up this PR? It should be as easy as submitting the right commits

tobast:
I just checked, and it seems to be totally possible. The first two commits (5525e99, cd65209) should go together in the point release, as the second one ensures that the tests are still passing; the third one (15b0941) can go in later (this is the one that tries to be clever about turns of second).

tobast:
This third commit is dropped from this PR in the current revision. The new PR #1167 covers it.

fwsmit:
Thanks, I'm merging this now. It would be good to add a few tests to prevent issues like this and similar in the future, but I'll be making that a separate issue

PR #1167, opened 2023-04-17
Turn of second threshold tolerance

tobast:
This patch was first included in PR #1158. The notifications' age is updated at each turn of second, but sometimes, the sleep between two loop turns will be slightly (a fraction of ms most of the time) shorter than what was requested, resulting in a loop run just before the turn of second. In this case, to avoid doing nothing and requesting a very short sleep until the turn of second, this PR introduces a slight tolerance:

* request a sleep until the exact turn of second,
* yet update the notification if the turn of second is very close ahead, and do not wake up until the next turn of second (unless some other event requires it).

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -673,7 +673,9 @@ void notification_update_text_to_render(struct notification *n)
         /* print age */
         gint64 hours, minutes, seconds;
         // Timestamp is floored to the second for display purposes -- see queues.c
-        gint64 t_delta = time_monotonic_now() - (n->timestamp - n->timestamp % S2US(1));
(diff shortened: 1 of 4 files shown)


tobast:
Rebased on master, now that #1158 is merged

fwsmit:
I'm wondering if it's worth to merge this. The advantage of this PR is that it avoids waking up twice in a short timespan. But the disadvantage is that it adds some complexity and has to possibility to add a bug.
What do you think. Should this be merged?

fwsmit:
I'm closing this PR to keep the code simpler. If we need this some time, we can always apply it later

PR #1170, opened 2023-04-23
Add guard against realloc fail in icon-lookup.c

Costinteo:
If realloc fails, its return value will be NULL. In the two calls to realloc in icon-lookup.c, this means the global icon_themes and default_themes_index will be overwritten, producing not only a probable error, but a memory leak as well.

This pull request aims to fix that by adding a couple of guards and keeping the global arrays intact, in case of failure.

Bugs found with cppcheck. It is really minor, I doubt it will ever happen. But it is in theory possible.

### DIFF
--- a/src/icon-lookup.c
+++ b/src/icon-lookup.c
@@ -53,8 +53,14 @@ int load_icon_theme_from_dir(const char *icon_dir, const char *subdir_theme) {
                 return -1;
         }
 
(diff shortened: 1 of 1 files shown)


bebehei:
Please just use g_realloc instead of alloc. Dunst heavily uses GLib and hence usually uses GLib's memory functions. These functions terminate upon allocation failure. (Dunst would not do anything else either)

Costinteo:
Thank you for the reply! I will open a new pull request, so as to not add a commit fixing my initial commit, but rather one unique commit. Seems like changing the commit history does not update the pull request properly.

PR #1182, opened 2023-06-12
Replace realloc with g_realloc in icon-lookup.c

Costinteo:
The purpose is to avoid any failure in realloc, leading to null pointer dereferences/memory leaks. Failure in g_realloc will also terminate Dunst.
@bebehei, waiting for any further suggestions/reviews. :)

### DIFF
--- a/src/icon-lookup.c
+++ b/src/icon-lookup.c
@@ -49,12 +49,12 @@ int load_icon_theme_from_dir(const char *icon_dir, const char *subdir_theme) {
         fclose(theme_index);
         if (ini->section_count == 0) {
                 finish_ini(ini);
-                free(ini);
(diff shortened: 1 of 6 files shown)


bebehei:
Hey @Costinteo, I've checked the code. The problems you found are fixed, but this PR will solve a fraction of the memory problems.

Why is the `calloc` in the same file not problematic? See here: https://github.com/dunst-project/dunst/blob/dfab9f057f89702ade5725c8247a520ab51026f4/src/icon-lookup.c#L67

fwsmit:
`calloc` should be replaced with `gmalloc0_n`, not `gmalloc0`, so I fixed that. Thanks for the PR!

PR #1183, opened 2023-06-18
fixed progress_bar_corner_radius issues

### LINKED ISSUE #1178: progress_bar_corner_radius not working correctly
### Issue description
These are like 3 issues in one:

1- When setting `progress_bar_corner_radius` to a negative value, dunst runs without problems but seems to add that value to the overall size of the bar which wouldn't seem like the expected behavior to me.

2- If the window has a low progress value (like 0) the corners get "inverted". To replicate, simply run: `dunstify "bar radius test" -h int:value:0`
image

3- The `progress_bar_frame_width` property doesn't play well with the corner radius. Using both at the same time results in the inner part of the bar being offset vertically a bit:
image

### Installation info

- Version: `v1.9.2`
- Install type: `manually`
- Window manager / Desktop environment: `BSPWM`
- Distro: `Arch Linux`

<details>

```ini 
    progress_bar = true       
    progress_bar_height = 20                                                               
    progress_bar_frame_width = 5    
    progress_bar_min_width = 0       
    progress_bar_max_width = 380       
    progress_bar_corner_radius = 7    
```
</details>

vatsal999:
fixes  #1178

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -515,7 +515,7 @@ void draw_rounded_rect(cairo_t *c, float x, float y, int width, int height, int 
         if (last) {
                 // bottom right
                 cairo_arc(c,
-                          x + width - corner_radius,
+                          (width < corner_radius) ? x : x + width - corner_radius,
                           y + height - corner_radius,
                           corner_radius,
                           degrees * 0,
@@ -542,7 +542,7 @@ void draw_rounded_rect(cairo_t *c, float x, float y, int width, int height, int 
                           degrees * 270);
                 // top right
                 cairo_arc(c,
-                          x + width - corner_radius,
+                          (width < corner_radius) ? x : x + width - corner_radius,
                           y + corner_radius,
                           corner_radius,
                           degrees * 270,
--- a/src/draw.c
+++ b/src/draw.c
@@ -764,7 +764,7 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, dou
                 cairo_set_line_width(c, frame_width * scale);
                 draw_rounded_rect(c,
                                 frame_x + half_frame_width,
-                                frame_y + half_frame_width,
+                                frame_y,
                                 progress_width - frame_width,
                                 progress_height,
                                 settings.progress_bar_corner_radius,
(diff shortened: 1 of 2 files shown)


src/settings.c
```diff
@@ -207,6 +207,10 @@ void check_and_correct_settings(struct settings *s) {
                         LOG_W("Progress bar corner radius clamped to half of progress bar height (%i).",
                                 progress_bar_max_corner_radius);
                 }
+                if (s->progress_bar_corner_radius < 0) {
+                        settings.progress_bar_corner_radius = 0;
+                        LOG_W("setting progress_bar_corner_radius is negative. Setting progress_bar_corner_radius=0");
```
fwsmit, line 212:
```suggestion
                        LOG_W("setting progress_bar_corner_radius is negative. Setting progress_bar_corner_radius to 0");
```

src/draw.c
```diff
@@ -764,7 +764,7 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, dou
                 cairo_set_line_width(c, frame_width * scale);
                 draw_rounded_rect(c,
                                 frame_x + half_frame_width,
-                                frame_y + half_frame_width,
```
fwsmit, line 767: With cairo you have to specify the center of the line to draw. So you want to place the line half of the width down from where you want the top to be. Otherwise you get blurry lines (see picture where line width is 1). Maybe the fill needs to be moved down.

Screenshot_20230619_213904

src/draw.c
```diff
@@ -515,7 +515,7 @@ void draw_rounded_rect(cairo_t *c, float x, float y, int width, int height, int
         if (last) {
                 // bottom right
                 cairo_arc(c,
-                          x + width - corner_radius,
+                          (width < corner_radius) ? x : x + width - corner_radius,
```
fwsmit, line 518: Why is width also left out here? Why not `x + width` instead of `x`?

fwsmit (changes requested):
Thank you for leaving these fixes! Could you take a look at the review comments I left?

fwsmit:
Hi, could you take a look at the comments I made above?

fwsmit:
With #1227 merged, this should not be needed anymore.

fwsmit:
Thank you for your work, though

PR #1193, opened 2023-08-16
Add multiple pause levels

### LINKED ISSUE #865: Feature Request: Do Not Disturb mode
I'd like to be able to optionally temporarily suppress certain kinds of notifications (Slack, Discord, …). However, I need to not suppress them all (still wish to see urgent notifications about low battery, possibly non-urgent notifications from Calendar), therefore `dunstctl set-paused` isn't sufficient.

I have a few ideas how to implement this:

1. Add a `dnd` mode that is similar to `pause`, but only defers notifications that have a `skip_display_dnd = true` rule
2. Invoke `script`s synchronously and let them somehow set attributes like `skip_display`

I'd be willing to try implementing this, but I'd love to discuss how to best approach this first.

matejdro:
This fixes #865, specifically by introducing multiple pause levels (as outlined in https://github.com/dunst-project/dunst/issues/865#issuecomment-1378273947).

Instead of pause being a toggle, it is now a number and notification rules can set `override_pause_level` which determines minimum pause level at which the notification is displayed.

Old `is-paused` and `set-paused` commands are deprecated, but they still function to maintain backwards compatibility (`set-paused true` sets paused level to 100, which is an arbitrary number, we could pick any other if desired).

PR is not finished yet, I still need to update tests and documentation, but I only edited the core for now to get feedback on my approach. Does this look like something we would want in dunst?

### DIFF
--- a/dunstctl
+++ b/dunstctl
@@ -29,8 +29,9 @@ show_help() {
 	                                    notification with given ID.
 	  history-rm [ID]                   Remove the notification from
 	                                    history with given ID.
-	  is-paused                         Check if dunst is running or paused
-	  set-paused [true|false|toggle]    Set the pause status
+	  is-paused                         Check if pause level is > 0
+	  get-paused-level    Get the current pause level
+	  set-paused-level [level]    Set the pause level
 	  rule name [enable|disable|toggle] Enable or disable a rule by its name
 	  debug                             Print debugging information
 	  help                              Show this help
@@ -122,6 +123,27 @@ case "${1:-}" in
 		fi
 		;;
         "rule")
+		[ "${2:-}" ] \
+			|| die "No rule name parameter specified. Please give the rule name"
+		state=nope
+		[ "${3}" = "disable" ] && state=0
+		[ "${3}" = "enable" ]  && state=1
+		[ "${3}" = "toggle" ]  && state=2
+		[ "${state}" = "nope" ] \
+			&& die "No valid rule state parameter specified. Please give either 'enable', 'disable' or 'toggle'"
+		method_call "${DBUS_IFAC_DUNST}.RuleEnable" "string:${2:-1}" "int32:${state}" >/dev/null
+		;;
+	"get-paused-level")
+		property_get pauseLevel | ( read -r _ _ paused; printf "%s\n" "${paused}"; )
+		;;
+	"set-paused-level")
+		[ "${2:-}" ] \
+			|| die "No status parameter specified. Please give a number as paused parameter."
+		[ "$2" ] && [ -z "${2//[0-9]}" ] \
+			|| die "Please give a number as paused level parameter."
+		property_set pauseLevel variant:uint32:"$2"
+		;;
+	"rule")
 		[ "${2:-}" ] \
 			|| die "No rule name parameter specified. Please give the rule name"
 		state=nope
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -100,6 +100,10 @@ static const char *introspection_xml =
     "            <annotation name=\"org.freedesktop.DBus.Property.EmitsChangedSignal\" value=\"true\"/>"
     "        </property>"
 
+    "        <property name=\"pauseLevel\" type=\"u\" access=\"readwrite\">"
+    "            <annotation name=\"org.freedesktop.DBus.Property.EmitsChangedSignal\" value=\"true\"/>"
+    "        </property>"
+
     "        <property name=\"displayedLength\" type=\"u\" access=\"read\" />"
     "        <property name=\"historyLength\" type=\"u\" access=\"read\" />"
     "        <property name=\"waitingLength\" type=\"u\" access=\"read\" />"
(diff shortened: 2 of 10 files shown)


fwsmit:
Your suggestion for adding a pause level seems like a good approach. I would be careful with adding too many pause levels, as it could make it overwhelming for people. I'd suggest adding a maximum of 10 pause levels, since I cannot think of a situation that requires more pause levels. The limit could always be raised later.

dunstctl
```diff
@@ -29,8 +29,9 @@ show_help() {
 	                                    notification with given ID.
 	  history-rm [ID]                   Remove the notification from
 	                                    history with given ID.
-	  is-paused                         Check if dunst is running or paused
-	  set-paused [true|false|toggle]    Set the pause status
+	  is-paused                         Check if pause level is > 0
+	  get-paused-level    Get the current pause level
```
fwsmit, line 33: You removed `set-paused` here. I would keep `set-paused` as an easier option for pausing notifications for people who don't use pause levels. So I would not deprecate `is-paused` and `set-paused`.

src/dbus.c
```diff
@@ -949,15 +955,29 @@ gboolean dbus_cb_dunst_Properties_Set(GDBusConnection *connection,
                                       GError **error,
                                       gpointer user_data)
 {
+        int targetPauseLevel = -1;
         if (STR_EQ(property_name, "paused")) {
-                dunst_status(S_RUNNING, !g_variant_get_boolean(value));
+                if (g_variant_get_boolean(value)) {
+                        targetPauseLevel = 100;
```
fwsmit, line 961: It may be useful to define a variable for `MAX_PAUSE_LEVEL`, set it to 10 and use it here.

fwsmit:
Your approach is good and the implementation seems mostly good. Could take a look at the build errors, my comments and the documentation?

matejdro:
Thanks for the review. Added several commits which should address things.

roobre:
Hi there!
Just wanted drop by and thank you both for your work on this feature! My search engine led me here when searching how to mute/pause certain notifications only.

I'm runing dunst compiled from this branch, with two simple `set-pause-level` schedules using systemd user timers and so far it's working great.

If I could make a small suggestion, from the example `dunstrc` config it was not immediately clear to me that `override_pause_level` can (and probably should?) go inside rules. Perhaps it might be worth adding it to the list of things that can be overridden in the default config file. The new option is also not present in the man page.

Thanks again for your work on this! ❤️

dunstrc
```diff
@@ -228,6 +228,12 @@
     # Maximum amount of notifications kept in history
     history_length = 20
 
+    ### DND ###
```
fwsmit, line 231: As @roobre noted in their comment, it's better to move this example to the rules section. You could even set some basic defaults like 30 for normal notifications and 60 for urgent notifications

dunstctl
```diff
@@ -29,8 +29,10 @@ show_help() {
 	                                    notification with given ID.
 	  history-rm [ID]                   Remove the notification from
 	                                    history with given ID.
-	  is-paused                         Check if dunst is running or paused
-	  set-paused [true|false|toggle]    Set the pause status
+	  is-paused                         Check if pause level is > 0
+	  set-paused [true|false|toggle]    Set the pause status	  
+	  get-paused-level                  Get the current pause level
```
fwsmit, line 34: I would rename `get-paused-level` and `set-paused-level` to `get-pause-level` and `set-pause-level` (without the d's) for more correct grammar.

fwsmit (changes requested):
Thanks for your improvements! The code looks good, but there are a few small documentation and naming improvements needed before this can be merged. Could you add an entry for `override_pause_level` in the man page as well?

docs/dunst.5.pod
```diff
@@ -553,6 +553,16 @@ Ignore the dbus closeNotification message. This is useful to enforce the timeout
 set by dunst configuration. Without this parameter, an application may close
 the notification sent before the user defined timeout.
 
+=item B<override_pause_level> (values: [0-100], default: 0)
+
+A Notification will appear whenever notification's 
+override_pause_level >= dunst's paused level. Setting this to values other than 0
+allows you to create partial pause modes, where more urgent notifications get through,
+but less urgent stay paused. For example, when you can set a low battery noficiation's
+override_pause_level to 2 and then set dunst's pause level to 2. This will cause dunst to
```
fwsmit, line 562: This documentation is very clear. But given the defaults for normal and critical level notifications, it's better to give another example value to battery notifications. Maybe 60 (same as critical)

matejdro: Updated that to 60

fwsmit: Thanks

fwsmit:
Thank you for this new feature. I will merge this as soon as the tests pass

igorline:
Cannot get this to work properly

have this in my `dunstrc` now
```
[Spotify]
    appname = Spotify
    background = "#FF0000"
    override_pause_level = 30
```

when I try to set pause level to 100 messages still go through with expected red background

`dunstctl set-pause-level 100`

matejdro:
Huh. If you remove override pause level, do messages still go through?

igorline:
yeah, tried various combinations and still get notifications bypassing

matejdro:
Which version do you have? This was just released in Dunst v10.0.

igorline:
I installed it from `master` branch

matejdro:
Are you setting override pause level to a higher value somewhere else (default configuration)?

I'm actually not sure what would be a good way to debug this. Maybe we should have added some more LOG_D statements for this?

PR #1203, opened 2023-09-14
Changes "the" to "them" in man pages 5 under the "follow" heading

areif-dev:
This is a quick fix for a grammatical error in the docs. In man pages 5 under the "Global section: follow" heading, change the sentence "When either of the is used..." to "When either of them is used..."


bebehei:
Deleted the last useless merge commit.

PR #1208, opened 2023-10-11
Fix commandline handling

zappolowski:
The output of `dunst -h` differs from the one of `dunst --help` due to how the usage string is constructed. Furthermore, the options `-print` and `--startup_notification` are not listed at all (due to incorrect ordering of options).

This PR refactors `cmdline_find_option` to handle more than two alternatives (which fixes the first issue, when calls to `cmdline_get_*` are merged) and reorders command line parsing in a way, that all currently available options are listed.

NB: I've added `-startup_notification` as alternative to `--startup_notification` as it was the only option only available with to hyphens and thus breaking consistency.


fwsmit:
That's a great find! It even simplifies the code :)

Could you add the new command line options to the `dunst.1.pod` man page?

fwsmit:
Thanks for the improvements! This will be merged now

PR #1209, opened 2023-10-12
Add fish completions

zappolowski:
Add completion files for all commands for [fish-shell](https://github.com/fish-shell/fish-shell).

NB: I left out completion for the `rule` subcommand of `dunstctl` as there currently is no way of getting defined rules in a generic way (e.g. both the completions for bash and zsh try to parse the default config, which might not be the configuration being used and also ignores drop-in files altogether).


fwsmit:
Thanks for your contribution! You're on a roll with PR's!

PR #1210, opened 2023-10-14
Resolve paths for default icon

### LINKED ISSUE #1173: Can't set default_icon path with environment variable ($HOME or ~)
### Issue description

The `default_icon` path cannot use the `$HOME` or `~` environment variables. The full path must be specified to for it to work. E.x.: `/home/user/.local/share/icons/my-icon.svg`

### Installation info

- Version: `1.91`
- Install type: [Arch package](https://archlinux.org/packages/community/x86_64/dunst/)
- Distro: `Arch Linux`

<details>

```ini
[urgency_low]
    # IMPORTANT: colors have to be defined in quotation marks.
    # Otherwise the "#" and following would be interpreted as a comment.
    background = "#222222"
    foreground = "#888888"
    timeout = 10
    # Icon for notifications with low urgency, uncomment to enable
    default_icon = $HOME/.local/share/icons/urg-low.svg

[urgency_normal]
    background = "#285577"
    foreground = "#ffffff"
    timeout = 10
    default_icon = ~/.local/share/icons/urg-norm.svg

[urgency_critical]
    background = "#900000"
    foreground = "#ffffff"
    frame_color = "#ff0000"
    timeout = 0
    default_icon = $HOME/.local/share/icons/urg-crit.svg
```
</details>

### Additional information

The config above is copied from the example config: lines 313-335. This is identical to my `dunstrc`, minus the actual `default_icon` path and colors.

To resolve this issue, I have tried things such as replacing `default_icon` with `new_icon` and `icon`. I have also tried changing the `icon_path`, which points to the directory holding the images for the icons.

According to #520 this issue was fixed and merged 4 years ago, so I'm not sure why I can't get it working.

I can confirm that the environment variables work in a shell script using `dunstify`.

zappolowski:
This (partially) fixes #1173 - support for expanding variables is done in #1215 

`default_icon` was erroneously declared as being a plain string while it can also be a full path. Thus changing its type should solve the described issue already (without any side effects I could currently see).


fwsmit:
It seems a little weird to add custom code for expanding the `$HOME` environment variable. This might also behave differently than people expect. I think it's better to look at the `wordexp()` function to see if we can use this function for expanding like a shell would. Otherwise just having the `~` expansion seems better.

zappolowski:
I was on the edge whether I should expand `$HOME` at all ... but I'm also in favor to remove it again and just expand `~/` in the beginning of a string.

`wordexp` could also expand commands like ``default_icon = `rm -rf ~`/asdf/`` if I'm not mistaken and thus I would either go for a white-list appraoch of environment variables to expand (like `$HOME`, `$XDG_CONFIG_DIR`, `$XDG_DATA_DIR` and alike) or just leave it as is.

zappolowski:
I've implemented a version of the resolution using `wordexp`. One thing to decide: shall we use `WRDE_UNDEF` or not?

1. If it is used, any undefined variable fails the expansion and the input is returned unaltered (with a warning).
2. If it is not used, it is replaced with an empty string, e.g. `$XDG_CONFIG_HOME/dunst` just becomes `/dunst`. But using it allows us to use special syntax like `${XDG_CONFIG_HOME:-$HOME/.config}/dunst`.

fwsmit:
I think it's better to keep the `~` expansion fix and the new feature of using `wordexp` to separate PR's. Could you create a new PR for the `wordexp` stuff? We can discuss it further there

fwsmit:
Thanks, this makes it easy to review :)

PR #1212, opened 2023-10-17
Export rules via dbus

zappolowski:
As noted in #1209 a completion for `dunstctl rule` is currently not feasible due to missing information about which rules exist. This PR proposes to add a new DBus method to list all configured rules and use this information for a proper completion.

NB: Completion for bash and zsh will be adjusted when this approach is considered okay. Also, I'd add more information about the rules in the response in this case.

- [x] add new subcommand to dunstctl(1)

### DIFF
--- a/dunstctl
+++ b/dunstctl
@@ -34,7 +34,8 @@ show_help() {
 	  get-pause-level                   Get the current pause level
 	  set-pause-level level             Set the pause level
 	  rule name enable|disable|toggle   Enable or disable a rule by its name
-	  rules                             Displays configured rules (in JSON)
+	  rules [--json]                    Displays configured rules (optionally
+	                                    in JSON)
 	  debug                             Print debugging information
 	  help                              Show this help
 	EOH
@@ -125,8 +126,23 @@ case "${1:-}" in
 		fi
 		;;
 	"rules")
-		busctl --user --json=pretty --no-pager call org.freedesktop.Notifications /org/freedesktop/Notifications org.dunstproject.cmd0 RuleList 2>/dev/null \
-			|| die "Dunst is not running."
+		case "${2:-}" in
+			"" | --json)
+				busctl --user --json=pretty --no-pager call "${DBUS_NAME}" "${DBUS_PATH}" "${DBUS_IFAC_DUNST}" RuleList \
+				| {
+					if [ "${2:-}" = '--json' ]
+					then
+						cat
+					else
+						jq --raw-output '.data[][] | ["[\(.name.data)]"], [to_entries[] | select(.key != "name") | "    \(.key) = \(.value.data)"] | join("\n")'
+					fi
+				} \
+					|| die "Dunst is not running."
+			;;
+			*)
+				die "Unknown format \"${2}\". Please use either \"--json\" or no option at all."
+			;;
+		esac
 		;;
 	"rule")
 		[ "${2:-}" ] \
--- a/dunstctl
+++ b/dunstctl
@@ -34,6 +34,7 @@ show_help() {
 	  get-pause-level                   Get the current pause level
 	  set-pause-level level             Set the pause level
 	  rule name enable|disable|toggle   Enable or disable a rule by its name
+	  rules                             Displays configured rules (in JSON)
 	  debug                             Print debugging information
 	  help                              Show this help
 	EOH
@@ -122,6 +123,10 @@ case "${1:-}" in
 		else
 			property_set paused variant:boolean:"$2"
 		fi
+		;;
+	"rules")
+		busctl --user --json=pretty --no-pager call org.freedesktop.Notifications /org/freedesktop/Notifications org.dunstproject.cmd0 RuleList 2>/dev/null \
+			|| die "Dunst is not running."
 		;;
 	"rule")
 		[ "${2:-}" ] \
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -6,12 +6,14 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#include "draw.h"
 #include "dunst.h"
 #include "log.h"
 #include "menu.h"
 #include "notification.h"
 #include "queues.h"
 #include "settings.h"
+#include "settings_data.h"
 #include "utils.h"
 #include "rules.h"
 #include "option_parser.h"
@@ -94,6 +96,9 @@ static const char *introspection_xml =
     "        <method name=\"RuleEnable\">"
     "            <arg name=\"name\"     type=\"s\"/>"
     "            <arg name=\"state\"    type=\"i\"/>"
+    "        </method>"
+    "        <method name=\"RuleList\">"
+    "            <arg direction=\"out\" name=\"rules\"           type=\"aa{sv}\"/>"
     "        </method>"
     "        <method name=\"Ping\"                  />"
 
@@ -187,6 +192,7 @@ DBUS_METHOD(dunst_NotificationPopHistory);
 DBUS_METHOD(dunst_NotificationRemoveFromHistory);
 DBUS_METHOD(dunst_NotificationShow);
 DBUS_METHOD(dunst_RuleEnable);
+DBUS_METHOD(dunst_RuleList);
 DBUS_METHOD(dunst_Ping);
 static struct dbus_method methods_dunst[] = {
         {"ContextMenuCall",                     dbus_cb_dunst_ContextMenuCall},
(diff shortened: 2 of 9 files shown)


dunstctl
```diff
@@ -31,6 +31,7 @@ show_help() {
 	                                    history with given ID.
 	  is-paused                         Check if dunst is running or paused
 	  set-paused [true|false|toggle]    Set the pause status
+	  rules                             Displays configured rules (in JSON)
```
fwsmit, line 34: I know the history command already formats in json by default, but I'm wondering if it's not better to format in human-readable format by default and allow for a json toggle.

fwsmit (commented):
Your approach looks good. I have only one comment.

fwsmit:
Please add some tests as well. Our dbus interface is currently very well tested

fwsmit:
These shell completions look very promising. There's so many exciting things you can do with it :smile:

dunstctl
zappolowski: This should be easily feasible.

My current idea is to have:
```
	  rules (as-json|human-readable)    Displays configured rules
```
with `human-readable` being the default when omitted. I don't know for sure, what's the correct notation for denoting an optional argument (or should it be mandatory, though I think it would be nice to have a default for interactive usage) and just used parentheses as an example.

How should the human-readable form look like? One could just emit ini-style information, basically recreating the original input.

fwsmit: I would prefer to use an argument like `--json` for json output. And it should definitely not be required, for the reason you mentioned. 

Optional commands are put in [square brackets].

fwsmit: I don't know what the best human-readable output is, but I would just look at what's easy to create. On second thought, the pretty json output might be good enough as well

fwsmit: I currently don't have a good test setup to see, so you'll have to judge for yourself what's best

zappolowski: Sure, I can do `--json` as well. I just wanted to be consistent how other commands are used.

> Optional commands are put in [square brackets].
Yes, that's how I would've expected it as well, but the usage output doesn't use it this way consistently. `history-pop` and `history-rm` look the same, but just `history-pop` uses it as an optional argument. `set-paused` and `rule` also use them for mandatory arguments.

fwsmit: > Sure, I can do `--json` as well. I just wanted to be consistent how other commands are used.

You don't have to do it just because I mentioned it. Only if you think it's a good idea. Consistency between different commands is also nice.

> > Optional commands are put in [square brackets].
> 
> Yes, that's how I would've expected it as well, but the usage output doesn't use it this way consistently. `history-pop` and `history-rm` look the same, but just `history-pop` uses it as an optional argument. `set-paused` and `rule` also use them for mandatory arguments.

Oh that's a little bit weird. You would only need one of these commands. You can fix that in a separate PR if you want

fwsmit:
Let me know if you'd like me to give it another round of review

fwsmit:
Are you planning to finish this PR? No pressure, you don't have to. But the feature seems cool

zappolowski:
Yes, I will try to finish this, though I cannot give a timeline when this will happen. Hopefully, within the next two weeks I will find some time to work on this (and the other PR).

zappolowski:
@fwsmit I've implemented a human-readable output as well. If you agree with this, I'd continue to look into tests and also add all the other missing fields to both calls.

dunstctl
```diff
@@ -123,6 +125,20 @@ case "${1:-}" in
...
+		case "${2:-}" in
+			"")
+				method_call "${DBUS_IFAC_DUNST}.RuleListHumanReadable" | sed '1s/^[[:space:]]\+//'
+			;;
+			--json)
+				busctl --user --json=pretty --no-pager call "${DBUS_NAME}" "${DBUS_PATH}" "${DBUS_IFAC_DUNST}" RuleList 2>/dev/null \
+					|| die "Dunst is not running."
+			;;
+			*)
+				die "Unknown format \"${2}\". Please use either \"as-json\" or \"human-readable\"."
```
fwsmit, line 138: This help message is outdated

fwsmit:
> @fwsmit I've implemented a human-readable output as well. If you agree with this, I'd continue to look into tests and also add all the other missing fields to both calls.

With the human readable printing I was not thinking to put it over dbus. Dbus is meant to be a stable interface, where as the human readable output could change at any time. I would prefer it if the dbus output was converted to a human-readable output by the script. The human-readable part is also just an idea. If you think it's readable enough from the json output, or that it's not necesary at all, you can also drop it.

> Edit: ~Just saw, that the build fails as `GStrvBuilder` is not available ... this was released with glib 2.68 almost three years ago. Can we switch to more recent versions (see my PRs on [the other project](https://github.com/dunst-project/docker-images)) or shall I find some alternative solution?~

Yeah, the docker images should probably be updated to remove EOL distro's

zappolowski:
> With the human readable printing I was not thinking to put it over dbus. Dbus is meant to be a stable interface, where as the human readable output could change at any time. I would prefer it if the dbus output was converted to a human-readable output by the script. The human-readable part is also just an idea. If you think it's readable enough from the json output, or that it's not necesary at all, you can also drop it.

I could try to come up with some jq magic here as well. But then I think, that jq should be made a (at least) optional dependency of this project. The completion for zsh already uses it unconditionally and also some other scripts in contrib rely on it.

I'm also fine with dropping a human-readable variant as of now and maybe add it later on when there's an actual need for it.

zappolowski:
> Please add some tests as well. Our dbus interface is currently very well tested

I've added some simple test. I hope that is enough and not all branches need to be tested separately.

Btw. none of the methods in the `org.dunstproject.cmd0` interface are tested yet (mine being the first one ... as can be seen by the need of creating a new function to reach that interface).

Another note, I assume the call to `g_variant_type_new` in `dbus_cb_dunst_NotificationListHistory` leaking memory, but don't want to bring it into this PR.

zappolowski:
@fwsmit: Do you consider having a specified order in the output of the JSON important (currently I'm listing filter rules first and then the adjusted values). I could shorten the code quite drastically by using `g_variant_dict_*` but then would lose that order. On the other hand, one should probably not assume any order in a JSON response anyhow (which means I will take another look at the jq-Script for the human readable output).

src/dbus.c
```diff
@@ -439,6 +445,145 @@ static void dbus_cb_dunst_NotificationRemoveFromHistory(GDBusConnection *connect
...
+                if (r->default_icon)
+                        g_variant_dict_insert(&dict, "default_icon", "s", r->default_icon);
+                if (r->new_icon)
+                        g_variant_dict_insert(&dict, "new_icon", "s", r->new_icon);
+                if (r->script)
+                        g_variant_dict_insert(&dict, "script", "s", r->script);
+                if (r->set_stack_tag)
+                        g_variant_dict_insert(&dict, "set_stack_tag", "s", r->set_stack_tag);
+                if (r->override_pause_level != -1)
+                        g_variant_dict_insert(&dict, "override_pause_level", "i", r->override_pause_level);
```
fwsmit, line 575: In this function, is it not possible to use the data in settings_data.h to programmatically fill the dictionary? All rules in the settings data struct have a non-zero rule (except probably for the first one, if that's a problem that's fixable).
This would make it more easy to maintain. Otherwise it's easy to forget to add it here after adding a rule. If you have idea's for how to do this, please let me know. Then we can discuss if it's worth doing.

zappolowski: My knowledge of C is rather limited, but if I got you right, this would mean compile time code generation, which would mean using macros. I'm not 100% sure on this, but I would highly doubt, that this is possible. Run-time code generation via reflection is not possible in C AFAIK.

My (maybe overly simplistic) approach: just write some comment to the definition of rule to remind the implementer to add it to `dbus_cb_dunst_RuleList` as well.

fwsmit: Okay, could you then add some documentation to settings_data.h?

bynect: maybe I can make it automatic when I refactor rule parsing. for now it is fine though

bynect:
Everything looks fine. Maybe you could add dunstctl rules to the manpage.

> NB: Completion for bash and zsh will be adjusted when this approach is considered okay. Also, I'd add more information about the rules in the response in this case.

not sure what you mean by that

zappolowski:
> not sure what you mean by that

This is outdated from the time when the output just consisted of some assorted fields.

But I will adjust the other shell completions as well as add the new subcommand to dunstctl's man page.

zappolowski:
I've updated the man page as well as some completions. I've tested the bash completion (and found some issues which I will fix in another PR) but the ones for zsh are dry-coded (and thus kept minimal).

bynect (approved):

PR #1215, opened 2023-10-23
Use wordexp to resolve variables like $HOME in path like settings

### LINKED ISSUE #1173: Can't set default_icon path with environment variable ($HOME or ~)

zappolowski:
This fixes #1173 - in combination with #1210

This is a spin-off of #1210 to properly handle environment variables for defining path like variables.

Open questions:
* Which flags to pass to `wordexp`?
  * `0` (like e.g. i3wm does) - allows for a maximum of flexibility. E.g. the output of a script could be used to set the variable (`default_icon = $(some_random_command)`), but this also allows for arbitrary code execution.
  * `WRDE_NOCMD` - prohibits expansion of commands, but allows for all variables - even undefined ones.
  * `WRDE_NOCMD | WRDE_UNDEF` - strictest way of expansion, which ignores undefined variables
* Is `LOG_W` enough to inform user of failed expansion? Should this be more detailed wrt. kind of failure?

### DIFF
--- a/test/utils.c
+++ b/test/utils.c
@@ -1,3 +1,5 @@
+#include <features.h>
+
 #include "../src/utils.c"
 #include "greatest.h"
 
@@ -152,6 +154,7 @@ TEST test_string_to_path(void)
         ASSERT_STR_EQ(exp, (ptr = string_to_path(g_strdup(exp))));
         free(ptr);
 
+        // This might fail, when a user named path exists on the host running the tests.
         exp = "~path/with/wrong/tilde";
         ASSERT_STR_EQ(exp, (ptr = string_to_path(g_strdup(exp))));
         free(ptr);
@@ -165,6 +168,19 @@ TEST test_string_to_path(void)
                       (ptr = string_to_path(g_strdup("~/.path/with/tilde and some space"))));
         free(exp);
         free(ptr);
+
+        ASSERT_STR_EQ((exp = g_strconcat(home, "/.path/with/HOME environment variable", NULL)),
+                      (ptr = string_to_path(g_strdup("$HOME/.path/with/HOME environment variable"))));
+        free(exp);
+        free(ptr);
+
+// Just glibc properly returns an error when using `WRDE_UNDEF` and an
+// undefined variable is found. musl accepts this flag and ignores it.
+#ifdef __GLIBC__
+        exp = "/some/$UNDEFINED/variable";
+        ASSERT_STR_EQ(exp, (ptr = string_to_path(g_strdup(exp))));
+        free(ptr);
+#endif
 
         PASS();
 }
--- a/.valgrind.suppressions
+++ b/.valgrind.suppressions
@@ -6,6 +6,19 @@
    obj:/lib/ld-musl-x86_64.so.1
 }
 
+# Ignore another musls' weird error
+{
+   musl_alpine_libc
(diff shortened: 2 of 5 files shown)


zappolowski:
I agree that command substitution is not that important at the moment and adding some functionality later on is easier than removing it (and breaking configs).

> It would be nice if the error message would be a bit more specific depending on the return value of wordexp.

I've expanded the warnings for each possible case. Please check, whether that suffices. I've also expanded the documentation where paths are expanded (basically, `browser` and `dmenu` also use paths, but I left them out as they most probably are absolute paths anyhow).

PS: I have to investigate why the build failed ... but it's again something on alpine.

zappolowski:
@fwsmit any further actions required?

fwsmit:
I'll take a look once I have the time

test/utils.c
```diff
@@ -166,6 +169,18 @@ TEST test_string_to_path(void)
         free(exp);
         free(ptr);
 
+        ASSERT_STR_EQ((exp = g_strconcat(home, "/.path/with/HOME environment variable", NULL)),
+                      (ptr = string_to_path(g_strdup("$HOME/.path/with/HOME environment variable"))));
+        free(exp);
+        free(ptr);
+
+// At least musl doesn't properly support wordexp.
```
fwsmit, line 177: That seems like a problem, since we do support musl

zappolowski: The problem here is, that `wordexp` on musl just does not expand the value if the variable is unset. In case the variable is know, it works as intended (the other tests in this file pass on musl as intended).

fwsmit: Ah okay. Could you clarify the comment then?

fwsmit:
Thanks, the implementation and documentation look good. I'll go ahead and merge it.

Issue #1207, opened 2023-10-10
[FR] Search on user's `dunstrc.d`, even if no `dunstrc` found


pataquets:
### Issue description

As I'm tuning my dunstrc config, I'd like to just use `dunstrc.d/*.conf` files and just use the installed, default `dustrc` file as basis.
However, I've found that placing my `dunstrc.d` dir in my `~/.config/` dir, without a matching `~/.config/dunstrc` file, causes the `.d` dir to be ignored. As soon as I copy the default-installed `dunstrc` file to my home config, overrides work.

I'd like to just have the `.d/*.conf` files in my config dir and get the base defaults from the installed files, so I don't have to track the base `dunstrc` config defaults from version to version.

### Installation info

- Version: `v1.9.2-14-g2cd719a`
- Install type: `built from source`
- Window manager / Desktop environment: `i3` (X11, from repos)
- Distro: `Ubuntu Jammy`

fwsmit:
Possible solutions are using /etc/dunstrc.d/.conf or creating an empty ~/.config/dunstrc.

bynect:
now multiple config files can be specified with `-conf`. also you could make an empty dunstrc like friso said

PR #1220, opened 2023-10-25
Allow multiple value for `-config` option

bynect:
This pr adds the ability to accept multiple paths for the -config option and also to handle directory paths.

Directories are simply scanned for drop-ins and if there is a dunstrc file it is loaded. Only the first dunstrc is loaded (for directories), unless the user purposefully specified multiple dunstrc in -config. 

I updated the tests and did some manual testing and so far it seems to work nicely.
This may be useful to resolve problems like that in #1207, where we want to load drop-ins and config from different directories.


fwsmit:
If I read the issue right, they don't need the option to add more items to the -config option, but only that dunst scans the ~/.config/dunstrc.d/*.conf files from the home directory, even if the ~/.config/dunstrc is not present. That seems like a simpler solution.
You could also just use /etc/dunstrc.d/*.conf or an empty ~/.config/dunstrc.

bynect:
> If I read the issue right, they don't need the option to add more items to the -config option, but only that dunst scans the ~/.config/dunstrc.d/*.conf files from the home directory, even if the ~/.config/dunstrc is not present. That seems like a simpler solution.
> You could also just use /etc/dunstrc.d/*.conf or an empty ~/.config/dunstrc.

Yes, I proposed this pr for providing a more general solution for people that (for testing or whatever) want to split their config files on different directories. Even if the other solution is easier, the difference in code would be not that much, so adding this should be worthwhile (since I already wrote it also 😁)

fwsmit:
> > If I read the issue right, they don't need the option to add more items to the -config option, but only that dunst scans the ~/.config/dunstrc.d/_.conf files from the home directory, even if the ~/.config/dunstrc is not present. That seems like a simpler solution.
> > You could also just use /etc/dunstrc.d/_.conf or an empty ~/.config/dunstrc.
> 
> Yes, I proposed this pr for providing a more general solution for people that (for testing or whatever) want to split their config files on different directories. 

I haven't seen this in other tools, so it seems to not be a standard thing people want.

> Even if the other solution is easier, the difference in code would be not that much, so adding this should be worthwhile (since I already wrote it also 😁)

I get that you already put the effort in this code, but it's also a command line option that dunst needs to maintain indefinitely. So it can add some maintenance burden. That's why I lean more towards simpler options

bynect:
> > > If I read the issue right, they don't need the option to add more items to the -config option, but only that dunst scans the ~/.config/dunstrc.d/_.conf files from the home directory, even if the ~/.config/dunstrc is not present. That seems like a simpler solution.
> > > You could also just use /etc/dunstrc.d/_.conf or an empty ~/.config/dunstrc.
> > 
> > Yes, I proposed this pr for providing a more general solution for people that (for testing or whatever) want to split their config files on different directories. 
> 
> I haven't seen this in other tools, so it seems to not be a standard thing people want.
> 
> > Even if the other solution is easier, the difference in code would be not that much, so adding this should be worthwhile (since I already wrote it also 😁)
> 
> I get that you already put the effort in this code, but it's also a command line option that dunst needs to maintain indefinitely. So it can add some maintenance burden. That's why I lean more towards simpler options

From what I know a lot of tool allow specifying more than one config file (either with cli options or imports in the files or etc). If you don't like the list method it could be changed to a multiple config options, like

```
Dunst -config A,B,C
# to
Dunst -config A -config B -config C
```

But this is just a way to do this so let me know if you know a better method

bynect:
I'm going to close this, maybe I'll come up with some tweaks in the future

Issue #1211, opened 2023-10-17
Multiple fonts not working together


astridbunny:
### Issue description
I want to use a symbol font together with a text font for notifications.
If I put the fonts in the order: 
`font = "mononoki 8, Material Icons 8"`
Text does not show up at all, only the icons.
However, the order:
`font = "Material Icons 8, mononoki 8"`
Has the opposite problem: icons will not show up at all. I'm unsure what could be causing this.

### Installation info

- Version: `<!-- output of dunst -v -->`
Dunst - A customizable and lightweight notification-daemon 1.9.2 (2023-04-20)
- Install type: `<!-- [package|manually|...] -->`
Package manager
- Window manager / Desktop environment: `<!--  -->`
i3-gaps
- Distro: `<!--  -->`
Void Linux

astridbunny:
"Fixed" the issue by merging the fonts in FontForge, but still unsure why this is happening.

bynect:
I have the same problem with fontawesome. I think that it's related to the way pango loads fonts. Needs further investigation

bynect:
Pango cairo seems to support loading only one font at a time. I made a pr for font fallback that should mitigate this (given that fontconfig is correctly configured). Alternatively specifying which font to use with pango markup should also work

fwsmit:
This should be mostly fixed with #1222. Are there any issues remaining?

bynect:
I have looked more into it, and this seems by design. Pango is made so that if you want to use multiple fonts you have to specify the fallbacks with fontconfig and they will be used as necessary, so that the font is not chosen at the application level but globally.

I will close this as we can't really do anything else without abandoning pango

PR #1222, opened 2023-10-26
Enable pango font fallback

bynect:
This is a simple addition to text rendering that allows fallback fonts. This improved my experience with #1211 when using a monospace with font awesome glyphs.

However this is not a complete solution! The problem with multiple fonts is in pango design.
It allows only one font to be loaded when rendering. In this patch I simply add the pango attribute for font fallback.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -76,7 +76,9 @@ void draw_setup(void)
 
         win = out->win_create();
 
+        LOG_D("Trying to load font: '%s'", settings.font);
         pango_fdesc = pango_font_description_from_string(settings.font);
+        LOG_D("Loaded closest matching font: '%s'", pango_font_description_get_family(pango_fdesc));
 
         if (settings.enable_recursive_icon_lookup)
                 load_icon_themes();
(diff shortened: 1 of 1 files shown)


bynect:
I just noticed that my editor removed some space 😅

src/draw.c
```diff
@@ -745,16 +749,16 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, dou
 
                 /* Draw progress bar
                 * TODO: Modify draw_rounded_rect to fix blurry lines due to fractional scaling
-                * Note: the bar could be drawn a bit smaller, because the frame is drawn on top 
+                * Note: the bar could be drawn a bit smaller, because the frame is drawn on top
                 */
                 // left side (fill)
                 cairo_set_source_rgba(c, cl->highlight.r, cl->highlight.g, cl->highlight.b, cl->highlight.a);
-                draw_rounded_rect(c, x_bar_1, frame_y, progress_width_1, progress_height, 
+                draw_rounded_rect(c, x_bar_1, frame_y, progress_width_1, progress_height,
```
fwsmit, line 756: Why is there a change in progress bar drawing?

bynect: I just noticed that my editor removed some space 😅 ^^

bynect:
Looking back at it, I am quite at loss as to why the ci failed

fwsmit:
> Looking back at it, I am quite at loss as to why the ci failed

It seems the test failed because of some other reason. Pushing some code should re-run the test (I cannot find the button to re-run).

fwsmit:
Does it fallback to the second font in the list or to some other font on the system?
It _should_ fallback to the second font in the system, but I'm not sure why that's not happening. Do you have an idea?

bynect:
> Does it fallback to the second font in the list or to some other font on the system?
> It _should_ fallback to the second font in the system, but I'm not sure why that's not happening. Do you have an idea?

According to pango docs it should fallback using system font settings (so ignoring the given font list). But it seems like this works only on supported systems (fontconfig and maybe others).

fwsmit:
@bynect could you push this branch, so the CI runs again? You might need to make a small change, which you can revert afterwards.

bynect:
@fwsmit I removed the change in the bar code and force pushed

fwsmit:
> @fwsmit I removed the change in the bar code and force pushed

Thanks! The tests pass now, so I'll go ahead an merge!

Issue #1225, opened 2023-11-02
Move to scdoc


apprehensions:
Currently, dunst uses doxygen/pod2man, which is pretty bloated.

scdoc is a very simple man page generator with a much readable markdown format than pod.

 I am willing to rewrite the docs in markdown, which itself isnt that hard anyway.

apprehensions:
https://git.sr.ht/~sircmpwn/scdoc 

many simple projects such as dunst use scdoc or manually written man pages.

fwsmit:
I don't think the pod format is too hard to read, but I'm not against replacing it either.

apprehensions:
scdoc is easier to package as it is a small C program. Pod on the other hand requires a whole new language, and its own formatting language instead of Markdown.

apprehensions:
Alongside this change, the docs for the dunst website may have to be redone.

Please consider having a simple suckless website such as https://soju.im/ https://soju.im/doc/soju.1.html

bebehei:
A few years ago, I would have supported this to 150%. But especially in the last year, I have learned that docs mostly aren't as simple as markdown is. I acknowledge, that the pod package is bloated, but on the other side, the pod format matured over years and is a perfect fit for man pages.

Just keep this in mind. I don't want to prevent you from any good actions.

apprehensions:
Technically speaking, scdoc is a very very similar format to markdown and could easily be converted back to markdown.

I believe for the time being it should be POD to retain Hugo compatibility up until the build system changes.

fwsmit:
I'll close this issue, as we are currently not planning to switch.

apprehensions:
Just as a reminder, sway uses scdoc, so i think dunst should too! (also reduces some meson complexity.), i can work on it after the meson PR has been merged.

PR #1226, opened 2023-11-02
initial meson move

### LINKED ISSUE #1224: Move to meson
Hi @fwsmit I just to hear your thoughts for moving Dunst to a better build system such as Meson.  The current makefile is using alot of GNUisms and isn't that portable and readable for that matter.

I would like to try to give this a shot.

apprehensions:
Currently, this has no support for the following features  that already existed within Makefile:

* ~~Valgrind~~ https://mesonbuild.com/Unit-tests.html#other-test-options
* ~~Test program~~  `ninja -C build test`
* ~~Documentation (#1225)~~
* Doxygen

Fixes #1224

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,8 +138,10 @@ doc: docs/dunst.1 docs/dunst.5 docs/dunstctl.1 docs/dunstify.1
 
 # Can't dedup this as we need to explicitly provide the name and title text to
 # pod2man :(
+docs/dunst.1.pod: docs/dunst.1.pod.in
+	${SED} "s|@sysconfdir@|${SYSCONFDIR}|" $< > $@
 docs/dunst.1: docs/dunst.1.pod
-	${SED} "s|##SYSCONFDIR##|${SYSCONFDIR}|" $< | ${POD2MAN} --name=dunst -c "Dunst Reference" --section=1 --release=${VERSION} > $@
+	${POD2MAN} --name=dunst -c "Dunst Reference" --section=1 --release=${VERSION} $< > $@
 docs/dunst.5: docs/dunst.5.pod
 	${POD2MAN} --name=dunst -c "Dunst Reference" --section=5 --release=${VERSION} $< > $@
 docs/dunstctl.1: docs/dunstctl.pod
@@ -153,11 +155,11 @@ doc-doxygen:
 .PHONY: service service-dbus service-systemd wayland-protocols
 service: service-dbus
 service-dbus:
-	@${SED} "s|##PREFIX##|$(PREFIX)|" org.knopwob.dunst.service.in > org.knopwob.dunst.service
+	@${SED} "s|@bindir@|$(BINDIR)|" org.knopwob.dunst.service.in > org.knopwob.dunst.service
 ifneq (0,${SYSTEMD})
 service: service-systemd
 service-systemd:
-	@${SED} "s|##PREFIX##|$(PREFIX)|" dunst.systemd.service.in > dunst.systemd.service
+	@${SED} "s|@bindir@|$(BINDIR)|" dunst.systemd.service.in > dunst.systemd.service
 endif
 
 ifneq (0,${WAYLAND})
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -1,6 +1,5 @@
-pod2man = find_program('pod2man', native: true)
-man1 = join_paths(get_option('mandir'), 'man1')
-man5 = join_paths(get_option('mandir'), 'man5')
+man1 = get_option('mandir') / 'man1'
+man5 = get_option('mandir') / 'man5'
 pod2man_version_arg = '--release=@0@'.format(meson.project_version())
 
 dunst1 = configure_file(
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -1,0 +1,61 @@
+pod2man = find_program('pod2man', native: true)
+man1 = join_paths(get_option('mandir'), 'man1')
+man5 = join_paths(get_option('mandir'), 'man5')
+pod2man_version_arg = '--release=@0@'.format(meson.project_version())
+
+dunst1 = configure_file(
+	input: 'dunst.1.pod.in',
+	output: 'dunst.1.pod',
+	configuration: conf_data,
+)
+
+custom_target(
+	'dunst1_pod2man',
+	input: dunst1,
+	output: 'dunst.1',
+	command: [
+		pod2man,
+		'--name=dunst',
+		'--center=Dunst Reference',
+		'--section=1',
+		pod2man_version_arg,
+		'@INPUT@',
+		'@OUTPUT@',
+	],
+	install: true,
+	install_dir: man1,
+)
+
+custom_target(
+	'dunst5_pod2man',
+	input: 'dunst.5.pod',
+	output: 'dunst.5',
+	command: [
+		pod2man,
+		'--name=dunst',
+		'--center=Dunst Reference',
+		'--section=5',
+		pod2man_version_arg,
+		'@INPUT@',
+		'@OUTPUT@',
+	],
+	install: true,
+	install_dir: man5,
+)
+
+custom_target(
+	'dunstctl_pod2man',
+	input: 'dunstctl.pod',
+	output: 'dunstctl.1',
+	command: [
+		pod2man,
+		'--name=dunst',
+		'--center=dunstctl Reference',
+		'--section=1',
+		pod2man_version_arg,
+		'@INPUT@',
+		'@OUTPUT@',
+	],
+	install: true,
+	install_dir: man1,
+)
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -3,58 +3,58 @@ man5 = get_option('mandir') / 'man5'
 pod2man_version_arg = '--release=@0@'.format(meson.project_version())
 
 dunst1 = configure_file(
-	input: 'dunst.1.pod.in',
-	output: 'dunst.1.pod',
-	configuration: conf_data,
+    input: 'dunst.1.pod.in',
+    output: 'dunst.1.pod',
+    configuration: conf_data,
 )
 
 custom_target(
-	'dunst1_pod2man',
-	input: dunst1,
-	output: 'dunst.1',
-	command: [
-		pod2man,
-		'--name=dunst',
-		'--center=Dunst Reference',
-		'--section=1',
-		pod2man_version_arg,
-		'@INPUT@',
-		'@OUTPUT@',
-	],
-	install: true,
-	install_dir: man1,
+    'dunst1_pod2man',
+    input: dunst1,
+    output: 'dunst.1',
+    command: [
+        pod2man,
+        '--name=dunst',
+        '--center=Dunst Reference',
+        '--section=1',
+        pod2man_version_arg,
+        '@INPUT@',
+        '@OUTPUT@',
+    ],
+    install: true,
+    install_dir: man1,
 )
 
 custom_target(
-	'dunst5_pod2man',
-	input: 'dunst.5.pod',
-	output: 'dunst.5',
-	command: [
-		pod2man,
-		'--name=dunst',
-		'--center=Dunst Reference',
-		'--section=5',
-		pod2man_version_arg,
-		'@INPUT@',
-		'@OUTPUT@',
-	],
-	install: true,
-	install_dir: man5,
+    'dunst5_pod2man',
+    input: 'dunst.5.pod',
+    output: 'dunst.5',
+    command: [
+        pod2man,
+        '--name=dunst',
+        '--center=Dunst Reference',
+        '--section=5',
+        pod2man_version_arg,
+        '@INPUT@',
+        '@OUTPUT@',
+    ],
+    install: true,
+    install_dir: man5,
 )
 
 custom_target(
-	'dunstctl_pod2man',
-	input: 'dunstctl.pod',
-	output: 'dunstctl.1',
-	command: [
-		pod2man,
-		'--name=dunst',
-		'--center=dunstctl Reference',
-		'--section=1',
-		pod2man_version_arg,
-		'@INPUT@',
-		'@OUTPUT@',
-	],
-	install: true,
-	install_dir: man1,
+    'dunstctl_pod2man',
+    input: 'dunstctl.pod',
+    output: 'dunstctl.1',
+    command: [
+        pod2man,
+        '--name=dunst',
+        '--center=dunstctl Reference',
+        '--section=1',
+        pod2man_version_arg,
+        '@INPUT@',
+        '@OUTPUT@',
+    ],
+    install: true,
+    install_dir: man1,
 )
--- a/meson.build
+++ b/meson.build
@@ -5,15 +5,11 @@ project(
 	license: 'MIT',
 	meson_version: '>=0.60.0',
 	default_options: [
-		'c_std=gnu99',
+		'c_std=gnu11',
 		'warning_level=1',
 		'b_ndebug=if-release',
 	],
 )
-
-add_project_arguments([
-	'-Wno-maybe-uninitialized',
-], language: 'c')
 
 if get_option('buildtype').startswith('debug')
 	add_project_arguments('-DDEBUG_BUILD', language: 'c')
--- a/meson.build
+++ b/meson.build
@@ -11,10 +11,6 @@ project(
     ],
 )
 
-if get_option('debug')
-    add_project_arguments('-DDEBUG_BUILD', language: 'c')
-endif
-
 cc = meson.get_compiler('c')
 
 cairo = dependency('cairo')
--- a/meson.build
+++ b/meson.build
@@ -42,15 +42,21 @@ wayland_protos = dependency('wayland-protocols', version: '>=1.32', required: ge
 wayland_cursor = dependency('wayland-cursor', required: get_option('wayland'))
 
 c_version_arg = '-DVERSION="@0@"'.format(meson.project_version())
+sysconfdir = get_option('sysconfdir') / 'xdg'
+
+add_project_arguments(
+	'-DSYSCONFDIR="@0@"'.format(get_option('prefix') / sysconfdir),
+	language: 'c'
+)
 
 subdir('src')
 
 install_data('dunstctl', install_dir: get_option('bindir'))
-install_data('dunstrc', install_dir: get_option('sysconfdir') / 'dunst')
+install_data('dunstrc', install_dir: sysconfdir / 'dunst')
 
 conf_data = configuration_data()
 conf_data.set('bindir', get_option('bindir'))
-conf_data.set('sysconfdir', get_option('sysconfdir'))
+conf_data.set('sysconfdir', sysconfdir)
 
 configure_file(
 	input: 'org.knopwob.dunst.service.in',
@@ -60,7 +66,7 @@ configure_file(
 )
 
 if systemd.found()
-	user_units_dir = systemd.get_pkgconfig_variable('systemduserunitdir')
+	user_units_dir = systemd.get_variable(pkgconfig: 'systemduserunitdir')
 	configure_file(
 		configuration: conf_data,
 		input: 'dunst.systemd.service.in',
--- a/meson.build
+++ b/meson.build
@@ -11,7 +11,7 @@ project(
 	],
 )
 
-if get_option('buildtype').startswith('debug')
+if get_option('debug')
 	add_project_arguments('-DDEBUG_BUILD', language: 'c')
 endif
 
@@ -35,17 +35,7 @@ wayland_client = dependency('wayland-client', required: get_option('wayland'))
 wayland_protos = dependency('wayland-protocols', version: '>=1.12', required: get_option('wayland'))
 wayland_cursor = dependency('wayland-cursor', required: get_option('wayland'))
 
-dunst_depends = [
-	cairo,
-	glib,
-	gio,
-	gdk_pixbuf,
-	pangocairo,
-	systemd,
-	libnotify,
-	realtime,
-	math,
-]
+dunst_depends = [ cairo, glib, gio, gdk_pixbuf, pangocairo, systemd, libnotify, realtime, math ]
 
 x11_support = x11.found() and xinerama.found() and xext.found() and xrandr.found() and xscrnsaver.found()
 wayland_support = wayland_client.found() and wayland_cursor.found() and wayland_protos.found()
@@ -59,13 +49,7 @@ if wayland_support and not x11_support
 endif
 
 if x11_support
-	dunst_depends += [
-		x11,
-		xinerama,
-		xext,
-		xrandr,
-		xscrnsaver,
-	]
+	dunst_depends += [ x11, xinerama, xext, xrandr, xscrnsaver ]
 	add_project_arguments('-DENABLE_X11', language: 'c')
 endif
 
--- a/meson.build
+++ b/meson.build
@@ -60,7 +60,7 @@ conf_data.set('sysconfdir', sysconfdir)
 
 configure_file(
 	input: 'org.knopwob.dunst.service.in',
-	output: '@BASENAME@',
+	output: 'dunst.service',
 	configuration: conf_data,
 	install_dir: get_option('datadir') / 'dbus-1/services',
 )
--- a/meson.build
+++ b/meson.build
@@ -38,7 +38,7 @@ libnotify = dependency('libnotify', required: get_option('dunstify'))
 realtime = cc.find_library('rt')
 math = cc.find_library('m')
 wayland_client = dependency('wayland-client', required: get_option('wayland'))
-wayland_protos = dependency('wayland-protocols', version: '>=1.32', required: get_option('wayland'))
+wayland_protos = dependency('wayland-protocols', version: '>=1.12', required: get_option('wayland'))
 wayland_cursor = dependency('wayland-cursor', required: get_option('wayland'))
 
 c_version_arg = '-DVERSION="@0@"'.format(meson.project_version())
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,87 @@
+project(
+	'dunst',
+	'c',
+	version: '1.9.2',
+	license: 'MIT',
+	meson_version: '>=0.60.0',
+	default_options: [
+		'c_std=gnu99',
+		'warning_level=1',
+		'buildtype=release',
+	],
+)
+
+add_project_arguments([
+	'-pedantic',
+	'-Wno-gnu-zero-variadic-macro-arguments',
+	'-Wno-overlength-strings',
+], language: 'c')
+
+if get_option('buildtype').startswith('debug')
+	add_project_arguments('-DDEBUG_BUILD', language: 'c')
+endif
+
+cc = meson.get_compiler('c')
+
+cairo = dependency('cairo')
+glib = dependency('glib-2.0')
+gio = dependency('gio-2.0')
+gdk_pixbuf = dependency('gdk-pixbuf-2.0')
+pangocairo = dependency('pangocairo')
+x11 = dependency('x11')
+xinerama = dependency('xinerama')
+xext = dependency('xext')
+xrandr = dependency('xrandr', version: '>=1.5')
+xscrnsaver = dependency('xscrnsaver')
+systemd = dependency('systemd', required: get_option('systemd'))
+libnotify = dependency('libnotify', required: get_option('dunstify'))
+realtime = cc.find_library('rt')
+math = cc.find_library('m')
+wayland_client = dependency('wayland-client', required: get_option('wayland'))
+wayland_protos = dependency('wayland-protocols', version: '>=1.32', required: get_option('wayland'))
+wayland_cursor = dependency('wayland-cursor', required: get_option('wayland'))
+
+c_version_arg = '-DVERSION="@0@"'.format(meson.project_version())
+
+subdir('src')
+
+install_data('dunstctl', install_dir: get_option('bindir'))
+install_data('dunstrc', install_dir: get_option('sysconfdir') / 'dunst')
+
+conf_data = configuration_data()
+conf_data.set('bindir', get_option('bindir'))
+conf_data.set('sysconfdir', get_option('sysconfdir'))
+
+configure_file(
+	input: 'org.knopwob.dunst.service.in',
+	output: '@BASENAME@',
+	configuration: conf_data,
+	install_dir: join_paths(get_option('datadir'), 'dbus-1/services'),
+)
+
+if systemd.found()
+	user_units_dir = systemd.get_pkgconfig_variable('systemduserunitdir')
+	configure_file(
+		configuration: conf_data,
+		input: 'dunst.systemd.service.in',
+		output: '@BASENAME@',
+		install_dir: user_units_dir,
+	)
+endif
+
+if libnotify.found()
+	executable(
+		'dunstify',
+		'dunstify.c',
+		dependencies: [ glib, libnotify, gdk_pixbuf ],
+		install: true,
+	)
+endif
+
+if get_option('test')
+	subdir('test')
+endif
+
+if get_option('docs').enabled()
+	subdir('docs')
+endif
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,7 @@ project(
 	default_options: [
 		'c_std=gnu99',
 		'warning_level=1',
-		'buildtype=release',
+		'b_ndebug=if-release',
 	],
 )
 
--- a/meson.build
+++ b/meson.build
@@ -78,9 +78,7 @@ if libnotify.found()
 	)
 endif
 
-if get_option('test')
-	subdir('test')
-endif
+subdir('test')
 
 pod2man = find_program('pod2man', native: true, required: get_option('docs'))
 if pod2man.found()
--- a/meson.build
+++ b/meson.build
@@ -82,6 +82,7 @@ if get_option('test')
 	subdir('test')
 endif
 
-if get_option('docs').enabled()
+pod2man = find_program('pod2man', native: true, required: get_option('docs'))
+if pod2man.found()
 	subdir('docs')
 endif
--- a/meson.build
+++ b/meson.build
@@ -1,18 +1,18 @@
 project(
-	'dunst',
-	'c',
-	version: '1.9.2',
-	license: 'MIT',
-	meson_version: '>=0.60.0',
-	default_options: [
-		'c_std=gnu11',
-		'warning_level=1',
-		'b_ndebug=if-release',
-	],
+    'dunst',
+    'c',
+    version: '1.9.2',
+    license: 'MIT',
+    meson_version: '>=0.60.0',
+    default_options: [
+        'c_std=gnu11',
+        'warning_level=1',
+        'b_ndebug=if-release',
+    ],
 )
 
 if get_option('debug')
-	add_project_arguments('-DDEBUG_BUILD', language: 'c')
+    add_project_arguments('-DDEBUG_BUILD', language: 'c')
 endif
 
 cc = meson.get_compiler('c')
@@ -35,35 +35,45 @@ wayland_client = dependency('wayland-client', required: get_option('wayland'))
 wayland_protos = dependency('wayland-protocols', version: '>=1.12', required: get_option('wayland'))
 wayland_cursor = dependency('wayland-cursor', required: get_option('wayland'))
 
-dunst_depends = [ cairo, glib, gio, gdk_pixbuf, pangocairo, systemd, libnotify, realtime, math ]
+dunst_depends = [
+    cairo,
+    glib,
+    gio,
+    gdk_pixbuf,
+    pangocairo,
+    systemd,
+    libnotify,
+    realtime,
+    math,
+]
 
 x11_support = x11.found() and xinerama.found() and xext.found() and xrandr.found() and xscrnsaver.found()
 wayland_support = wayland_client.found() and wayland_cursor.found() and wayland_protos.found()
 
 if not x11_support and not wayland_support
-	error('either wayland or x11 support is required')
+    error('either wayland or x11 support is required')
 endif
 
 if wayland_support and not x11_support
-	add_project_arguments('-DWAYLAND_ONLY', language: 'c')
+    add_project_arguments('-DWAYLAND_ONLY', language: 'c')
 endif
 
 if x11_support
-	dunst_depends += [ x11, xinerama, xext, xrandr, xscrnsaver ]
-	add_project_arguments('-DENABLE_X11', language: 'c')
+    dunst_depends += [x11, xinerama, xext, xrandr, xscrnsaver]
+    add_project_arguments('-DENABLE_X11', language: 'c')
 endif
 
 if wayland_support
-	dunst_depends += [ wayland_client, wayland_cursor ]
-	add_project_arguments('-DENABLE_WAYLAND', language: 'c')
+    dunst_depends += [wayland_client, wayland_cursor]
+    add_project_arguments('-DENABLE_WAYLAND', language: 'c')
 endif
 
 c_version_arg = '-DVERSION="@0@"'.format(meson.project_version())
 sysconfdir = get_option('sysconfdir') / 'xdg'
 
 add_project_arguments(
-	'-DSYSCONFDIR="@0@"'.format(get_option('prefix') / sysconfdir),
-	language: 'c'
+    '-DSYSCONFDIR="@0@"'.format(get_option('prefix') / sysconfdir),
+    language: 'c',
 )
 
 subdir('src')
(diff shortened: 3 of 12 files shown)


apprehensions:
I had underestimated how big this project is, which is why i initially thought this would be quite easy, as i had figured dunst is this simple notification daemon program.

fwsmit:
What is difficult about the test binary? It just needs to be compiled and run with valgrind. Meson probably has some way to run binaries

apprehensions:
Thankfully, Meson comes with coverage reports, so i hope we can drop the coverage reporting all together and only have valgrind tests and run the test program.

https://mesonbuild.com/howtox.html#producing-a-coverage-report

sound good to you?

fwsmit:
Yeah, seems good to me. Maybe @bebehei has a stronger opinion about this, since he set up most of our testing infrastructure.

bebehei:
Nice!

Meson was on my "fancy things I might implement" list.

Need to try it with some spare time at the evening.

apprehensions:
If you are a bit more experienced with Meson, I'd suggest you take control ™D

bebehei:
Sorry for giving false hints, but I do not have any experience with meson.

apprehensions:
I'll be dropping coverage tests for the time being.

apprehensions:
@fwsmit tests program is failing here as well.

apprehensions:
Notable changes:
* GitHub CI workflow is Ubuntu only, due to dependency constraints, and how it shouldn't realistically fail to run on musl or other distributions, the source is their problem typically.

.github/workflows/main.yml
```diff
@@ -7,107 +7,21 @@ on:
...
-  enforce-g_malloc:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v2
-      with:
-        # Clone the whole branch, we have to fetch tags later
-        fetch-depth: 0
-
-    - name: enforce-g_malloc
-      run: "! git grep -P '(?<!g_)malloc' ':!.valgrind.suppressions' ':!.github/workflows/main.yml'"
```
bebehei, line 20: This is super important, that devs use GLib here and not the libc defaults. It does not relate to cmake/meson, so no reason to remove.

apprehensions: In which codebase is this specific to? Dunst? I thought that because dunst never used any 'malloc' calls and was using 'g_malloc' it was already done without needing to change.

.github/workflows/main.yml
```diff
@@ -7,107 +7,21 @@ on:
...
-
-    - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v1
-      with:
-        token: ${{ secrets.CODECOV_TOKEN }}
-        flags: unittests
-        name: ${{ matrix.distro }}-${{ matrix.CC }}
-        fail_ci_if_error: true
-      if: "matrix.CC == 'gcc'"
+        sudo apt install -y ninja-build python3-pip gcc pkg-config libwayland-dev libx11-dev libxss-dev libxinerama-dev libxrandr-dev libnotify-dev libdbus-1-dev libwayland-dev libpango1.0-dev
```
bebehei, line 17: Sort those arguments and split them into multi line. Otherwise this will be a start of a huge mess.

Also you need to add `export DEBIAN_FRONTEND=noninteractive` to it. See some other docker reference: https://github.com/moby/moby/issues/4032#issuecomment-34597177 (tianon is the docker Debian library image maintainer)

_(Additionally I don't think you need `sudo`, but I'm not sure about this. Keep in if you need to add sudo, the call should be `sudo env DEBIAN_FRONTEND=noninteractive apt-get` to pass the variable.)._

```suggestion
       export DEBIAN_FRONTEND=noninteractive
       apt-get install -y \
            gcc \
            libdbus-1-dev \
            libnotify-dev \
            libpango1.0-dev \
            libwayland-dev \
            libwayland-dev \
            libx11-dev \
            libxinerama-dev \
            libxrandr-dev \
            libxss-dev \
            ninja-build \
            pkg-config \
            python3-pip
```

Taken deliberately from [Dockerfile best practices](https://docs.docker.com/develop/develop-images/guidelines/#sort-multi-line-arguments). The "general" section is a very good guideline for basically any code project.

src/meson.build
```diff
@@ -0,0 +1,53 @@
+dunst_depends = [
+	realtime, math,
+	glib, cairo, gio, gdk_pixbuf, pangocairo,
```
bebehei, line 3: Sort this as multiline, like mentioned in workflow.yml, too?

apprehensions:
Can `main.c` be moved into `src`?

.github/workflows/main.yml
bebehei: Right. It's for those who are unaware of it. So the CI will fail and automatically tell about their misuse.

bebehei:
> Can `main.c` be moved into `src`?

Probably yes.

apprehensions:
I am rather concerned over the fact that 'main.c' isnt simply just the `dunst_main` function from `dunst.c`, seems complicated for no good reason that i know of currently..

bynect:
> I am rather concerned over the fact that 'main.c' isnt simply just the `dunst_main` function from `dunst.c`, seems complicated for no good reason that i know of currently..

I was also wondering why that is.(some leftover?)

Probably main.c could be removed altogether by putting main in dunst.c?

apprehensions:
> Probably main.c could be removed altogether by putting main in dunst.c?

Other code in the test suites depend on code from dunst.c. I've tried to move dunst_main to main.c myself but unsure what to do about the `setup_done` variable.

bynect:
> > Probably main.c could be removed altogether by putting main in dunst.c?
> 
> Other code in the test suites depend on code from dunst.c. I've tried to move dunst_main to main.c myself but unsure what to do about the `setup_done` variable.

Well, since the main function is heavily tied to everything in dunst.c probably moving main there would be better.

apprehensions:
Moving main.c to dunst.c, while dunst.c is required by the test suite causes problems, as the test suite depends on dunst's code too.

bynect:
> Moving main.c to dunst.c, while dunst.c is required by the test suite causes problems, as the test suite depends on dunst's code too.

Some simple meta programming can fix that honestly.

```c
#ifndef TESTING

int main(int argc, char **argv) {
    return dunst_main(argc, argv);
}
#endif
```

apprehensions:
* Need to implement debug buildtype

apprehensions:
anything missing from the makefile so far?

bynect:
I was wondering, why remove outright the makefile when they can coexist afaik? Shouldn't the option to use meson be added alongside of makefiles?

Then, after some time (to get feedback from the users), removing the latter can be thought of in another pr.

apprehensions:
It doesn't make sense to keep both build systems, it will be messy.

bynect:
Yes, but it is still an additional dependency that is really not that required. I mean, it doesn't seem to add any value to what the makefiles can already do. So people would have to install and learn meson instead of just using make (installed by default everywhere) just for the sake of using meson. 

Well, that is my idea.  @fwsmit or @bebehei should decide a reasonable compromise.

Narrat:
There were some projects (like mpv, util-linux) which moved to meson and often it was a gracious time for switching.
First deprecating the old system with one release and with another removing it. But this transition period allowed to fix some eventual bugs/quirks with the new buildsystem and allowed package maintainers a smooth transition with a greater time window.
It can be frustrating if you want to just make a minor version bump and suddenly the whole thing fails. And often in those cases you're on a time budget and couldn't follow upstream changes :D And then you need to invest some extra time.
True, this is a rather uncomplicated project and switching to meson is often fairly easy.. But still. Just for package maintainers sake I would also favour a transition time.

apprehensions:
uh oh

apprehensions:
unfortunately, i'm not really that great at writing this new change in the CHANGELOG or README, sorry

Makefile
```diff
@@ -109,8 +109,10 @@ doc: docs/dunst.1 docs/dunst.5 docs/dunstctl.1
 
 # Can't dedup this as we need to explicitly provide the name and title text to
 # pod2man :(
+docs/dunst.1.pod: docs/dunst.1.pod.in
+	${SED} "s|@sysconfdir@|${SYSCONFDIR}|" $< > $@
```
fwsmit, line 113: Why do you change the sysconfdir in a different way from the other variables?

fwsmit:
The changes look good overall. I've left one comment. Does @bebehei have any last comments?

> unfortunately, i'm not really that great at writing this new change in the CHANGELOG or README, sorry

Okay, I can take a look at that.

Makefile
apprehensions: To retain the same configuration formatting as Meson.

.github/workflows/main.yml
```diff
@@ -11,103 +11,38 @@ jobs:
...
     - name: enforce-g_malloc
       run: "! git grep -P '(?<!g_)malloc' ':!.valgrind.suppressions' ':!.github/workflows/main.yml'"
 
   build:
-    strategy:
-      matrix:
-        CC:
-          - clang
-          - gcc
-        distro:
```
fwsmit, line 28: Does it not run on all distro's anymore when moving to meson?

fwsmit:
It seems the tests don't run. At least in the github tests it says the following

    0s
    Run ninja -C build test
    ninja: Entering directory `build'
    [0/1] Running all tests.
    No tests defined.

alebastr (commented):
Most of the CI images are waaaay too old to run this, but work with the `Makefile` and pre-compiled protocol sources. Like the Debian old-old-old-stable image with wayland-protocols 1.0.7.

This is certainly worth revisiting from the project maintainers, but for now meson workflow could be added in parallel to the old workflow file, with a more recent set of distros.

***
There's an interesting snippet in the `test/main.c`:
```c
const char *base;
// <...>
int main(int argc, char *argv[]) {
        char *prog = realpath(argv[0], NULL);
        if (!prog) {
                fprintf(stderr, "Cannot determine actual path of test executable: %s\n", strerror(errno));
                exit(1);
        }
        base = dirname(prog);
        // <...>
        // initialize settings
        char *config_path = g_strconcat(base, "/data/dunstrc.default", NULL);
        load_settings(config_path);
```

The test runner expects to find the required data files at `dirname("build-directory/test/test-runner")` and obviously fails.

A very quick hack to make the tests work with both buildsystems:
```diff
diff --git a/test/meson.build b/test/meson.build
index 7145916..988b8d5 100644
--- a/test/meson.build
+++ b/test/meson.build
@@ -35,4 +35,7 @@ test_prog = executable('test-runner',
 	install: false,
 )
 
-test('Run tests', test_prog)
+test('Run tests',
+    test_prog,
+    workdir : meson.project_source_root(),
+)
diff --git a/test/test.c b/test/test.c
index 0e63125..faaf99b 100644
--- a/test/test.c
+++ b/test/test.c
@@ -33,9 +33,13 @@ SUITE_EXTERN(suite_input);
 GREATEST_MAIN_DEFS();
 
 int main(int argc, char *argv[]) {
-        char *prog = realpath(argv[0], NULL);
+        const char *try_files[] = {"test/test.c", "test.c", argv[0], NULL};
+        char *prog = NULL;
+        for (const char **file = try_files; *file && !prog; ++file) {
+                prog = realpath(*file, NULL);
+        }
         if (!prog) {
-                fprintf(stderr, "Cannot determine actual path of test executable: %s\n", strerror(errno));
+                fprintf(stderr, "Cannot determine path of test data: %s\n", strerror(errno));
                 exit(1);
         }
         base = dirname(prog);
```

src/wayland/protocols/meson.build
```diff
@@ -0,0 +1,44 @@
...
+wayland_scanner_client = generator(
+	wayland_scanner,
+	output: '@BASENAME@-client-header.h',
+	arguments: ['client-header', '@INPUT@', '@OUTPUT@'],
+)
+
+client_protocols = [
+	wl_protocol_dir / 'stable/xdg-shell/xdg-shell.xml',
+	wl_protocol_dir / 'unstable/xdg-output/xdg-output-unstable-v1.xml',
+	wl_protocol_dir / 'staging/xdg-activation/xdg-activation-v1.xml',
```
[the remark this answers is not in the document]
apprehensions: > xdg-output
https://github.com/dunst-project/dunst/blob/master/Makefile#L138-L139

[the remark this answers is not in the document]
alebastr: It's neither used in the code nor linked to the binary. Just a leftover after https://github.com/dunst-project/dunst/commit/6a6cd6697fa220e0d6415f69c2ff1b628ac06413

meson.build
```diff
@@ -0,0 +1,86 @@
...
+
+configure_file(
+	input: 'org.knopwob.dunst.service.in',
+	output: '@BASENAME@',
+	configuration: conf_data,
+	install_dir: join_paths(get_option('datadir'), 'dbus-1/services'),
+)
+
+if systemd.found()
+	user_units_dir = systemd.get_pkgconfig_variable('systemduserunitdir')
```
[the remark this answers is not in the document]
apprehensions: I have noticed a few notification or systemd service affiliated projects don't actually install the systemd service themselves if the dependency is found, they keep it optional:
https://gitlab.archlinux.org/archlinux/packaging/packages/seatd/-/blob/main/PKGBUILD?ref_type=heads#L50
https://gitlab.archlinux.org/archlinux/packaging/packages/mako/-/blob/main/PKGBUILD?ref_type=heads#L47
Should dunst have the same behavior?
seatd used to install the systemd service but has dropped installing it since.

meson.build
```diff
@@ -0,0 +1,86 @@
...
+		dependencies: [ glib, libnotify, gdk_pixbuf ],
+		install: true,
+	)
+endif
+
+if get_option('test')
+	subdir('test')
+endif
+
+if get_option('docs').enabled()
```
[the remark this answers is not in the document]
apprehensions: very nice.

docs/meson.build
```diff
@@ -0,0 +1,61 @@
+pod2man = find_program('pod2man', native: true)
+man1 = join_paths(get_option('mandir'), 'man1')
+man5 = join_paths(get_option('mandir'), 'man5')
```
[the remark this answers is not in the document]
apprehensions: thanks

.github/workflows/main.yml
apprehensions: No, im just simplifying the CI.

apprehensions:
> It seems the tests don't run. At least in the github tests it says the following

I had the test program get built only on a build option, but i've dropped that and now only `ninja -C build test` can run the program - but the test program will *always* get built so, hope that's alright.

.github/workflows/main.yml
apprehensions: Additionally can be somewhat solved with #1232

Makefile
apprehensions: Oh, i see what you mean.

This is simply because `pod2man` can figure out the section on it's own, as for the title i'll proceed with adding it back.

apprehensions: Ah, its because this file in particular gets generated from a .in, the actual man page generation can be seen below for `docs/dunst.1`.

meson.build
```diff
@@ -0,0 +1,86 @@
...
+xinerama = dependency('xinerama')
+xext = dependency('xext')
+xrandr = dependency('xrandr', version: '>=1.5')
+xscrnsaver = dependency('xscrnsaver')
+systemd = dependency('systemd', required: get_option('systemd'))
+libnotify = dependency('libnotify', required: get_option('dunstify'))
+realtime = cc.find_library('rt')
+math = cc.find_library('m')
+wayland_client = dependency('wayland-client', required: get_option('wayland'))
+wayland_protos = dependency('wayland-protocols', version: '>=1.32', required: get_option('wayland'))
```
alebastr, line 41: You want 1.12 here to fix the CI build.
This project doesn't require anything newer than 1.12, and 1.32 is not available in the ubuntu-lts CI runner.

alebastr:
It appears that the protocol headers generated by meson were unused. The app was always including the existing prebuilt files from src/wayland/protocols, because that's the only possible match for `#include "protocols/*-client-header.h"`

I've tried to fix that with https://github.com/alebastr/dunst/commit/b9cd5499952a16cce7bf826ace4031fc59751305, which relies on `-I` passed either by meson or by Makefile.

.github/workflows/main.yml
fwsmit: It's actually useful to run the CI on multiple distro's because of their slight differences. This goes unnoticed in writing the code, but will get caught by the CI. For example, using a GLib feature that was only added in a newer version didn't compile in an old CI image. Now a check could be added for glib version and it compiles on all distro's again.

Makefile
fwsmit: Ah okay. That's fine

meson.build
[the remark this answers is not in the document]
fwsmit: For now it's best to replicate the makefile behavior as much as possible. So take a look at the makefile and see what it does there

meson.build
fwsmit: Yeah, just keep it on the oldest version that still works.

fwsmit (commented):
Thanks for you review @alebastr. I've added some comments as well.
The changes are looking good. Only try to touch parts outside the meson build files as little as possible. It's not that those changes are all bad per say, but most of those changes are unrelated to the move to meson. So they should be done in a separate PR to keep this one manageable.

meson.build
```diff
@@ -0,0 +1,86 @@
...
+wayland_cursor = dependency('wayland-cursor', required: get_option('wayland'))
+
+c_version_arg = '-DVERSION="@0@"'.format(meson.project_version())
+
+subdir('src')
+
+install_data('dunstctl', install_dir: get_option('bindir'))
+install_data('dunstrc', install_dir: get_option('sysconfdir') / 'dunst')
```
alebastr, line 49: This reverts move to `/etc/xdg/dunst/` and installs the config as `/etc/dunst/dunstrc`, right?

Maybe meson should just use `get_option('sysconfdir') / 'xdg'` both here and in conf_data unconditionally.

alebastr: Also needs something like
```meson
prefix = get_option('prefix')

add_project_arguments(
    '-DSYSCONFDIR="@0@"'.format(prefix / get_option('sysconfdir') / 'xdg'),
    language: 'c'
)
...
conf_data.set('bindir', prefix / get_option('bindir'))
conf_data.set('sysconfdir', prefix /get_option('sysconfdir'))
```
(get_option('...dir') won't include prefix when used in a string context)

apprehensions: Or just put it in `/usr/share/dunst/dunstrc`...

.github/workflows/main.yml
apprehensions: This is why a version requirement can be added into the Meson glib dependency.

meson.build
apprehensions: On a seperate note, why is `SYSCONFDIR` supposed to be defined? it seems to be used only for a Doxygen workaround.

alebastr: It's used in `src/settings.c:get_xdg_conf_basedirs` to pick the correct default configuration path. If you don't define `SYSCONFDIR`, dunst will look for configuration in `/usr/local/etc/xdg`, which is a bit... BSD'ish.

> It is never mentioned nor added in the Makefile.

See `DEFAULT_CPPFLAGS` in the [config.mk](https://github.com/dunst-project/dunst/blob/master/config.mk)

apprehensions:
> I've tried to fix that with https://github.com/alebastr/dunst/commit/b9cd5499952a16cce7bf826ace4031fc59751305, which relies on -I passed either by meson or by Makefile.

https://github.com/alebastr/dunst/blob/b9cd5499952a16cce7bf826ace4031fc59751305/Makefile#L139-L140

If the Makefile works, i dont believe any modification should be required to the code.

alebastr:
> https://github.com/alebastr/dunst/blob/b9cd5499952a16cce7bf826ace4031fc59751305/Makefile#L139-L140
> 
> If the Makefile works, i dont believe any modification should be required to the code.

Why did you add `src/wayland/protocols/meson.build` then? If you believe the result of this file should not be used, then it's not necessary.

apprehensions:
May i know how you know its unused?

alebastr:
> May i know how you know its unused?

Open `src/wayland/protocols/wlr-layer-shell-unstable-v1-client-header.h` and add `#error "This file should not be included`.

meson puts the generated code for protocols at `build-dir/src/wayland/protocols/libclient_protos.a.p/`, which cannot be resolved by include with `protocols/` prefix.

apprehensions:
> Why did you add src/wayland/protocols/meson.build then?

This is rather interesting. Removing the subdir requirement for this somehow makes it still run properly? why is this?

meson.build
fwsmit: Yeah, /usr/local/etc/xdg is a bit long for my tastes as well, but it's best to leave this the same for now. If you want to change that, it can be discussed in another PR

.github/workflows/main.yml
fwsmit: Still, testing on multiple distro's provides value in my opinion (for example alpine has musl libc, which has some differences from gnu libc). So please revert this change

fwsmit:
Hi @apprehensions, do you have some time to work on this? It'd be nice to finish this soon, since otherwise other PR's might interfere with this work

apprehensions:
I am mostly within constraint of how the source code of dunst is structured, and keeping it backwards compatible with Makefile. It would be also great if you can fix #1228 as it occurs in this PR.

fwsmit:
> I am mostly within constraint of how the source code of dunst is structured, and keeping it backwards compatible with Makefile.

What do you mean exactly? Is it constraining to you to keep backwards compatibility? If so we can discuss some things you might want to change. But I still want the makefile to exist alongside meson for at least one release. It's not neccesary to do things exactly the same way as in the makefile, but that makes it easier for me to review the PR.

> It would be also great if you can fix #1228 as it occurs in this PR.

As  I noted in that issue, I couldn't reproduce it, so that makes it a little bit hard to fix. An option for you to still run the test suite locally is to use the docker CI images (https://github.com/dunst-project/docker-images/). These can also be used locally to run the tests in docker. See the instructions in the readme for how to do that.

> To be honest, i don't believe it is my right to fix these inconsistencies regarding wayland protocol generation, so i suggest that get fixed in a seperate PR by @alebastr.

That's fine by me. The makefile doesn't address this either.

I would like it if you addressed all open discussions in this PR (and leaving a comment where you're not going to implement the suggestion). When that's done I can write some documentation and create issues for the things that still need to be done.

meson.build
apprehensions: `PREFIX/etc` is not a valid directory in the Linux FHS. I do not understand the need for it to be in `/usr/local/etc` or `/usr/etc`.

apprehensions:
dealing with CI is horrible

fwsmit:
> dealing with CI is horrible

Yeah, github CI is a mess. But if you're not able to do it in this PR, you could also revert the changes to the CI for now. It's okay if the CI keeps running on make for until it's removed.

meson.build
alebastr: Right, that's one of the confusing parts of meson. `get_option('prefix') / get_option('sysconfdir')` works quite counter-intuitive, but it's the right thing to use.

Consider:
| meson setup args | `get_option('prefix')` | `get_option('sysconfdir')` | `get_option('prefix') / get_option('sysconfdir')` | `install_dir: get_option('sysconfdir')`
| --- | --- | --- | --- | ---
| default | `/usr/local` | `etc` | `/usr/local/etc` | `/usr/local/etc`
| `--prefix=/usr/local` | `/usr/local` | `etc` | `/usr/local/etc` | `/usr/local/etc`
| `--prefix=/usr` | `/usr` | `/etc` | `/etc` | `/etc`
| `--sysconfdir=/etc` | `/usr/local` | `/etc` | `/etc` | `/etc`

With the current checkout, `man build/docs/dunst.1` will show you that the substituted value was `etc/xdg`, which is not what we want.

apprehensions:
It would be better to have Meson support in CI, as the original idea was to deprecate Makefile.

fwsmit:
> It would be better to have Meson support in CI, as the original idea was to deprecate Makefile.

Or course. But if that's too hard for this PR, it can also be done in a separate PR (since make is not yet removed).

fwsmit:
> It is not a good idea to keep Meson's CI checking in a seperate PR. The main goal atleast that was proposed by the maintainers, is to switch primarily to Meson and keep Make deprecated. To have the PR that switches the project to Meson but still use Make in it's CI is not a good idea.

Well, then the CI needs to be fixed before this is merged. Currently the CI doesn't work. The issue seems to be related to https://github.com/actions/checkout/issues/1590

apprehensions:
Simplicity is key, move to sourcehut and experience simpler and faster build frameworks.

apprehensions:
> Like I said in the past I think that this switch is kind of breaking

I meant 'deprecated' in the intention that it would be un-used but not explicitly removed.

> If it is too difficult to do everything in a single pr, the ci part could be moved to a different pr like @\fwsmit suggested. (especially if the meson code is done and just the ci is failing)

The meson code is technically complete. I believe that the CI should be functioning before this gets merged, to gurantee that the Meson move is functioning. The CI is failing due to the rather complicated test framework and the need for many distributions it seems.

In the meantime, i will just remove the CI modifications and let someone else handle it.

bynect:
> > Like I said in the past I think that this switch is kind of breaking
> 
> I meant 'deprecated' in the intention that it would be un-used but not explicitly removed.
> 
> > If it is too difficult to do everything in a single pr, the ci part could be moved to a different pr like @\fwsmit suggested. (especially if the meson code is done and just the ci is failing)
> 
> The meson code is technically complete. I believe that the CI should be functioning before this gets merged, to gurantee that the Meson move is functioning. The CI is failing due to the rather complicated test framework and the need for many distributions it seems.
> 
> In the meantime, i will just remove the CI modifications and let someone else handle it.

If you did a lot of the work on the ci you could simply create a new pr and send them there.

Also, sorry to bother you but could you take into account #1290 and make a way to compile only wayland without x11?
It should be fairly easy, you just need to removed the x deps and define a ENABLE_X11 as I did in the makefile.

Anyway the changes look good to me.

meson.build
```diff
@@ -0,0 +1,107 @@
+project(
+	'dunst',
+	'c',
+	version: '1.9.2',
+	license: 'MIT',
+	meson_version: '>=0.60.0',
+	default_options: [
+		'c_std=gnu99',
```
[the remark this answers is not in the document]
apprehensions: Lowering the C standard requirement as much as possible should be favored than to support something modern.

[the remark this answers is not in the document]
bynect: Yes I agree, in the first place I don't know why it was raised to 2011 (we don't even use the never features at the moment)

apprehensions:
> Are you referring to compiler warnings? 

Yes: https://kitten.pastes.sh/ninja.log

> Shouldn't it be possible with meson to create a test target to build separately the test program?

Good question... I don't think so unfortunately, only behind a configure-time option it would

bynect:
> > Are you referring to compiler warnings? 
> 
> Yes: https://kitten.pastes.sh/ninja.log i believe it comes from the `warning_level=1`.

Does this option change the flags sent to the compiler? Also at the end of the warnings it says a cflag was not recognized.

> > Shouldn't it be possible with meson to create a test target to build separately the test program?
> 
> No: https://github.com/mesonbuild/meson/issues/2518, the only workaround behind it is a configure time option.

I see. It's not too much of a problem but is there a way to silence the warnings?

apprehensions:
> Does this option change the flags sent to the compiler?

I'm not sure, but changing `warning_level` to `0` seems to solve it, but i'm not sure if that is a good idea.

bynect:
> > Does this option change the flags sent to the compiler?
> 
> I'm not sure, but changing `warning_level` to `0` seems to solve it, but i'm not sure if that is a good idea.

According to
https://mesonbuild.com/Builtin-options.html#details-for-warning_level

It will add additional `-Wall` options. But since you already add that to the whole project I don't think it is really that useful. Either you change warning_level or remove the -Wall

apprehensions:
> But since you already add that to the whole project 

Where?

bynect:
> > But since you already add that to the whole project 
> 
> Where?

I updated the comment. I thought I saw it but checking it back there wasn't

bynect:
I tried building with `meson build` and `ninja -C build` and everything worked well. I also tried with the x11 and wayland options. (nitpick question, is it possible to move dunst directly in the build dir?)

But when I tried to run the tests with `meson test -C build` the runner segfaults. When I read the log a lot of test failed, supposedly because the files in `test/data` where not found by the runner. 

[testlog.txt](https://github.com/dunst-project/dunst/files/14453516/testlog.txt)

Also, I didn't get any warnings about the test (with gcc) so I think it is somehow compiler specific.

apprehensions:
Congrats you found #1228 on your own, which I can replicate *without* Meson.

alebastr:
> But when I tried to run the tests with `meson test -C build` the runner segfaults. When I read the log a lot of test failed, supposedly because the files in `test/data` where not found by the runner.

```
WARNING: Cannot open '/home/nect/dev/dunst/build/test/data/dunstrc.default': 'No such file or directory'
```
I'm just going to point to https://github.com/dunst-project/dunst/pull/1226#pullrequestreview-1817345622, because it's most certainly a meson port issue.

bynect:
> Congrats you found #1228 on your own, which I can replicate _without_ Meson.

It actually is not the same thing, because I tried commenting the notification test suite (which is what causes the problem there) and it still is being problematic. Also a lot (and I mean it) of tests fail, so there is something else going on. Obviously I tried make and it works for the tests. 



> > But when I tried to run the tests with `meson test -C build` the runner segfaults. When I read the log a lot of test failed, supposedly because the files in `test/data` where not found by the runner.
> 
> ```
> WARNING: Cannot open '/home/nect/dev/dunst/build/test/data/dunstrc.default': 'No such file or directory'
> ```
> 
> I'm just going to point to [#1226 (review)](https://github.com/dunst-project/dunst/pull/1226#pullrequestreview-1817345622), because it's most certainly a meson port issue.

Thanks for pointing that out. I thought that a way to fix this problem was added already.

meson.build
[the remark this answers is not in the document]
zappolowski: It's pretty easy: we're building with `-Werror` and `_Static_assert` (coming from glib) requires `-std=gnu11`.

meson.build
```diff
@@ -0,0 +1,107 @@
...
+glib = dependency('glib-2.0')
+gio = dependency('gio-2.0')
+gdk_pixbuf = dependency('gdk-pixbuf-2.0')
+pangocairo = dependency('pangocairo')
+x11 = dependency('x11', required: get_option('x11'))
+xinerama = dependency('xinerama', required: get_option('x11'))
+xext = dependency('xext', required: get_option('x11'))
+xrandr = dependency('xrandr', required: get_option('x11'), version: '>=1.5')
+xscrnsaver = dependency('xscrnsaver', required: get_option('x11'))
+systemd = dependency('systemd', required: get_option('systemd'))
```
zappolowski, line 36: `systemd` is not required to build the project.

meson.build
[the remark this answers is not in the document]
bynect: Where are we using static assert though?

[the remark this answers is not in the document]
zappolowski: > ... (coming from glib) ...

We're not, but we use a dependency which does.

meson.build
apprehensions: Yes, it is not required, hence: it requires the `systemd` option to be on, in case the distributor wants to intsall the units directory.

```meson
if systemd.found()
	user_units_dir = systemd.get_variable(pkgconfig: 'systemduserunitdir')
```

zappolowski: Why can't installation of the files depend on `get_option('systemd')` then?

apprehensions: It does. Again: `dependency` is used so it can retrieve the unit directory.

meson.build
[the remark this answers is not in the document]
bynect: using gnu99 doesn't give any error. If they use static_assert they use it internally/optionally 

> > ... (coming from glib) ...
> 
> We're not, but we use a dependency which does.

using gnu99 doesn't give any error. If they use static_assert they use it internally/optionally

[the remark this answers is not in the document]
zappolowski: You didn't try `clang`, did you?

```shell
$ export CC=clang CFLAGS=-Werror # setup env as it's used on CI
$ meson setup build
$ meson compile -C build
```
the last command fails with messages like
```
../test/icon.c:106:9: error: '_Static_assert' is a C11 extension [-Werror,-Wc11-extensions]
  106 |         g_clear_pointer(&icon_path, g_free);
      |         ^
/usr/include/glib-2.0/glib/gmem.h:138:5: note: expanded from macro 'g_clear_pointer'
  138 |     G_STATIC_ASSERT (sizeof *(pp) == sizeof (gpointer)); \
      |     ^
/usr/include/glib-2.0/glib/gmacros.h:870:31: note: expanded from macro 'G_STATIC_ASSERT'
  870 | #define G_STATIC_ASSERT(expr) _Static_assert (expr, "Expression evaluates to false")
      |                               ^
```

(this was run using clang 17.0.6 but should also fail on the 16 series).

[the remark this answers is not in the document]
bynect: https://gitlab.gnome.org/GNOME/glib/-/issues/2495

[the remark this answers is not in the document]
bynect: Tldr: pedantic can't be used with glib

apprehensions:
...What is that status on this?

bynect:
if i remember correctly the only problem was with the test runner?

apprehensions:
Why is there an installation section in the wiki? shouldn't it be in the README?

bynect:
> Why is there an installation section in the wiki? shouldn't it be in the README?

I guess it should be in both. the wiki is still documentation

hholst80:
> I had underestimated how big this project is,

That is how any project get started.  If we knew up front how much work it would be, nobody would ever start anything. ;-)

Looking forward to the final result, I have packaged up 150+ source packages and the stinkers are those who does not follow a strict autoconf / meson or cmake build.

The best most pleasant packages to work with are those build on Meson.

bynect:
can we finish this? it was mostly done if I remember correctly

bynect:
I will try it later. Then if everything works I'll add it to the next release alongside makefile. After a while the makefile can be removed after proper notice is given to the package maintainers

meson.build
```diff
@@ -0,0 +1,105 @@
...
+	default_options: [
+		'c_std=gnu11',
+		'warning_level=1',
+		'b_ndebug=if-release',
+	],
+)
+
+if get_option('buildtype').startswith('debug')
+	add_project_arguments('-DDEBUG_BUILD', language: 'c')
+endif
```
eli-schwartz, line 16:
```meson
if get_option('debug')
```

meson.build
eli-schwartz: You can try falling back to /usr/lib/systemd/user

It's relatively common for projects to have a `-D systemd-userdir=''` option in meson_options.txt that if empty (the default) attempts to use `dependency('systemd').get_variable` but can be overridden; this is useful for systems such as Gentoo, where the files should always be installed even if the init system is openrc, to allow the resulting built package to be efficiently cached across users of diverse init systems.

src/meson.build
```diff
@@ -0,0 +1,74 @@
...
+	'option_parser.c', 
+	'output.c', 
+	'queues.c', 
+	'rules.c', 
+	'settings.c', 
+	'utils.c', 
+)
+
+if x11_support
+	add_global_arguments('-DENABLE_X11', language: 'c')
```
eli-schwartz, line 34: Why add_global here? Previously you used add_project -- which makes more sense anyway, in most cases.

eli-schwartz: I would also suggest moving this to the top-level file anyway... you already have generic option checking code, why do this inside a subdir during target creation?

src/meson.build
```diff
@@ -0,0 +1,74 @@
...
+		xscrnsaver,
+	]
+	dunst_src_files += files(
+		'x11/screen.c', 
+		'x11/x.c', 
+	)
+endif
+
+if wayland_support
+	add_global_arguments('-DENABLE_WAYLAND', language: 'c')
```
eli-schwartz, line 50: Same...

test/meson.build
```diff
@@ -0,0 +1,41 @@
...
+	if fs.name(dunst_src_file) not in test_src_files
+		test_src_files += dunst_src_file
+	endif
+endforeach
+
+test_prog = executable('test-runner',
+	test_src_files,
+	dependencies: dunst_depends,
+	c_args: c_version_arg,
+	install: false,
```
eli-schwartz, line 35: `build_by_default: false` to tell meson that this target is only needed when running `meson test` or `ninja test`, and isn't needed by `ninja all` or `ninja install`

eli-schwartz (changes requested):

test/meson.build
apprehensions: Correct. why should the test runner be built if the user won't execute the test suite?

bynect: build_by_default is true by default though

bynect:
@apprehensions ltgm but could you apply the suggestion of elischwartz ?

test/meson.build
apprehensions: woops

bynect:
the build of dunst seems to work well.

I have a suggestion for the test

```diff
diff --git a/test/meson.build b/test/meson.build
index b74a29f..491ccc9 100644
--- a/test/meson.build
+++ b/test/meson.build
@@ -40,5 +40,6 @@ test_prog = executable(
 test(
     'Run tests',
     test_prog,
+    args: ['-v'],
     env: environment({'TESTDIR': meson.current_source_dir()}),
)
```

also, maybe I am being dumb here, but I can't replicate the output of the test I get in makefile/normally.
in the log meson puts all of the logging which normally is hidden. I don't know which line is causing it though

eli-schwartz:
Meson by default wraps stdout/stderr and summarizes results, unless the tests fail (ninja test defaults to --print-errorlogs). You can get the logs directly if you pass -v / --verbose to `meson test`.

So ensuring that the test output isn't "quiet by default" internally, seems correct to me.

Makefile
```diff
@@ -138,8 +138,10 @@ doc: docs/dunst.1 docs/dunst.5 docs/dunstctl.1 docs/dunstify.1
...
 docs/dunst.1: docs/dunst.1.pod
-	${SED} "s|##SYSCONFDIR##|${SYSCONFDIR}|" $< | ${POD2MAN} --name=dunst -c "Dunst Reference" --section=1 --release=${VERSION} > $@
```
bynect, line 142: shouldn't sysconfdir be replaced like before?

meson.build
```diff
@@ -0,0 +1,130 @@
+project(
+    'dunst',
+    'c',
+    version: '1.9.2',
+    license: 'MIT',
+    meson_version: '>=0.60.0',
+    default_options: [
+        'c_std=gnu11',
```
bynect, line 8: gnu99 seems to be enough.
also the only divergence from c99 that I can find is the use of variadic macros in our codebase with no extra args

Makefile
bynect: my bad it was moved above

bynect:
> > but I can't replicate the output of the test I get in makefile/normally.
> 
> I believe this is caused by the 'greenest' awk, which is not really necessary IMO. If the tests were fully transitioned to Meson, then it could be pretty without the program.

greenest just colours some keywords so its not the culprit



> Meson by default wraps stdout/stderr and summarizes results, unless the tests fail (ninja test defaults to --print-errorlogs). You can get the logs directly if you pass -v / --verbose to `meson test`.
> 
> So ensuring that the test output isn't "quiet by default" internally, seems correct to me.

thanks for pointing out the verbose flag. however I am quite confused by the fact that while running the 
test normally the numerous debug prints are turned off (which really helps focusing on the test results)
but the test compiled with meson has them turned on.

there is an envvar called DUNST_TEST_LOG. when this is not set the log should be quiet
(meson doesn't set it either so I don't understand this beheaviour)

bynect:
For now meson is the secondary build system, however I am positive we can move totally to meson in the future.
I would like to leave a couple of release before removing the makefile.
also to move to meson we have to make the ci work  like the makefile.

bynect:
anyway, thanks for your work ❤️

bynect:
It now occurs to me that the prefix is not handled correctly, as pointed out by Narrat in #1462.

When the prefix is set to something the bindir is not actually used as prefix+bin.
I supposed that meson automatically managed the appending but I guess not.

Is there an idiomatic way to solve this? @apprehensions @eli-schwartz

eli-schwartz:
> When the prefix is set to something the bindir is not actually used as prefix+bin.
> I supposed that meson automatically managed the appending but I guess not.

There are valid use cases for knowing both a relative and an absolute bindir.

Given:
```
--prefix=/usr --bindir=/usr/bin

--prefix=/usr --bindir=bin
```

the value of `get_option('bindir')` is "bin" and the absolute path can be gotten with

```meson
abs_bindir = get_option('prefix') / get_option('bindir')
```

For something like this:

```meson
install_data('dunstctl', install_dir: get_option('bindir'))
```
Meson knows to install dunstctl in the directory sigified as bindir, and to find that relative to the prefix. The ambiguity comes when you want to render the string representation of that directory inside a configuration file, such as a systemd service.

meson.build
```diff
@@ -0,0 +1,126 @@
...
+    language: 'c',
+)
+
+subdir('src')
+
+install_data('dunstctl', install_dir: get_option('bindir'))
+install_data('dunstrc', install_dir: sysconfdir / 'dunst')
+
+conf_data = configuration_data()
+conf_data.set('bindir', get_option('bindir'))
```
eli-schwartz, line 81: ... but forgotten here.

PR #1213, opened 2023-10-22
Improve draw_rounded_rect and fix progress bar

bynect:
I started this pr to fix the problem with the progress bar described in #1178.

Before             |  After
:-------------------------:|:-------------------------:
image | image
image  | image
image

However I got carried away and also updated `draw_rounded_rectangle` to be more generic on which corners to round.
I plan on adding a configuration option to select which corners should be rounded.

These are some of the possible corner configurations:

1 | 2
:-------------------------:|:-------------------------:
image | image
image | image

PR #1227, opened 2023-11-07
Corners patch part 1

bynect:
I divided the patch in #1213 to simplify things a bit. This first part only deals with the rendering of the corners. The second part (which I will add soon) adds support for the customization option to change corner profile. Anyway, refer to #1213 for any information on the changes.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -520,16 +520,19 @@ void draw_rounded_rect(cairo_t *c, float x, float y, int width, int height, int 
         y *= scale;
         corner_radius = round(corner_radius * scale);
 
+        // Nothing valid to draw
         if (width <= 0 || height <= 0 || scale <= 0)
                 return;
 
         const double degrees = M_PI / 180.0;
 
+        // This seems to be needed for a problem that occurs mostly when using cairo_stroke
+        // and not cairo_fill. Since the only case where we have partial angles is the progress
+        // bar and there we use fill, maybe this can be removed completely?
         enum corner_pos skip = C_NONE;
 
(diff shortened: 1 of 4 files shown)


bynect:
Should I close the other?

fwsmit:
> Should I close the other?

No that's not neccesary

fwsmit:
Sorry that I've taken so long to review this. This PR still contains more than just the fix for the progress bar rendering, but I'll try to roll with it. Could you explain how the bug happened in the old progress bar implementation? Good testing is nice, but it's also good to know what went wrong.

fwsmit:
The code does look good, but I'll have to take a closer later.

bynect:
> Sorry that I've taken so long to review this. This PR still contains more than just the fix for the progress bar rendering, but I'll try to roll with it. Could you explain how the bug happened in the old progress bar implementation? Good testing is nice, but it's also good to know what went wrong.

To bw fair, the only thing that is not about corners is the bar adjustment, which is minimal and I already commented in the original pr. (It just deals with either a pango antialiasing/number rounding issue that makes the lines slightly skewed)

When you merge this I can make the part 2 of the pr for costumizing corners

bynect:
did you manage to see it?

> The code does look good, but I'll have to take a closer later.

fwsmit:
Thanks for the poke. The code looks good, I haven't found any issues with it. Thanks for your patience and working on this fix. I will go ahead and merge this. I look forward to the V2 patch!

bynect:
Thanks. I'll prepare the patch as soon as I have a little spare time since the code is already there. Sometime next week at most

PR #1231, opened 2023-11-09
Update greatest to v1.5

zappolowski:
The new version disallows nesting of tests. Therefore, `test_markup_strip_a` and `test_markup_strip_img` were converted to suites. This also means, that the number of tests is formally reduced by 2.


fwsmit:
Looks good. Always great to update libraries.

PR #1233, opened 2023-11-10
Use brackets consistently in dunstctl usage

zappolowski:
As noted in [this comment](https://github.com/dunst-project/dunst/pull/1212#discussion_r1389414771) usage of brackets is a bit inconsistent in the usage output of dunstctl.

Just optional argument should be put in `[]`. Required ones are without them.


fwsmit:
Great, thanks!

PR #1238, opened 2023-11-29
Fix a typo and add periods to end all hints consistently.

maxz:
I came across the typo while reading the man page and then the first four entries ending without periods and all others ending in a period seemed strange, so I changed that as well.


fwsmit:
Thanks!

PR #1249, opened 2024-01-02
fix progress bar rounding at 0% by disabling rounding in these circumstances

Fxzzi:
Hello dunst-project!!

A very small PR here;

An example of the original issue: 
image
You can see here with rounding enabled on the progress bar, at 0% the rounding does not render correctly (it technically renders correctly, but not in the way a user would expect)

This PR simply fixes this by making sure the progress width is not 0 before drawing the rounded sides.

Here is how it looks after:
image

rounding at all other percentage levels is identical to before as expected:
image

Let me know if any changes are required, I'll be happy to help :)


bynect:
this is a duplicate of #1227, which in turn is a part of #1213

fwsmit:
Hey, thanks for this PR. As noted above this is already fixed by #1227.

Issue #1214, opened 2023-10-22
idle_threshold setting doesn't work on Wayland


dismint:
### Issue description
dunst is unable to process the `idle_threshold` setting on Wayland. It simply returns a warning in the log that it is unable to set the variable. When using `force_wayland=true`, dunst reports it cannot process `idle_threshold`, which is to be expected considering the man page reads:
```
Don't timeout notifications if user is idle longer than this time. See TIME FORMAT for valid times.
Set to 0 to disable.

A client can mark a notification as transient to bypass this setting and timeout anyway. Use a rule with 'set_transient = no' to disable this behavior.

Note: this doesn't work on xwayland.
```
When run on X11, it works as expected.

Running on Wayland produces the following results from `dunst -verbosity debug`:

```
INFO: Using Wayland output
DEBUG: [   handle_global:0428] Found toplevel manager 28
INFO: New output found - id 0
DEBUG: [   handle_global:0422] Binding to output 50
DEBUG: Using cross-namespace EXTERNAL authentication (this will deadlock if server is GDBus < 2.73.3)
DEBUG: [         wake_up:0063] Ignoring wake up
WARNING: couldn't set idle timeout
DEBUG: [             run:0077] RUN, reason 0
INFO: Idle status queried: 0
INFO: Wayland: Hiding window
```
### Installation info

- Version: `1.9.2`
- Install type: arch `extra/dunst`
- Window manager / Desktop environment: `Hyprland`
- Distro: `EndeavourOS`

<details>
<summary>dunstrc</summary>
```ini
[global]
	# overall styling
	monitor = 0
	follow = mouse
	enable_posix_regex = true
	notification_limit = 2

	# styling of each notification
	width = 300
	origin = top-center
	offset = 0x25
	frame_width = 4
	corner_radius = 4

	gap_size = 10

	idle_threshold = 120ms

	sticky_history = true

	mouse_left_click = do_action, close_current
```
</details>

Narrat:
Did you try the setting without the SI unit?

dismint:
Yes I've tried it with `ms`, without `ms`, in quotes with `ms`, in quotes without `ms`. All lead to this same result.

Narrat:
Okay, because I'm not sure if dunst capable of doing this unit conversion.
But indeed. I can see it in the debug output too. Was previously using the default and hadn't set the config option. But the default seems to be broken too, as notifications timed out which shouldn't have. And i know it was working in the past :D
[...]

PR #1250, opened 2024-01-10
Wayland: multiseat, new protocols and other random fixes

### LINKED ISSUE #1132: Wayland: Monitor number makes it difficult to consistently choose the correct display
### Issue description

dunst only allows choosing a monitor based on its "number"
This number can change on boot and also with display hotplugging
It would be ideal to be able to define an output by its name

### Installation info

- Window manager / Desktop environment: `sway`
- Nvidia proprietary drivers

### Details

dunst [seems to](https://github.com/dunst-project/dunst/blob/58206fe50ff3ef97bc12f49caf830e82e10c08a6/src/wayland/wl.c#L376) get a list of wayland output numbers on start, but this list appears to be dynamic, adding or removing a new display messes with the order, and changes the output dunst displays on silently. (I guess because it follows the wayland client example implementation)

The ideal solution imo would be to be able to supply an output name _a la_ sway, ie: `"Ancor Communications Inc ASUS PB278 C9LMTF278905"` or the short output port name `HDMI-A-1` (these can be found with `swaymsg -t get_outputs`).
They should be available in `wl_output::name` according to the docs.

Ideally both formats would be supported (maybe it just works with common libraries?) because each potentially has advantages for different users (desktop vs laptop vs whatever, depending on how stuff is handled).

I think this solution should probably replace the current output number behavior for wayland, though I dont know how other compositors and desktop environments manage this, and I dont know if that would be an annoyance for people that want to use X and Wayland concurrently or want to migrate between them seamlessly.

Adding a second option like `monitor_name` doesn't seem to be too bad for the end user at least.

Implementing this myself is a bit over my head unfortunately.

alebastr:
A bunch of fixes for Wayland backend:
 - Complete multiseat support
 - `ext-idle-notify-v1`, required by wlroots >= 0.17
 - `cursor-shape-v1` for compositor-rendered cursors to reduce headache with themes and scaling
 - Support for `global.monitor = <name>` (eDP-1, etc...)

Tested on Fedora 39 and FreeBSD 14. Haven't tried with older systems, but everything version-dependent should be properly guarded.
Cursor-shape and ext-idle-notify could be tested with a compositor already ported to the new wlroots (Sway master or v1.9 branches).

I'm 100% unsure about the X11 impl for monitor names. It's only been tested with XWayland, and frankly I'm not familiar with the Xlib.

Fixes: #1132
Likely also addresses #1214


Narrat:
Tested the branch and at least the warning mentioned in #1214 seems to be gone.
But didn't had the impression it was working. But I need to test that further.

alebastr:
> Tested the branch and at least the warning mentioned in #1214 seems to be gone. But didn't had the impression it was working. But I need to test that further.

Can you set a very short (like 10s short) idle_threshold, start dunst as `WAYLAND_DEBUG=client dunst -verbosity debug` and collect the output?

I've been testing various corner-cases with Sway 1.9 snapshot and everything has been working so far. labwc shouldn't have any significant differences in behavior with Sway.

Narrat:
All good. Can confirm this PR does fix #1214 
I hadn't noticed that I removed the setting from the config (and the default seems to be to not set a time)? Typical user error :D
Thank you for the work on the wayland related side of things

fwsmit:
Thanks for this much needed wayland work! With a quick look, the X11 impl seems to be fine as well, but I'm not too familiar with the xlib either. If someone is able to test it on X11, that would be great.

Makefile
```diff
@@ -38,6 +38,7 @@ ifeq (0,${WAYLAND})
 SRC := $(sort $(shell ${FIND} src/ -not \( -path src/wayland -prune \) -name '*.c'))
 else
 # with Wayland support
+CFLAGS += -DHAVE_WL_CURSOR_SHAPE -DHAVE_WL_EXT_IDLE_NOTIFY
```
fwsmit, line 41: Could you explain why you added separate flags for these protocols?

alebastr: A bit of a pedantry, a bit of a preparation for the meson PR.

It is highly recommended to regenerate the protocol sources from xml at the build time, as the `wayland-scanner` output changes and may start relying on a new APIs (like the [`wl_proxy_marshal_flags`](https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/162) introduction in 1.19.91).

Doing that the way this repo is organized currently, with the pre-built protocol sources, will result in breaking compatibility with libwayland 1.19 (and it was a pain to get an old enough `wayland-scanner` to build the files for this PR).
On the other hand, building all the protocols every time would require w-p >= 1.32, which is a bit too hard on the users. The middle ground is to make support for the newer protocols conditional on the w-p version available during build.

I'm honestly giving up on solving this via Makefile (nobody runs `make wayland-protocols` at their system/during packaging, right?) and laying the groundwork for a proper conditionals in meson when #1226 lands. E.g.
```meson
if wayland_protos.version() >= '1.27'
     client_protocols += wl_protocol_dir / 'staging/ext-idle-notify/ext-idle-notify-v1.xml'
     add_project_arguments('-DHAVE_WL_EXT_IDLE_NOTIFY', language: 'c')
endif

if wayland_protos.version() >= '1.32'
    client_protocols += [
        wl_protocol_dir / 'staging/cursor-shape/cursor-shape-v1.xml',
        wl_protocol_dir / 'unstable/tablet/tablet-unstable-v2.xml',
    ]
    add_project_arguments('-DHAVE_WL_CURSOR_SHAPE', language: 'c')
endif
```
And sorry if it looks weird out of context.

fwsmit: Okay, makes sense. Thanks for the explanation.

src/wayland/wl.c
```diff
@@ -122,34 +140,52 @@ static void output_handle_scale(void *data, struct wl_output *wl_output,
...
+#ifdef WL_OUTPUT_NAME_SINCE_VERSION
+        max_version = WL_OUTPUT_NAME_SINCE_VERSION;
+#endif
         bool recreate_surface = false;
         static int number = 0;
         LOG_I("New output found - id %i", number);
         output->global_name = global_name;
-        output->wl_output = wl_output;
+        output->wl_output = wl_registry_bind(registry, global_name, &wl_output_interface,
+                        CLAMP(version, 3, max_version));
```
fwsmit, line 179: I think this can be done in a less confusing way. You're trying to select the version WL_OUTPUT_NAME_SINCE_VERSION if that's available and otherwise version 3, right? Why need the clamp?

src/wayland/wl.c
```diff
@@ -236,35 +292,94 @@ static const struct wl_touch_listener touch_listener = {
...
-                ctx.touch.wl_touch = wl_seat_get_touch(wl_seat);
-                wl_touch_add_listener(ctx.touch.wl_touch,
-                        &touch_listener, ctx.seat);
+                if (seat->touch.wl_touch == NULL) {
+                        seat->touch.wl_touch = wl_seat_get_touch(wl_seat);
+                        wl_touch_add_listener(seat->touch.wl_touch,
+                                &touch_listener, seat);
+                }
+        } else if (seat->touch.wl_touch != NULL) {
+                destroy_seat_touch(seat);
```
fwsmit, line 367: If I read this correctly, the touch input is destroyed when a second input is added. It seems to me that if a second input is added, the old one should stay. But let me know if it works differently.

src/wayland/protocols.c
```diff
@@ -0,0 +1,13 @@
...
+#include "protocols/idle.h"
+
+#ifdef HAVE_WL_CURSOR_SHAPE
+#include "protocols/cursor-shape-v1.h"
+#include "protocols/tablet-unstable-v2.h"
+#endif
+
+#ifdef HAVE_WL_EXT_IDLE_NOTIFY
+#include "protocols/ext-idle-notify-v1.h"
+#endif
```
fwsmit, line 13: Is this file used anywhere? It might be used in wl.c instead of inporting the protocols separately.

fwsmit:
I've taken a look at all the code. I have some questions and minor things, but over all it looks good.

src/wayland/wl.c
alebastr: The version announced by the compositor can be lower than `WL_OUTPUT_NAME_SINCE_VERSION`. In this case it would not be correct to bind to it (IIRC, it's even a protocol error).

So we have a minimum supported version (3), maximum supported (`WL_OUTPUT_NAME_SINCE_VERSION` if compiled with the newer headers), and we should be able to accept anything within this range.

src/wayland/wl.c
alebastr: The touch input is destroyed when the touch capability disappears from the seat. If the capability does not change, the existing wl_touch will be kept. Same goes for pointer.

A seat can have only one touch (pointer, keyboard...) object and only if the corresponding capability was announced via capabilities event.

src/wayland/wl.c
fwsmit: I believe you are able to bind to a lower version than the compositor announces. Otherwise compositors would not be able to talk with older apps.
So my suggestion is to either bind to version 3 or the other required version. Not anything in between. I think that would make the code a little bit clearer

alebastr: So... `version = version >= max_version ? max_version : 3`? I don't believe that to be a significant improvement over `CLAMP(version, 3, max_version)`.

Besides, that works in the current case (where WL_OUTPUT_NAME_SINCE_VERSION is 4), but with other protocols we may benefit from additional events/requests introduced in the versions between min and max. Binding strictly to min or max would prevent that.
That's why, as a general pattern, clamping is better.

fwsmit: >  That's why, as a general pattern, clamping is better.
Okay, I'll accept that argument

src/wayland/wl.c
fwsmit: Ah okay, makes sense

src/wayland/wl.c
```diff
@@ -236,35 +299,94 @@ static const struct wl_touch_listener touch_listener = {
...
+                }
+        } else if (seat->pointer.wl_pointer != NULL) {
+                destroy_seat_pointer(seat);
         }
+
         if (capabilities & WL_SEAT_CAPABILITY_TOUCH) {
-                ctx.touch.wl_touch = wl_seat_get_touch(wl_seat);
-                wl_touch_add_listener(ctx.touch.wl_touch,
-                        &touch_listener, ctx.seat);
+                if (seat->touch.wl_touch == NULL) {
```
fwsmit, line 368: In what situation would this not be NULL?

fwsmit:
The wl.c file is getting a bit big and could use some refactoring to separate files. I'm not saying it should happen in this PR, but this PR does make it a little bit bigger as well.

src/wayland/wl.c
alebastr: Any other capability on the seat has changed. E.g. a keyboard was unplugged, we got a new capability event with `capabilities == pointer | touch` and we already have existing wl_touch/wl_pointer objects.

fwsmit: Okay, then it's good :+1:

fwsmit:
Thanks for all the fixes and wayland features! Now dunst is up to date with the protocols again

fwsmit:
I'll go and merge this PR. If you want to work on refactoring the wayaland code a bit, that would be cool. But don't feel obliged :)

PR #1263, opened 2024-01-21
Feature: Install completions by default

### LINKED ISSUE #1262: Feature request: Add the shell completions to make install
### Issue description

Shell completions aren't installed by `dunst`'s `Makefile`, hence aren't packaged by distribution policy to provide vanilla packages. Adding them, even gated by a flag, would legitimate packaging them.

### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon 1.9.2 (2023-04-20)`
- Install type: `extra/dunst 1.9.2-1`
- Distro: `Arch Linux`

zappolowski:
This resolves #1262.


fwsmit:
LGTM. If we ship completions by default, they have to be removed from the contrib directory, though. That contains only community additions.

zappolowski:
>  If we ship completions by default, they have to be removed from the contrib directory, though.

Do you have a preferred directory for them to be moved to?

I just quickly went through some project (distilled from installed bash completions on my system):
| project   |           directory |
| :-------- | ------------------: |
| `mako`    |           `contrib` |
| `ninja`   |              `misc` |
| `podman`  |       `completions` |
| `systemd` | `shell-completions` |

fwsmit:
I like "completions" the best. So if you don't have a strong preference you can go for that

zappolowski:
@fwsmit I just realized that when completions are installed by default, we should really make `jq` an official dependency (even more necessary with the changes to `dunstctl` pending in #1212).

If you agree on this, I would:
1. add `jq` to the README as dependency (this PR)
2. rework the existing completions to (in another PR)
   1. get rid of the fallback to `awk` in case `jq` is not installed (`dunstify.fishcomp`, `dunstctl.fishcomp`)
   2. add `jq` based parsing to `dunstctl.bashcomp`

Do you agree on this approach?

fwsmit:
Jq is a little bit of a weird dependency to have for dunst. Are there simple alternatives? Maybe the output of dunstctl history and rules could be tweaked?
Also, awk seems to be required for at least the zsh completions.

zappolowski:
> Jq is a little bit of a weird dependency to have for dunst.

I partially agree here, but I currently don't see any (time and code) efficient way of getting the needed information out of the DBus data while keeping `dunstctl` a simple shell script as it currently is. One way of getting forward would be to rewrite `dunstctl` in C.

> Maybe the output of dunstctl history and rules could be tweaked?

Do you have any proposals here? I don't see how it can be made more simple than an array of array of key-values pairs (the inner one basically forming a dictionary). If the output of DBus needs to be parsed (which is required to provide the actual data), it will be really ugly in any kind of shell if there are no external dependencies are allowed - e.g. just getting multi line strings properly parsed will be one major issue.

> Also, awk seems to be required for at least the zsh completions.

`systemd` (through `busctl`) are also (not mentioned) dependencies since the introduction of the `history` subcommand of `dunstctl`.

Some more thoughts of how to get forward:
* don't install completions or any other scripts which require external dependencies at all - leave it to the user to get them from the repository - basically closing this PR
* drop human-readable formatting of `dunstctl rules` (not related to this PR, but it's affected by this discussion) as it is not easily feasible without introducing new dependencies

fwsmit:
> > Jq is a little bit of a weird dependency to have for dunst.
> 
> I partially agree here, but I currently don't see any (time and code) efficient way of getting the needed information out of the DBus data while keeping `dunstctl` a simple shell script as it currently is. One way of getting forward would be to rewrite `dunstctl` in C.
> 
> > Maybe the output of dunstctl history and rules could be tweaked?
> 
> Do you have any proposals here? I don't see how it can be made more simple than an array of array of key-values pairs (the inner one basically forming a dictionary). If the output of DBus needs to be parsed (which is required to provide the actual data), it will be really ugly in any kind of shell if there are no external dependencies are allowed - e.g. just getting multi line strings properly parsed will be one major issue.

Yeah, the only thing I can think of is to do some clever grepping of the output of dunstctl history. But that's more flakey than properly parsing the json. So I think adding the dependency is best. Most distro's ship jq anyways.


> > Also, awk seems to be required for at least the zsh completions.
> 
> `systemd` (through `busctl`) are also (not mentioned) dependencies since the introduction of the `history` subcommand of `dunstctl`.

Good point, could you add it, while you're at it? 

> 
> Some more thoughts of how to get forward:
> 
>     * don't install completions or any other scripts which require external dependencies at all - leave it to the user to get them from the repository - basically closing this PR
> 
>     * drop human-readable formatting of `dunstctl rules` (not related to this PR, but it's affected by this discussion) as it is not easily feasible without introducing new dependencies

Yeah, I just wanted to bring it up to brainstorm a bit about alternatives, but they all seem worse than just adding the dependency.

bynect:
As a side note, is the gdk-pixbuf dependency missing in the documentation?

Narrat:
> As a side note, is the gdk-pixbuf dependency missing in the documentation?

At least it isn't mentioned in the README. Was mentioned in the CHANGELOG and RELEASE_NOTES

bynect:
I think it should be added in the readme, alongside librsvg which is a runtime dependency (of gdk-pixbuf)

> > As a side note, is the gdk-pixbuf dependency missing in the documentation?
> 
> At least it isn't mentioned in the README. Was mentioned in the CHANGELOG and RELEASE_NOTES

I think it should be added in the readme, alongside librsvg which is a runtime dependency (of gdk-pixbuf)

zappolowski:
@bynect @Narrat I suggest to merge this branch as is and update the dependencies on another (dedicated) PR.

bynect:
Fine by me. I noticed that several dependencies are outdated both in the readme, wiki pages and documentation in general. We may want also to update the website with some more up to date info.

Makefile
```diff
@@ -235,9 +235,26 @@ install-dunstify: dunstify
...
 
+ifneq (0,${COMPLETIONS})
+install: install-completions
+install-completions:
+	install -Dm644 completions/dunst.bashcomp ${DESTDIR}${BASHCOMPLETIONDIR}/dunst
+	install -Dm644 completions/dunstctl.bashcomp ${DESTDIR}${BASHCOMPLETIONDIR}/dunstctl
+	install -Dm644 completions/_dunst.zshcomp ${DESTDIR}${ZSHCOMPLETIONDIR}/_dunst
+	install -Dm644 completions/_dunstctl.zshcomp ${DESTDIR}${ZSHCOMPLETIONDIR}/_dunstctl
+	install -Dm644 completions/dunst.fishcomp ${DESTDIR}${FISHCOMPLETIONDIR}/dunst
+	install -Dm644 completions/dunstctl.fishcomp ${DESTDIR}${FISHCOMPLETIONDIR}/dunstctl
```
bynect, line 246: stupid question: what happens if I don't have fish shell? does this command fail or does it do nothing?

zappolowski: It just copies a file during `make install`. None of the shells is actually involved at this stage.

bynect: So if the path doesn't exist it does nothing 👍

zappolowski: I'm not sure what you mean by that, but the path will be created within `$DESTDIR` and the file will be put there. It's just that the code therein isn't run.

bynect: Ah, I was just wondering what happened if the path was not found. So it creates the dirs

bynect: What about doing a simple check like

```make
install-completions:
	install -Dm644 completions/dunst.bashcomp ${DESTDIR}${BASHCOMPLETIONDIR}/dunst
	install -Dm644 completions/dunstctl.bashcomp ${DESTDIR}${BASHCOMPLETIONDIR}/dunstctl
	[ -d "${DESTDIR}${ZSHCOMPLETIONDIR}" ] && install -Dm644 completions/_dunst.zshcomp ${DESTDIR}${ZSHCOMPLETIONDIR}/_dunst
	[ -d "${DESTDIR}${FISHCOMPLETIONDIR}" ] && install -Dm644 completions/_dunstctl.zshcomp ${DESTDIR}${ZSHCOMPLETIONDIR}/_dunstctl
	[ -d "${DESTDIR}${FISHCOMPLETIONDIR}" ] && install -Dm644 completions/dunst.fishcomp ${DESTDIR}${FISHCOMPLETIONDIR}/dunst
	[ -d "${DESTDIR}${FISHCOMPLETIONDIR}" ] && install -Dm644 completions/dunstctl.fishcomp ${DESTDIR}${FISHCOMPLETIONDIR}/dunstctl

```

bynect:
Everything seems fine. I was wondering if we should check if the completion dirs exist before writing to them (I left a suggestion in the comments). But it's not really important I guess

Makefile
zappolowski: I don't see why this check should be needed. The `-D` flag of `install` already takes care to make sure that the path exists before copying over the file.

zappolowski: I think, we have a misunderstanding what 

> make sure that the path exists

means. I use it in the way, that the path is created (not checked for its existence) before any files are copied and this is also what is documented in the man page of `install`.

If you goal is to skip installation of bash, fish or zsh completion if any shell is not installed, this won't generally work as `$DESTDIR` is a common idiom used e.g. in packaging where an empty directory is used (to avoid artifacts in the package). Thus existence checks will never succeed and thus no shell completions would ever be installed.

bynect: Right, my fault 🤣
Go ahead and merge

fwsmit: You should be able to do the merging @bynect. You can do that when you think it's ready

PR #1268, opened 2024-01-25
Corners patch 2

bynect:
This is the 2* part of #1227.

Before we added the scaffolding for custom corners and fixed their behavior.
With this patch we add the ability to customize which corners to round trough the options:
`corners`, `progress_bar_corners` and `icon_corners`

Along with these options I updated the documentation, dunstrc and added a couple tests.


bynect:
it seems like the string parsing causes problem with valgrind...

bynect:
@fwsmit ready for me 👍🏻

Issue #1228, opened 2023-11-09
Test program has segfault


apprehensions:
```
~/src/dunst-og $ make test-valgrind
valgrind \
        --suppressions=.valgrind.suppressions \
        --leak-check=full \
        --show-leak-kinds=definite \
        --errors-for-leak-kinds=definite \
        --num-callers=40 \
        --error-exitcode=123 \
        ./test/test -v
==17713== Memcheck, a memory error detector
==17713== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==17713== Using Valgrind-3.21.0 and LibVEX; rerun with -h for copyright info
==17713== Command: ./test/test -v
==17713== 

* Suite suite_utils:
PASS test_string_replace_char:  (1081 ticks, 0.001 sec)
PASS test_string_replace_all:  (2692 ticks, 0.002 sec)
PASS test_string_append:  (2841 ticks, 0.002 sec)
PASS test_string_strip_quotes:  (1186 ticks, 0.001 sec)
PASS test_string_strip_delimited:  (2480 ticks, 0.002 sec)
PASS test_string_to_path:  (1484 ticks, 0.001 sec)
PASS test_string_to_time:  (1038 ticks, 0.001 sec)

7 tests - 7 passed, 0 failed, 0 skipped (22409 ticks, 0.022 sec)

* Suite suite_option_parser:
PASS test_cmdline_get_string:  (2322 ticks, 0.002 sec)
PASS test_cmdline_get_list:  (1930 ticks, 0.001 sec)
PASS test_cmdline_get_path:  (1033 ticks, 0.001 sec)
PASS test_cmdline_get_int:  (2041 ticks, 0.002 sec)
SKIP test_cmdline_get_double: Skipping test_cmdline_get_double, as it seems we're running under musl+valgrind! (6486 ticks, 0.006 sec)
PASS test_cmdline_get_bool:  (964 ticks, 0.000 sec)
PASS test_cmdline_create_usage:  (3585 ticks, 0.003 sec)
PASS test_string_to_int:  (1064 ticks, 0.001 sec)
PASS test_string_to_int_invalid:  (1231 ticks, 0.001 sec)
SKIP test_string_to_double: Skipping test_string_to_double, as it seems we're running under musl+valgrind! (183 ticks, 0.000 sec)
PASS test_string_to_double_invalid:  (2142 ticks, 0.002 sec)
PASS test_string_to_enum:  (651 ticks, 0.000 sec)
[...]

PR #1269, opened 2024-01-25
Assert that notification icon is not null in tests

bynect:
This is a temporary fix to #1228 until we figure out what causes the issue in first place (icons not being loaded). 

This simply adds assertion so that if the icon is null the test suite will not segfault

The ci will fail...

### DIFF
--- a/test/icon-lookup.c
+++ b/test/icon-lookup.c
@@ -79,9 +79,10 @@ TEST test_new_icon_overrides_raw_icon(void) {
         rule->new_icon = g_strdup("edit");
 
         ASSERT(n->icon);
-
         int old_width = cairo_image_surface_get_width(n->icon);
         rule_apply(rule, n);
+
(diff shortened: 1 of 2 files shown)


bynect:
The ci actually failed with another error unrelated to the changes and is not reproducible locally...
quite the mystery

fwsmit:
I think the memory leak was addressed by @zappolowski in #1274

fwsmit:
This change seems fine to me. Although the tests will probably not pass locally for you now then?

bynect:
> This change seems fine to me. Although the tests will probably not pass locally for you now then?

Yes they fail but I still don't know why... but at least no segfault

apprehensions:
It is a better idea to skip the test if the icon is missing.

fwsmit:
It should be SKIP()

Issue #1270, opened 2024-01-25
Recursive icon lookup not finding theme 


bynect:
### Issue description

I noticed that recursive icon lookup does not work anymore.
I set the icon_theme to Adwaita and `enable_recursive_lookup` to true.
I tried both the system installed version and the latest git commit.

The theme is in `/usr/share/icons/Adwaita`.
But when I enabled debug verbosity it seems like this path is not even tested

### Installation info

- Version: latest git
- Window manager / Desktop environment: `dwm`
- Distro: gentoo

<details>
<summary>Minimal dunstrc</summary>

```ini
# See dunst(5) for all configuration options

[global]

    ### Icons ###

    # Recursive icon lookup. You can set a single theme, instead of having to
    # define all lookup paths.
    enable_recursive_icon_lookup = true

    # Set icon theme (only used for recursive icon lookup)
    icon_theme = Adwaita
    # You can also set multiple icon themes, with the leftmost one being used first.
    # icon_theme = "Adwaita, breeze"

    # Paths to default icons (only neccesary when not using recursive icon lookup)
    icon_path = /usr/share/icons/

    ### Misc/Advanced ###

    # dmenu path.
    dmenu = /usr/local/bin/dmenu -p dunst:

    # Browser for opening urls in context menu.
    browser = /usr/local/bin/xdg-open


```
</details>

PR #1271, opened 2024-01-25
Always check /usr/share/icons in recursive lookup

bynect:
As the name says. It may be good since /usr/share/icons is so ubiquitous to check it anyway. Especially if `XDG_DATA_DIRS` is changed unknowingly from the user (like it happened to me  #1270).

### DIFF
--- a/src/icon-lookup.c
+++ b/src/icon-lookup.c
@@ -166,8 +166,9 @@ void get_theme_path() {
         add_paths_from_env(theme_path, "XDG_DATA_HOME", "icons", data_home_default);
         g_free(data_home_default);
 
-        add_paths_from_env(theme_path, "XDG_DATA_DIRS", "icons", "/usr/local/share/:/usr/share/");
+        add_paths_from_env(theme_path, "XDG_DATA_DIRS", "icons", "/usr/local/share/");
         g_ptr_array_add(theme_path, g_strdup("/usr/share/pixmaps"));
+        g_ptr_array_add(theme_path, g_strdup("/usr/share/icons"));
         for (int i = 0; i < theme_path->len; i++) {
                 LOG_D("Theme locations: %s\n", (char*)theme_path->pdata[i]);
         }


zappolowski:
According to [the spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) an unset or empty `XDG_DATA_DIR` should be regarded as being set to `/usr/local/share:/usr/share` - as it was implemented before.

If the user explicitly sets `XDG_DATA_DIR` incorrectly we should not try to fix it. Maybe the intention was to get rid of `/usr/share` in that case? We don't know.

IMHO: I consider the problem described in #1270 a user configuration issue and thus nothing which should be changed by `dunst`.

bynect:
> According to [the spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) an unset or empty `XDG_DATA_DIR` should be regarded as being set to `/usr/local/share:/usr/share` - as it was implemented before.
> 
> If the user explicitly sets `XDG_DATA_DIR` incorrectly we should not try to fix it. Maybe the intention was to get rid of `/usr/share` in that case? We don't know.
> 
> IMHO: I consider the problem described in #1270 a user configuration issue and thus nothing which should be changed by `dunst`.

This could be valid if dunst didn't also check `/usr/share/pixmaps` on its own accord. Also if you check /usr/share/icons always even if XDG_DATA_DIRS is empty the behavior is the exact same. The only difference would be the order of resolution in this case, but I can invert the line of share/bitmaps and share/icons.

I don't see how making a 2 line change to improve dunst ease of use is bad honestly

zappolowski:
It's hard to judge from the description of the original issue how the messed up `XDG_DATA_DIR` came to be. If it's installed from a package, it's the package maintainer to blame. `XDG_DATA_DIR` should be augmented and not replaced in such cases. (Just out of curiosity: how did you install wine in such a way that it needs to adjust `XDG_DATA_DIR`?)

> even if XDG_DATA_DIRS is empty the behavior is the exact same
If I actually want to use an adjusted `XDG_DATA_DIR` now (e.g. with `/usr/share` explicitly removed), it's plain impossible after your changes - the behavior changed in a breaking way.

Adding `/usr/share/pixmaps` is something different than changing the interpretation of of a specified environment variable.

fwsmit:
Changing you XDG_DATA_DIRS probably causes more issues than just dunst icons, so it's better to report this issue to the gentoo wine package

bynect:
> Changing you XDG_DATA_DIRS probably causes more issues than just dunst icons, so it's better to report this issue to the gentoo wine package

To be fair XDG_DATA_DIRS was unset before and I didn't think that it was used by dunst for icon resolution. Maybe it should be made clearer when the theme is not found for easier troubleshooting

zappolowski:
> Maybe it should be made clearer when the theme is not found for easier troubleshooting

I don't understand this one. When `icon_theme` in `dunstrc` references an unknown icon theme (or a icon theme used therein inherits from another icon theme which cannot be found), `dunst` already issues a warning. Shouldn't this be enough for troubleshooting?

> To be fair XDG_DATA_DIRS was unset before and I didn't think that it was used by dunst for icon resolution.

Maybe I'm misunderstanding you, but `XDG_DATA_DIR` was obviously used ... you even changed the line using it.

> I just did emerge wine and it set XDG_DATA_DIRS which was previously unset.

I don't have Gentoo myself, but asked a friend to run `emerge wine` and from what I could see it sets up `XDG_DATA_DIRS` to `/usr/local/share:/usr/share:/etc/eselect/wine/share`, which seems correct and does not need your fix, as `/usr/share` is still present. How does your "messed up" version look like?

> After being set dunst wasn't working anymore, but I didn't notice at first. Now, since /usr/share/icons is quite ubiquitous I don't why having an extra check is problematic.

My main point is, that no fix should be needed when `XDG_DATA_DIRS` is set correctly and thus solving the root issue (incorrect `XDG_DATA_DIRS`) seems to be the better approach. Also, although this fix might work for `dunst`, all other tools adhering to the Icon Theme Specification in combination with the XDG Base Directory Specification would still be broken and would need fixing as well.

> If you want the user to be able to set a different icon path they can simply set xdg_data_dirs and it will have precedence in the resolution order. The same problem could be said anyway for the /usr/share/bitmaps path, which is searched even if the user didn't set it in the environment.

`/usr/share/bitmaps` is not searched, but `/usr/share/pixmaps` and as already stated, that's because it's mandated in the Icon Theme Specification.

> So in the end if a power user wants to change the icon path they can

That's incorrect, as there's no way to get rid of `/usr/share/icons` now, as that one is enforced.

fwsmit:
Let's stick to the spec. I'll close this PR.

PR #1274, opened 2024-01-28
Add tests for NotificationListHistory

zappolowski:
... and while at it also fix all memory leaks from this method.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -331,10 +331,8 @@ static void dbus_cb_dunst_NotificationListHistory(GDBusConnection *connection,
 {
         LOG_D("CMD: Listing all notifications from history");
 
-        GVariant *answer = NULL;
-        GVariantBuilder *builder;
-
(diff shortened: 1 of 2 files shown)


fwsmit:
Could you explain the memory leak?

zappolowski:
From what I've seen it comes from `g_variant_new_from_bytes` (which is not necessary here and can be avoided by using typed versions like I did) and `g_variant_type_new`, which even states, that it transfers ownership and thus should be freed.

zappolowski:
I've added some minor clean up:
* avoid heap allocation of the outer `builder` - just leave it on the stack as management is easier this way
* inline `answer` just for conciseness - `g_dbus_method_invocation_return_value` takes care of the floating variant created by `g_variant_new` already

PS: I cannot reproduce the pipeline failure. That image ran successfully on my machine.

bynect:
looks good. Are you done with the changes or something is still pending?

Also friso is away at the moment

PR #1276, opened 2024-02-04
Wayland: suport GTK/CSS cursor names

alebastr:
Adwaita, the default icon theme for many distributions, decided to stop providing legacy Xcursor icons in favor of the GTK/CSS names[1]. This requires us to probe for "default" instead of the "left_ptr", as the latter is no longer available in Adwaita.
Keep the fallback to "left_ptr" for more conservative themes.

[1]: https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/commit/74e9b79471236320d2af4925d6c5bb7df22380ce

### DIFF
--- a/src/wayland/wl.c
+++ b/src/wayland/wl.c
@@ -750,9 +750,14 @@ bool wl_init(void) {
                 LOG_W("couldn't find a cursor theme");
                 return true;
         }
-        struct wl_cursor *cursor = wl_cursor_theme_get_cursor(ctx.cursor_theme, "left_ptr");
+        // Try cursor spec (CSS) name first
+        struct wl_cursor *cursor = wl_cursor_theme_get_cursor(ctx.cursor_theme, "default");
         if (cursor == NULL) {
-                LOG_W("couldn't find cursor icon \"left_ptr\"");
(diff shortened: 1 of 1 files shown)


bynect:
I was wondering if it isn't better to first check for left_ptr and use default as a fallback. Because I was searching around and SDL checks first for left_ptr.

alebastr:
Doesn't really matter, `left_ptr` and `default` should be equivalent in all the themes that have both (that's any cursor theme updated within the past decade).

I've seen both orders in major projects, and [some projects](https://github.com/swaywm/sway/pull/7649/files) don't even have a fallback to `left_ptr`.

bynect:
Thanks for the information, I'll merge the pr 👍🏻 

Also, sorry to bother you but I am not really familiar with the wayland output. Do you know if/how to support for transparent colors in wayland? Is it already working?

alebastr:
> Also, sorry to bother you but I am not really familiar with the wayland output. Do you know if/how to support for transparent colors in wayland? Is it already working?

It works if you specify the colors in RGBA format. E.g. `notify-send -h string:bgcolor:#0000007f ...`

PR #1278, opened 2024-02-10
Add instructions to reload the .dunstrc file without restarting.

bobkron:
Since dunst does not have a reload feature, added instructions in the docs to reload `.dunstrc` after making changes,
**without** needing to restart the window manager every time a change is made during configuration.

### DIFF
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -24,7 +24,7 @@ All experimental settings are marked with I<Experimental>
 
 To Reload C<.dunstrc>, run this command in your terminal :
  
-        $ killall dunst;notify-send foo
+        $ killall dunst
 
 =head2 Global section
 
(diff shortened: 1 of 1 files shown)


bynect:
I would say that killall dunst suffices, since the notification daemon is started automatically whenever a notification is sent

bobkron:
Actually yes, that would suffice...will make the changes right away.

Narrat:
Wouldn't it be enough to just describe what to do? There may be systems where `killall` is not available. And as there are a lot of different programs around for such a task (`kill`, `killall`, `pkill`, ...) it could be benefical to just explain in general what to do.

bynect:
> Wouldn't it be enough to just describe what to do? There may be systems where `killall` is not available. And as there are a lot of different programs around for such a task (`kill`, `killall`, `pkill`, ...) it could be benefical to just explain in general what to do.

you have a point. maybe just saying: if you kill dunst (eg killall dunst) it will be started automatically when a notification is sent

PR #1282, opened 2024-02-16
Fix compiler warning/error on master

zappolowski:
The callback `output_handle_description` is just used when `WL_OUTPUT_NAME_SINCE_VERSION` is defined and so, this function should defined only then as well as it otherwise is flagged as being unused which breaks the build.


fwsmit:
I added this function while fixing a merge conflict, but didn't notice the error. Thanks for fixing

PR #1284, opened 2024-02-18
Suppress alpine allocation issue

zappolowski:
This is related to the introduction of `wordexp` in e9a27c0 but did go unnoticed as it seems to be detected just by newer versions of valgrind (I would rule out musl here as the `wordexp` code was not touched in ages and `getdelim` was last mentioned in the 1.1.22 version while the old alpine image already shipped 1.1.24).

This isn't an issue with the current images but newer versions of alpine fail to complete without it (see [this comment](https://github.com/dunst-project/docker-images/pull/8#issuecomment-1948756512)).

### DIFF
--- a/.valgrind.suppressions
+++ b/.valgrind.suppressions
@@ -19,6 +19,16 @@
    obj:*
 }
 
+# Ignore yet another musls' weird error
+{
+   musl_alpine_libc
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:realloc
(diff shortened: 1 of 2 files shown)


zappolowski:
@fwsmit This also includes the changes from #1282 to have a working pipeline. The branches should be merged in creation order (the superflous commit) should vanish then.

fwsmit:
Hopefully the ci works now

Issue #1273, opened 2024-01-27
Feature: Color gradient


bynect:
Hello, I was fiddling around and implementing a color gradient feature.
It is still unfinished though and I would like to hear your ideas.

I wanted to add the option to have a gradient to the background and progress bar.
Basically we have a list a colors that are put in a `cairo_pattern_t`.
If there is only one color the behavior is the same as it is now.

For the syntax of I was thinking of one of the following
```ini
# separate
background = whatever # ignored
background_gradient = 0xffaa22,0xfaf2a
```

or 

```ini
background = red, green, blue
```

Example:
![img-1706811365](https://github.com/dunst-project/dunst/assets/68197565/d701da6d-0102-4484-962a-df3d0200d504)

fwsmit:
I'm just seeing this now. If you want me to react, you can tag me.
The gradient looks cool. Currently the color of the progress bar is called the highlight color. You can expand that one to also accept a list.

bynect:
> I'm just seeing this now. If you want me to react, you can tag me.
> The gradient looks cool. Currently the color of the progress bar is called the highlight color. You can expand that one to also accept a list. 

Ok so I'll go with the list ruote 👍

I'll present a pr when I'm done

PR #1286, opened 2024-02-19
Gradients

bynect:
This is the implementation of #1273.

At the moment I only implemented the gradient for `highlight`, but it can be easily applied also to `background`, `foreground` etc.
I am using a cairo linear pattern which allows for any number of colors so the user is free.

TODO:
* Maybe some extra testing, at the moment everything seems to work smoothly though
* Apply gradients to more colors settings
* Should we apply delta corrections like the one made by calculate_foreground_color?

## Example
img-1708384655
```ini
highlight = "#ffffff, #cc397b, #39ff8a"
```


bynect:
@fwsmit gladly the ci passed, so let me know what do you think 👍🏻

fwsmit:
What do you mean with delta correction?
One thing that may be cool to add is an angle for the gradient. But already looks cool like this

bynect:
> What do you mean with delta correction?
> One thing that may be cool to add is an angle for the gradient. But already looks cool like this 

Good idea. I was thinking of adding the possibility to provide a lost of xy points for the gradient breaks. But just an angle is also fine.

Let me know if I should add gradient support for other color.

For the delta correction is something that's done for the foreground but I don't think it's needed

src/draw.c
```diff
@@ -29,11 +29,16 @@ struct color {
         double a;
 };
 
+struct gradient {
+        struct color *cs;
```
zappolowski, line 33: I'd suggest to use `colors` here as it is easier to follow on the usage side.

src/draw.c
```diff
@@ -120,6 +125,40 @@ static struct color string_to_color(const char *str)
...
+                g.cs = g_malloc(sizeof(struct color));
+                g.cs[g.length++] = hex_to_color(0xF, 1);
+                /* return black (only one color) on error */
+        } else {
+                g.cs = g_malloc0(g_strv_length(strv) * sizeof(struct color));
+                for (int i = 0; strv[i] != NULL; i++) {
+                        char *end, *str = strv[i];
+                        if (STR_EMPTY(str)) continue;
+
+                        uint_fast32_t val = strtoul(str+1, &end, 16);
```
zappolowski, line 144: Why `uint_fast32_t` ... shouldn't it be `unsigned long`?

Are you using `str+1` here to skip the initial `#`? This should be done, iff it is actually a `#`.

How about handling of named colors like `Yellow`? Maybe I'm missing something here, but I couldn't find, where this is handled.

src/draw.c
```diff
@@ -120,6 +125,40 @@ static struct color string_to_color(const char *str)
...
+                                LOG_W("Invalid color string: '%s'", str);
+                                continue;
+                        }
+
+                        switch (end - (str+1)) {
+                                case 3:  g.cs[g.length++] = hex_to_color((val << 4) | 0xF, 1); break;
+                                case 6:  g.cs[g.length++] = hex_to_color((val << 8) | 0xFF, 2); break;
+                                case 4:  g.cs[g.length++] = hex_to_color(val, 1); break;
+                                case 8:  g.cs[g.length++] = hex_to_color(val, 2); break;
+                        }
```
zappolowski, line 155: There should be a warning, when an incorrectly formatted color was found here.

src/draw.c
```diff
@@ -120,6 +125,40 @@ static struct color string_to_color(const char *str)
...
+        if (strv == NULL) {
+                LOG_W("Invalid empty color gradient string");
+                g.cs = g_malloc(sizeof(struct color));
+                g.cs[g.length++] = hex_to_color(0xF, 1);
+                /* return black (only one color) on error */
+        } else {
+                g.cs = g_malloc0(g_strv_length(strv) * sizeof(struct color));
+                for (int i = 0; strv[i] != NULL; i++) {
+                        char *end, *str = strv[i];
+                        if (STR_EMPTY(str)) continue;
```
zappolowski, line 142: Just for understanding: this would mean an empty element would be allowed and yield a opaque black?

src/notification.c
```diff
@@ -68,8 +68,9 @@ void notification_print(const struct notification *n)
         printf("\tformatted: '%s'\n", n->msg);
         printf("\tfg: %s\n", n->colors.fg);
         printf("\tbg: %s\n", n->colors.bg);
-        printf("\thighlight: %s\n", n->colors.highlight);
-        printf("\tframe: %s\n", n->colors.frame);
+        printf("\thighlight: ");
+        for (int i = 0; n->colors.highlight[i]; i++) printf("%s ", n->colors.highlight[i]);
```
zappolowski, line 72: Nit pick: this would leave a trailing white space and moving the line break to the following entry seems a bit fragile (when lines are moved around). How about something like

```c
gchar *highlight = g_strjoinv(" ", n->colors.highlight);
printf("\thighlight: %s\n", highlight);
g_free(highlight);
```

bynect: Fair enough, I wasn't satisfied with that in the first place

src/draw.c
bynect: I am following the previous behavior

src/draw.c
bynect: If you mean the variable name, sure

src/draw.c
bynect: Again, i am following the previous behavior (but I don't like it either)

bynect:
@fwsmit @zappolowski I don't particularly like the way dunst decodes color strings at every draw call, since it makes more difficult validation and is plain inefficient. I was thinking of storing in the config directly the color struct and parse once in the option_parser.

Is there something that prevents me from doing that (like protocols) or can I go ahead? 

Obviously I would close this pr and make it again later (and better) to store gradients in the config

Also I would remove the fact that invalid colors are black and just error out on parsing

zappolowski:
I had a similar idea in mind, but TBH I didn't have the time yet to fully dig through all the code and thus cannot foresee what the implications are. From quickly skimming through the Desktop Notification Specification I could not find anything which prevents this (e.g. sending color information along with the actual notification data).

Having said that, I think parsing should be done once when reading the configuration and not on every notification being created. So, go for it, but maybe do this refactoring first and then (on top of that) add the new gradient feature.

bynect:
> I had a similar idea in mind, but TBH I didn't have the time yet to fully dig through all the code and thus cannot foresee what the implications are. From quickly skimming through the Desktop Notification Specification I could not find anything which prevents this (e.g. sending color information along with the actual notification data).
> 
> Having said that, I think parsing should be done once when reading the configuration and not on every notification being created. So, go for it, but maybe do this refactoring first and then (on top of that) add the new gradient feature.

Yes that was the plan, and it will make the whole feature easier

bynect:
I will rework this completely. At the moment my idea is to store a cairo_pattern_t in the gradient struct

Issue #1171, opened 2023-04-24
1.9.2: core dump in self tests


0-wiz-0:
When running the self tests on NetBSD 10.99.3/amd64, I see a segfault:
```
PASS test_notification_referencing:  (0 ticks, 0.000 sec)
bash: line 2: 14069 Segmentation fault      (core dumped) ./test/test -v
     11348 Done                    | ./test/greenest.awk
gmake: *** [Makefile:76: test] Error 139
```

bynect:
could you add the result of test-valgrind please?

0-wiz-0:
Sorry, no, valgrind does not support NetBSD.

0-wiz-0:
But here's a backtrace of the core dump:
```
Core was generated by `test'.
Program terminated with signal SIGSEGV, Segmentation fault.
t#0  0x0000770644f242c2 in _cairo_surface_is_image (surface=<optimized out>) at ../src/cairo-image-surface-inline.h:77
77      ../src/cairo-image-surface-inline.h: No such file or directory.
(gdb) bt
#0  0x0000770644f242c2 in _cairo_surface_is_image (surface=<optimized out>) at ../src/cairo-image-surface-inline.h:77
#1  cairo_image_surface_get_width (surface=0x0) at ../src/cairo-image-surface.c:641
#2  0x0000000000914ad7 in get_icon_width (icon=<optimized out>, scale=scale@entry=1) at test/../src/icon.c:87
#3  0x000000000091dbb0 in test_notification_icon_scaling_toosmall () at test/notification.c:154
#4  suite_notification () at test/notification.c:240
#5  0x00000000009306a0 in greatest_run_suite (suite_name=<optimized out>, suite_cb=0x91d3ba <suite_notification>) at test/test.c:33
#6  greatest_run_suite (suite_cb=0x91d3ba <suite_notification>, suite_name=<optimized out>) at test/test.c:33
#7  0x000000000093789b in main (argc=2, argv=0x7f7fff2ec678) at test/test.c:57
```

bynect:
I see. I am not sure but this seems similar to a problem I have encountered (#1228)

bynect:
I made a temporary fix #1269 which is merged on master. Could you try doing this on the master branch to see if it is resolved?

0-wiz-0:
git head completes the test suite for me, with:
```
Pass: 185, fail: 5, skip: 0.
```
and the following tests failing (but no core dumps):
```
[...]

PR #1288, opened 2024-02-21
Portability fixes

0-wiz-0:
These patches are from pkgsrc and increase portability of dunst to other operating systems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ LDFLAGS := ${DEFAULT_LDFLAGS} ${LDFLAGS} ${LIBS}
 
 ifeq (0,${WAYLAND})
 # without wayland support
-SRC := $(sort $(shell ${FIND} src/ -not \( -path src/wayland -prune \) -name '*.c'))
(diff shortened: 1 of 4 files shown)


bynect:
Related to #1171.

For the features.h it seems like it is used just to check `__GLIBC__ ` once, so maybe we can change it so we never need it?

Similarly for the button codes, I think we can remove the reliance on linux codes and just define an enum. (since they are arbitrary number anyway and are not used nowhere else)
What do you think?@fwsmit

fwsmit:
What operating systems are you trying to port to? That would be good to know.
The button codes are plain numbers, but we get them from wayland / X11 I believe. So we need to use those specific numbers. I didn't check it though

0-wiz-0:
pkgsrc is cross-platform.
The evdev patches are needed at least on NetBSD, and the find patch is (I think) needed on Solaris/Illumos.

bynect:
I resolved a conflict with #1290.
Also thanks for the pr, I'll merge it 👍🏻

PR #1293, opened 2024-02-26
Split wl.c into multiple files

alebastr:
Non-functional change: the refactoring requested in https://github.com/dunst-project/dunst/pull/1250#issuecomment-1917988542


bynect:
Everything seems fine 👍🏻

PR #1295, opened 2024-02-29
Fix test when building wayland only

bynect:
Also stop the makefile if neither X11 or WAYLAND are set


bynect:
Perfect 👍🏻
Sorry for the error in #1290

PR #1296, opened 2024-02-29
Fix logging in icon code

bynect:
Remove extra newlines and move a string empty upwards

### DIFF
--- a/src/icon-lookup.c
+++ b/src/icon-lookup.c
@@ -38,7 +38,7 @@ int get_icon_theme(char *name) {
  * @retval -1 means no index was found
  */
 int load_icon_theme_from_dir(const char *icon_dir, const char *subdir_theme) {
-        LOG_D("Loading theme %s/%s\n", icon_dir, subdir_theme);
+        LOG_D("Loading theme %s/%s", icon_dir, subdir_theme);
         char *theme_index_dir = g_build_filename(icon_dir, subdir_theme, "index.theme", NULL);
         FILE *theme_index = fopen(theme_index_dir, "r");
         g_free(theme_index_dir);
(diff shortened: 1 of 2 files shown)


zappolowski (approved):

Rentib:
Was it expected that setting default_icon = "" uses an icon?

```
[urgency_normal]
  timeout = 10
  default_icon = ""
```

This for example still uses an icon for notifications with normal urgency.
The icon used is the default one.

bynect:
> Was it expected that setting default_icon = "" uses an icon?
> 
> ```
> [urgency_normal]
>   timeout = 10
>   default_icon = ""
> ```
> 
> This for example still uses an icon for notifications with normal urgency. The icon used is the default one.

What do you mean it uses the default icon? Also could you please open an issue

Rentib:
> > Was it expected that setting default_icon = "" uses an icon?
> > ```
> > [urgency_normal]
> >   timeout = 10
> >   default_icon = ""
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This for example still uses an icon for notifications with normal urgency. The icon used is the default one.
> 
> What do you mean it uses the default icon? Also could you please open an issue
> 
> Also, is this something that happened after this commit?

If you specify icon_theme in configuration file, then for all notifications dunst will use an icon from the theme, unless you configure it otherwise. If you check for example Papirus icons then dunst will use emblem-information icon I think.

Example config:
```
[global]
  enable_recursive_icon_lookup = true
  icon_theme = "Adwaita, Papirus"
```

bynect:
> > > Was it expected that setting default_icon = "" uses an icon?
> > > ```
> > > [urgency_normal]
> > >   timeout = 10
> > >   default_icon = ""
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This for example still uses an icon for notifications with normal urgency. The icon used is the default one.
> > 
> > What do you mean it uses the default icon? Also could you please open an issue
> > 
> > Also, is this something that happened after this commit?
> 
> If you specify icon_theme in configuration file, then for all notifications dunst will use an icon from the theme, unless you configure it otherwise. If you check for example Papirus icons then dunst will use emblem-information icon I think.
> 
> Example config:
> ```
> [global]
>   enable_recursive_icon_lookup = true
>   icon_theme = "Adwaita, Papirus"
> ```

I don't get this behavior. Please open an issue including your dunst version and dunstrc so that I can investigate

PR #1297, opened 2024-03-01
Get rid of new warning about dangling pointers

zappolowski:
The same as already done in fe5ac680af0a7c6831205b02360172dbdc06185b but plain buffers were re-introduced when the corners patch landed.

@bynect fyi

### DIFF
--- a/test/option_parser.c
+++ b/test/option_parser.c
@@ -747,7 +747,7 @@ TEST test_string_to_corners(void)
         s.parser = string_parse_corners;
         s.parser_data = corners_enum_data;
 
-        char buf[50];
+        static char buf[50];
 
         // do not go until last element, since it's ENUM_END (all 0)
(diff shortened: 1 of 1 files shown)


bynect:
> The same as already done in [fe5ac68](https://github.com/dunst-project/dunst/commit/fe5ac680af0a7c6831205b02360172dbdc06185b) but plain buffers were re-introduced when the corners patch landed.
> 
> @bynect fyi

Thanks, I didn't notice that the buffers had to be static. Just for information, with which compiler flags are you getting the warnings?

zappolowski:
I get them as warnings by just running the tests locally:
```sh
$ make WAYLAND=0 -j test -B
```

To mimic the (future - new CI images aren't pushed to Dockerhub yet) build failures just export `CFLAGS=-Werror` as theses images run builds this way.

PR #1301, opened 2024-03-01
Don't search for icon path if icons are disabled

### LINKED ISSUE #1114: dunst: Attempts to find icons even when disabled
### Issue description

I disabled icons for all notifications (I don't have them or any theme that contains icons). In my dunstrc I have:

```
[global]
icon_position = off
```

However, I still see the followning on every notification:

`dunst[1589]: WARNING: No icon found in path: 'dialog-information'`

### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon 1.8.1 (2022-03-02)`
- Install type: `debian unstable package 1.8.1-1`
- Window manager / Desktop environment: `spectrwm`

bynect:
Close #1114

### DIFF
--- a/src/notification.c
+++ b/src/notification.c
@@ -340,11 +340,11 @@ void notification_transfer_icon(struct notification *from, struct notification *
 
 void notification_icon_replace_path(struct notification *n, const char *new_icon)
 {
-        ASSERT_OR_RET(n,);
+        ASSERT_OR_RET(n && n->icon_position != ICON_OFF,);
         ASSERT_OR_RET(new_icon,);
-        if(n->iconname && n->icon && strcmp(n->iconname, new_icon) == 0) {
+
+        if (n->iconname && n->icon && STR_EQ(n->iconname, new_icon))
                 return;
-        }
 
         // make sure it works, even if n->iconname is passed as new_icon
         if (n->iconname != new_icon) {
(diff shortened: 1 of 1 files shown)


bynect:
This should work unless you can enable icons to an old notification with disabled icons. Because at that point the notification wouldn't have loaded the icon. However I don't think there is a way to change icon_position between creation and display (?)

src/queues.c
zappolowski, line 1: Unrelated change to the issue.

src/notification.c
```diff
@@ -342,9 +342,10 @@ void notification_icon_replace_path(struct notification *n, const char *new_icon
 {
         ASSERT_OR_RET(n,);
         ASSERT_OR_RET(new_icon,);
-        if(n->iconname && n->icon && strcmp(n->iconname, new_icon) == 0) {
+        ASSERT_OR_RET(n->icon_position != ICON_OFF,);
```
zappolowski, line 345: This could be merged to the first assertion as it is related to `n`:
```c
ASSERT_OR_RET(n && n->icon_position != ICON_OFF,);
```

src/queues.c
bynect: True

zappolowski (approved):
Please rebase to current master and squash the 2nd commit into the first one.

Issue #1094, opened 2022-07-28
Setting min_icon_size = 0 disables recursive icon search


pmattern:
Title says it basically all.

Steps to reproduce:
1. Make sure that icon theme Adwaita is available and Dunst is using the default configuration file.
1. Launch a notification featuring an icon that's provided by the icon theme. The icon will show up.
1. Set `min_icon_size = 0`, either within the default configuration file or a custom copy in `$XDG_CONFIG_HOME/dunst`. Restart Dunst.
1. Repeat step 2. The icon won't show up any more.

Seems to be a regression which was introduced after release 1.8.1. Unfortunately, I lack the time to bisect this right now.
Maybe related to #1086, but the differences warrant an issue on its own, IMHO.

1ef38e5 on Arch Linux x86_64 per [AUR](https://aur.archlinux.org/) package [dunst-git](https://aur.archlinux.org/packages/dunst-git).

fwsmit:
The meaning of min_icon_size has changed slightly. It's now used for looking up icons in your icon theme. Since no icons have size 0, nothing gets found. So it's not a bug, but intended behaviour.
Does this behaviour limit your configuration significantly?

pmattern:
Not sure I understand what the new behavior is meant to be.   
If `min_icon_size = n` is supposed to mean "look for icons of at least size n", then `n=0` shouldn't fail. But if it's supposed to mean "look for icons of exactly size n", then wording `min_icon_size` is pretty misleading.

Whatever the new behavior is meant to be, you badly need to update the configuration file's comment. Wasting time due to plain wrong documentation really sucks, tbh.

That new behavior will probably never be an issue for me (wasting time figuring out why icons stopped showing up was, though).   
But I recon many users won't like it - the purpose according to the current comment ("high-dpi") seems reasonable to me, and in #1086 you've already seen complaints about icons being too small (albeit for a different reason).

fwsmit:
It's looking for icons that can be scaled to a certain size according to the index.theme from an icon theme. It's not looking for bigger icons afterwards. The wording might not be the best but it makes sense in the instance where you have an icon not from your theme since it's scaled between those bounds.

If you think you can improve the documentation, go ahead and submit a pr or be more detailed than 'this sucks'.

I guess we'll hear from users if they have a problem with the new behavior. Until then let's not make it a problem.

cyberpunkrocker-zero:
[...]

PR #1302, opened 2024-03-03
Make recursive icon lookup work with size == 0

### LINKED ISSUE #1116: Some icons can't be found
### Issue description
Run `notify-send --icon=nm-no-connection  hgfhgfhgf`. This icon wont be found

### Installation info

- Version: `1.9.0`
- Install type: `Arch official package`
- Window manager / Desktop environment: `BSPWM`

<details>

```ini
[global]
width="(0, 1200)"
height=9999
enable_recursive_icon_lookup=yes
icon_theme=Papirus
max_icon_size=50
```
</details>

bynect:
Only works with scalable icons, but at least it works.

@fwsmit do you think this is a good enough solution for #1094?

### DIFF
--- a/src/icon-lookup.c
+++ b/src/icon-lookup.c
@@ -253,7 +253,7 @@ char *find_icon_in_theme(const char *name, int theme_index, int size) {
                                 break;
 
                         case THEME_DIR_SCALABLE:
-                                match_size = dir.min_size <= size && dir.max_size >= size;
+                                match_size = (size == 0 || dir.min_size <= size) && dir.max_size >= size;
                                 break;
 
                         case THEME_DIR_THRESHOLD:


fwsmit:
Yeah, i think it's best to get rid of min size. It doesn't make much sense with the new icon lookup, since it only uses one size.

bynect:
> Yeah, i think it's best to get rid of min size. It doesn't make much sense with the new icon lookup, since it only uses one size. 

Honestly I don't know if it is better to ignore the min size for icon lookup. Because at the moment it is used for determining the size of non scalable icons.

In this pr I made it so that if minsize is 0 we get whatever size we found. Which seems sensible... but i don't know if some warning should be issued or something 🤔

fwsmit:
What I mean is that having a max and min size doesn't make much sense right now, since only one size is used for the lookup. So it's best to rename this setting or somehow use both sizes again

bynect:
> What I mean is that having a max and min size doesn't make much sense right now, since only one size is used for the lookup. So it's best to rename this setting or somehow use both sizes again

proposal = make a `icon_size` a length so that you can say (min,max) or size

fwsmit:
Yeah, but what are you doing with scalable icons? Are you making them as big as possible? In practice most icons are scalable, so only the biggest size is used.
For consistency of the notifications it's better for all icons to be the same size. For some notifications, like album art for music, you might want bigger icons. But that should be done with a rule.

So in what case would you want a min and max size?

bynect:
> Yeah, but what are you doing with scalable icons? Are you making them as big as possible? In practice most icons are scalable, so only the biggest size is used.
> For consistency of the notifications it's better for all icons to be the same size. For some notifications, like album art for music, you might want bigger icons. But that should be done with a rule.
> 
> So in what case would you want a min and max size? 

I guess you are referring to making an icon_size length value.

The option is to have a value (min, max) like for width. Or alternatively a single value that makes min =max and in practice would mean that the icon has a fixed size. This latter part could be achieved by setting icon_max_size = icon_min_size but that would change the semantics of those in another way.

So the idea is to implement this more versatile single option that has also a sane behavior if you want fixed size icons

bynect:
I will have to redo this based on #1365

bynect:
Closing this since I have to rework the logic anyway for #1401

PR #1303, opened 2024-03-03
Update GitHub actions

bebehei:
- Fix the initial error message
- Use the newly created images from GitHub Actions/Packages

Fixes dunst-project/docker-images#10


bebehei:
.. no CI spawning? 🤨

bebehei:
Suspected first, GH actions did not trigger. Ok, the `env` was broken. Was only visible, when you checked explicitly the GitHub Actions page.

bebehei:
Ok, now it's failing because of the typical git "dubious ownership" problem. I expected, this wasn't a problem here at all, since it's everywhere the same user.

Apparently, someone else figured out, that UID 1001 should be used:

https://github.com/actions/runner/issues/2033#issuecomment-1598547465

Also, the reference, why UID 1001 is used is there: https://github.com/actions/runner/blob/8415f13babf8199551f817eb16b718dcdd5b31e2/images/Dockerfile#L38

It feels a bit, that someone just thought it's a good idea to assign another UID.

bebehei:
Finally, it's **our fault**, that the pipeline does not run. 🎉

bebehei:
Ok, this is it now for today. We're not dealing anymore with GH Actions/whatever problems, but with our own build pipeline.

I'll be probably back at tuesday.

zappolowski:
@bebehei The error is caused but what I described in [this comment](https://github.com/dunst-project/docker-images/pull/7#issuecomment-1950986704). I've added [this commit](https://github.com/dunst-project/docker-images/pull/7/commits/52541a3da3ddd2e616aa25e40472c8f58e8e5238) to make it work locally. So, something like
```patch
diff --git a/.github/workflows/main.yml b/.github/workflows/main.yml
index 56cf410..6d5d389 100644
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,7 +37,7 @@ jobs:
 
     env:
       CC: ${{ matrix.CC }}
-      EXTRA_CFLAGS: "-Werror"
+      EXTRA_CFLAGS: "-Werror -gdwarf-4"
     steps:
     - uses: actions/checkout@v2
       with:
```
could already make it work (it's just needed for debian-bookworm and ubuntu-jammy, but I couldn't find how to set `env` conditionally ... might have time the coming weekend to have a closer look).

bebehei:
@zappolowski Thanks. I've seen this and had it in my mind. But I also had to kick out the dbus testing suite. I guess this is our other part we have to fix, too.

zappolowski:
@bebehei I took another look into the latest build failures and it's rather wild.

The Debian bookworm image currently lacks a required package `libclang-rt-dev`, but even when added it's not really working as then some linker issues arise ... but just when run automatically; when I entry the container in interactive mode, I can successfully run the `test-coverage` target (which is the one currently failing). Even wilder: it also fails when I manually run `/srv/entrypoint`, but succeeds, when I adjust the shebang to `#!/bin/bash`. Using a non-interactive build (like done in the CI) with using `#!/bin/bash` still fails, though. So, my current guess, is that some setup might be missing.

zappolowski:
@bebehei I figured it out ... at least I cannot reproduce the failure locally anymore.

1. Update the `debian-bookworm` image ... see [this PR](https://github.com/dunst-project/docker-images/pull/11)
2. add `--coverage` to `CFLAGS`, e.g.
  ```diff
diff --git a/.github/workflows/main.yml b/.github/workflows/main.yml
index 3d7c38f..25c2229 100644
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,7 +37,7 @@ jobs:
 
     env:
       CC: ${{ matrix.CC }}
-      EXTRA_CFLAGS: "-Werror ${{ (matrix.distro == 'debian-bookworm' || matrix.distro == 'ubuntu-jammy') && '-gdwarf-4' || ' '}}"
+      EXTRA_CFLAGS: "-Werror ${{ matrix.distro == 'debian-bookworm' && '--coverage -gdwarf-4' || ' '}} ${{ matrix.distro == 'ubuntu-jammy' && '-gdwarf-4' || ' ' }}"
     steps:
     - uses: actions/checkout@v4
       with:
  ```

zappolowski:
I've just updated the debian-bookworm image and it seems like it works without addition of `--coverage` ... for whatever reason.

Currently the pipeline is green, but some tests and alpine is still disabled.

bebehei:
Thanks for merging the stuff in #1312 and caring about it. I guess, I can close this PR then.

PR #1306, opened 2024-03-04
Store colors in a struct

bynect:
I changed every usage of colors from a string to a color struct. Colors are now parsed only once in the option_parser instead of every single draw call.

Note: Hopefully I haven't missed anything. I would be super grateful if someone could do some extra testing on their own if they encounter anything strange


bynect:
Still working on the tests

bynect:
Fixes what we talked about in #1286

bynect:
Note that invalid colors sent to dbus are ignored

fwsmit:
Yeah, looks good!

src/option_parser.c
```diff
@@ -170,36 +170,77 @@ int string_parse_sepcolor(const void *data, const char *s, void *ret)
...
+                        return true;
+                }
         }
+        return false;
 }
 
+#define UINT_MAX_N(bits) ((1 << bits) - 1)
+
+// Parse a #RRGGBB[AA] string
+int string_parse_color(const void *data, const char *s, void *ret)
```
zappolowski, line 198:
```suggestion
int string_parse_color(const void *data, const char *s, struct color *color)
```

Shouldn't the target always be a `struct color *`? Then we should be able to get rid of line 205 as well.

src/draw.h
```diff
@@ -40,6 +36,27 @@ enum corner_pos {
...
 
+struct color {
+        double r;
+        double g;
+        double b;
+        double a;
+};
+
+#define COLOR_UNINIT { -1, -1, -1, -1 }
+#define COLOR_VALID(c) ((c).r >= 0 && (c).g >= 0 && (c).b >= 0 && (c).a >= 0)
```
zappolowski, line 47: Should we check `c.x > 1` here as well (instead of just when trying to print it)? Cairo colors have to be in the [0, 1] range and thus `COLOR_VALID` should enforce this everywhere.

src/option_parser.c
```diff
@@ -170,36 +170,77 @@ int string_parse_sepcolor(const void *data, const char *s, void *ret)
...
+{
+        if (STR_EMPTY(s) || *s != '#') {
```
zappolowski, line 200: I know it was basically this way already before your changes, but `dunst.5` states, that named colors like "Yellow" are supported, which doesn't seem to be the case anymore (I don't know since when). Either it is removed from the documentation (breaking change, should be added - retroactively - to `CHANGELOG.md`) or it is possible to somehow get it back to work (starting point could be [this list](https://gitlab.freedesktop.org/xorg/app/rgb/raw/master/rgb.txt)).

src/draw.h
```diff
@@ -40,6 +36,27 @@ enum corner_pos {
...
+#define COLOR_SAME(c1, c2) ((c1).r == (c2).r && (c1).g == (c2).g && (c1).b == (c2).b && (c1).a == (c2).a)
```
zappolowski, line 48: This one is just used in tests. I think, it should be moved to `test/helpers.h`.

src/notification.c
```diff
@@ -495,14 +498,10 @@ void notification_init(struct notification *n)
...
-        if (!n->colors.bg)
-                n->colors.bg = g_strdup(defcolors.bg);
-        if (!n->colors.highlight)
-                n->colors.highlight = g_strdup(defcolors.highlight);
-        if (!n->colors.frame)
-                n->colors.frame = g_strdup(defcolors.frame);
+        if (!COLOR_VALID(n->colors.fg)) n->colors.fg = defcolors.fg;
+        if (!COLOR_VALID(n->colors.bg)) n->colors.bg = defcolors.bg;
+        if (!COLOR_VALID(n->colors.highlight)) n->colors.highlight = defcolors.highlight;
+        if (!COLOR_VALID(n->colors.frame)) n->colors.frame = defcolors.frame;
```
zappolowski, line 504: You're a bit inconsistent on your style (see `src/rules.c`). Sometimes you make single statement if a oneliner and sometimes it's put on two lines. Personally, I prefer to have the two line variant.

src/settings.c
```diff
@@ -160,12 +160,16 @@ void settings_init(void) {
         }
 }
 
+// NOTE: This function is probably outdated and no longer relevant
+//       Since it does not even fully display rule information we
+//       may want to remove it or change it
```
zappolowski, line 165: I'm fine with removing it ... it basically just echoes the user input. The only use case could be to find typos in the configuration, but a) is this just printed at DEBUG level and b) is there already a WARNING issued, whenever an unknown key is encountered.

src/option_parser.c
bynect: Yes, I was trying to use TYPE_CUSTOM before but then I changed

src/draw.h
bynect: My idea was that either a color was valid or it was initialized with all -1. But yes, maybe an extra check for correctness is fine

src/option_parser.c
bynect: Mmm, didn't know about that. I moved the parsing code from draw.c. maybe we can add color names in another pr.

Did you think about embedding an array with all the info from that file?

src/settings.c
bynect: The main problem is that it prints only some values, so it doesn't really help that much

src/notification.c
bynect: Right. I try to minimize the use of lines, so if both the condition and statements are really short I put them in a single line (especially if there are a lot chained like here). I will try to be more consistent in the future

bynect:
I will do some fixes and merge it then 👍

zappolowski:
> I will do some fixes and merge it then 👍

I wouldn't merge as long as CI doesn't work properly. I also accidentally merged a branch yesterday - but luckily, this was just update of a pod file and thus the chance of breakage is rather low.

bynect:
Right, ci is still down 😬

zappolowski:
@bynect As I've merged the new pipeline you could rebase this branch and then hopefully we can go forward merging it.

bynect:
@zappolowski should I merge it now?

zappolowski:
Yes, just go ahead and merge it.

zappolowski (approved):

PR #1312, opened 2024-03-10
Rewrite github action

zappolowski:
Yet another try of getting the pipelines to work. This continues/replaces #1303.

Notable changes:
* update to use latest images from `dunst-project/docker-images`
* updated the images for Debian bookworm and Alpine (see the other repository) to add some more missing dependencies
* `make test-coverage` has issues running in parallel ⇒ `-j` should not be used
* don't kill matrix jobs when one fails in order to see which targets are really broken
* update actions where possible
* `misc-doxygen` doesn't need to be a matrix, if it is a `1x1` matrix

TODO:
* [x] why generate coverage data for clang, when it's discarded anyhow (just gcc is uploaded); why upload coverage data for each distro? shouldn't one suffice?
* [ ] `codecov/codecov-action@v3` is outdated, but `v4` doesn't work an all images
* [x] uploading coverage report fails on forks, this could be fixed in `codecov/codecov-action@v4` (at least the description seems to imply this)
* [x] creating coverage reports on forks could be disabled altogether as it's done anyhow when the PR is created

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,16 +11,14 @@ jobs:
   enforce-g_malloc:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-      with:
-        # Clone the whole branch, we have to fetch tags later
-        fetch-depth: 0
+      - uses: actions/checkout@v4
 
-    - name: enforce-g_malloc
-      run: "! git grep -P '(?<!g_)malloc' ':!.valgrind.suppressions' ':!.github/workflows/main.yml'"
+      - name: enforce-g_malloc
+        run: "! git grep -P '(?<!g_)malloc' ':!.valgrind.suppressions' ':!.github/workflows/main.yml'"
 
   build:
     strategy:
+      fail-fast: false
       matrix:
         CC:
           - clang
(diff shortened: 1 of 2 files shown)


bynect:
Seems good to me, however I am not sure what changes from #1303 (?)

> * why generate coverage data for clang, when it's discarded anyhow (just gcc is uploaded); why upload coverage data for each distro? shouldn't one suffice?

I second that.

zappolowski:
> however I am not sure what changes from https://github.com/dunst-project/dunst/pull/1303 (?)

* a different approach to fix [this issue](https://github.com/dunst-project/dunst/actions/runs/8132948470/job/22223961717?pr=1303#step:4:8) - instead of running as a different user, explicitly allow it using `git config` (as it's done in `actions/checkout` already)
* `alpine` image is not excluded anymore
* all tests are running - in #1303 DBus tests are deactivated

bynect:
> > however I am not sure what changes from https://github.com/dunst-project/dunst/pull/1303 (?)
> 
> * a different approach to fix [this issue](https://github.com/dunst-project/dunst/actions/runs/8132948470/job/22223961717?pr=1303#step:4:8) - instead of running as a different user, explicitly allow it using `git config` (as it's done in `actions/checkout` already)
> * `alpine` image is not excluded anymore
> * all tests are running - in #1303 DBus tests are deactivated 

Thanks for the clarification. Is something still missing or will be able to fix ci by merging this one?

zappolowski:
From my side I'm okay with merging, maybe @bebehei wants to take another look as he originally started working on this.

bebehei:
I'll have alook tomorrow.

bynect:
news?

zappolowski:
My suggestion: merge it in the current state (if @bebehei has no strong objections) and get the CI up and running again. This way we can merge some other PRs which are currently blocked by the broken CI.

Then I can iron out the missing parts.

zappolowski:
I've reworked coverage creation a bit. Now, it's just created when all tests pass, only for gcc and only for the main repository (hopefully, haven't tested it yet). I've chosen to run that step on the fedora image as this is usually quite up-to-date, not a rolling release distribution and provides lcov in the repositories.

bynect:
i don't know why but coverage is failing (yesterday was fine though)

zappolowski:
Yes, I think it's related to [this issue](https://github.com/codecov/codecov-action/issues/1359).

I'll just update to v4 ... which should work as we're using fedora based images for the coverage reports (IIRC v4 had issues on older ubuntu based images).

bynect:
> Yes, I think it's related to [this issue](https://github.com/codecov/codecov-action/issues/1359).
> 
> I'll just update to v4 ... which should work as we're using fedora based images for the coverage reports (IIRC v4 had issues on older ubuntu based images).

Thanks

Also, I wanted to make a minor version soon since a lot of commits piled up. I wrote an email to the other maintains but unfortunately I don't have your email. Do you have some prs still unfinished that you want to include?

zappolowski:
You should be able to get my address from the commits (just look at the signature).

Other than that:
* #1326 to get the pipeline running
* #1314 to fix the bug (will rebase before merging to verify pipeline again)

PR #1314, opened 2024-03-11
Do not resolve paths in get_pixbuf_from_file

### LINKED ISSUE #1307: icon file with two `'` characters in path fail to load
### Issue description

the commit adding `wordexp` in [`string_to_path`](https://github.com/dunst-project/dunst/blob/master/src/utils.c#L181), https://github.com/dunst-project/dunst/commit/e9a27c057865cae91ea0c7a45602d1ff0a2b9155 , seems to have introduced an error parsing file paths with an even number of `'`s. i've noticed because of an mpd status script which displays with icons contained in the music directory, hence lots of odd file names and directories

it seems an odd number of `'` errors out with `WRDE_SYNTAX` and returns the string without expansion... which in my case is a success, as the file path is unmodified. when an even number of `'` is present though, they're stripped away, returning an invalid file path without them

if there's some effective way to escape these with `\` i also haven't figured it out

### Installation info

- Version: both 1.10.0 (2024-02-19) and latest git

zappolowski:
The arguments should be already resolved:
* if it's either `new_icon` or `default_icon` it should be done while parsing the configuration
* if it's coming from DBus whoever sent the message should give a fully resolved path

This fixes #1307.

### DIFF
--- a/src/icon.c
+++ b/src/icon.c
@@ -185,19 +185,17 @@ static GdkPixbuf *icon_pixbuf_scale_to_size(GdkPixbuf *pixbuf, double dpi_scale,
 
 GdkPixbuf *get_pixbuf_from_file(const char *filename, int min_size, int max_size, double scale)
 {
-        char *path = string_to_path(g_strdup(filename));
         GError *error = NULL;
         gint w, h;
 
-        if (!gdk_pixbuf_get_file_info (path, &w, &h)) {
+        if (!gdk_pixbuf_get_file_info (filename, &w, &h)) {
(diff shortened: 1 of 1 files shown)


bynect:
Am I right in thinking that string_to_path is called in the option parser for the icon paths?

zappolowski:
> Am I right in thinking that string_to_path is called in the option parser for the icon paths?

Yes, both `default_icon` and `new_icon` are `TYPE_PATH` and `string_to_path` is called on them in `set_from_string`.

bynect:
👍🏻

PR #1315, opened 2024-03-12
Update dunst.5.pod

vorboyvo:
Clarify mouse_[left/middle/right]_click documentation and fix factual error

1. changed mouse_left/middle/right_click to mouse_[left/middle/right]_click to make it more obvious that it's a fill-in-the-blank
2. the defaults as listed were (slightly) wrong; in src/settings_data.h, the mouse_middle_click behaviour is in fact "do_action, close_current" and not just "do_action" as the current documentation implies. I refactored the documentation slightly to fix this error.


bynect:
Thanks 👍🏻

PR #1327, opened 2024-04-08
Just create coverage reports for pull requests

zappolowski:
No one will read them for master anyhow.

Currently, uploading fails due to a missing token, which I cannot configure as I'm lacking the needed permissions.


bynect:
It seems the condition should both hold because I am getting ci runs on my fork

PR #1331, opened 2024-04-09
Fix coverage for forks

bynect:
fix #1327

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,7 +67,7 @@ jobs:
 
   coverage:
     needs: build
-    if: github.event_name == 'pull_request'
+    if: ${{ github.repository == 'dunst-project/dunst' && github.event_name == 'pull_request' }}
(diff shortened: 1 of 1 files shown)


bynect:
My bad, I just realized that it's all the workflows and not just coverage that runs on my fork

bynect:
image

zappolowski:
The error you're encountering is due to missing git tags. As `git describe --tags` is used to create `VERSION`, this falls back to `VERSION := "1.10.0-non-git"` which then makes the tests fail.

Short: Also fork the tags from the main repository and you're good.

bynect:
> The error you're encountering is due to missing git tags. As `git describe --tags` is used to create `VERSION`, this falls back to `VERSION := "1.10.0-non-git"` which then makes the tests fail.
> 
> Short: Also fork the tags from the main repository and you're good.

But before this didn't cause any issue

PR #1335, opened 2024-04-14
Re-enable coverage for all pushes

zappolowski:
There should be a valid token now and thus the restriction for pull requests should not be needed anymore.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,7 +67,6 @@ jobs:
 
   coverage:
     needs: build
-    if: github.event_name == 'pull_request'
     env:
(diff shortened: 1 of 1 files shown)


bynect:
Shouldn't it be repo == dunst-project/dunst like before? Or was it from another part of the workflow?

zappolowski:
This shouldn't be needed anymore, as v4 of the action should already disable uploading of the report on forks (which wasn't the case before).

bynect:
> This shouldn't be needed anymore, as v4 of the action should already disable uploading of the report on forks (which wasn't the case before).

Ok then go ahead.

Issue #1316, opened 2024-03-13
Wayland: text is garbled on non-scalled monitors


matejdro:
### Issue description

When using dunst with Sway and using multiple monitors, where at least one monitor is using scaling, dunst's text will be garbled on non-scaled monitors:

![image](https://github.com/dunst-project/dunst/assets/507922/3571c616-856a-4617-9942-1e5f1ab01f30)

On scaled monitors it appears fine:

![image](https://github.com/dunst-project/dunst/assets/507922/30df2c98-8df0-43e7-ac25-a0403862c4b7)

### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon 1.10.0 (2024-02-19)`
- Install type: `package`
- Window manager / Desktop environment: `sway version 1.9`
- Distro: `Arch`

<details>
<summary>Minimal dunstrc</summary>
default + `follow` option enabled to allow displaying on multiple monitors
</details>

Setting  `force_xwayland` to `true` fixes the issue, but it also breaks `follow`, which makes it an useless workaround.

bynect:
did you enable `per_monitor_dpi`?

matejdro:
Yes, with no effect. From what I can see, this settings is X11-only.

bynect:
> Yes, with no effect. From what I can see, this settings is X11-only.

I am checking the code and it is indeed an x11 only option.

But does the error occur only when the scaled screen is connected?

matejdro:
Yes, only when the scaled screen is connected. Otherwise, it works fine.

bynect:
are you able to compile dunst ? I can add some logging in the wl_get_scale function. Unfortunately I can't test it myself because I don't have such wayland environment

bynect:
I have made that small modification in my fork (https://github.com/bynect/dunst/tree/log-scale). If you could try it maybe we can get a better understanding of what's going on

matejdro:
Sorry this took a while.

Here is log from notification on non-scaled monitor:

```
WARNING: No icon found in path: 'dialog-information'
DEBUG: [         wake_up:0067] Waking up
DEBUG: [             run:0077] RUN, reason 1
INFO: Idle status queried: 0
DEBUG: [calculate_dimensions:0306] Notification dimensions 300x91
DEBUG: [            draw:0982] Window dimensions 309x97
DEBUG: [  render_content:0748] Layout height 75
DEBUG: [wl_display_surface:0800] Buffer size (scaled) 618x194
[...]

PR #1337, opened 2024-04-16
Refactor draw code to use `cairo_surface_set_device_scale`

bynect:
This fixes something that has been bugging me for a while.
Basically now dunst is scaling itself when drawing every single shape and text (with cairo and pango).
With this patch we render everything at a uniform scaling factor of 1 and scale everything afterwards using cairo_surface_set_device_scale.

This should have some advantages (apart from conciseness), like making scaling on wayland possible (like I discussed with @alebastr under #1316)

NOTE: Everything works on Xorg, however I can't test it on Wayland and my display is not Hi-Dpi, so I am not sure if the pango dpi scaling is being applied correctly. If someone can test it it would be great 👍🏻


bynect:
I think that all the get_icon_width and get_\*scale\*icon function aren't needed anymore

bynect:
I have no clue about why the test are failing 😢

bynect:
I am trying to debug the use of uninitialized variables but even with libAsan I can't find it

src/wayland/wl_seat.c
```diff
@@ -100,7 +103,8 @@ static void pointer_handle_button(void *data, struct wl_pointer *wl_pointer,
                 uint32_t button_state) {
         struct dunst_seat *seat = data;
 
-        input_handle_click(button, button_state, seat->pointer.x, seat->pointer.y);
+        double scale = wl_get_scale();
+        input_handle_click(button, button_state, seat->pointer.x/scale, seat->pointer.y/scale);
```
bynect, line 107: before wayland didn't scale the user input

bynect:
Finally figured it out. Basically I forgot that the test didn't call draw_setup. Since I moved the initialization of the PangoContext in draw_setup it was undefined and thus buggy, but only in tests.

bynect:
Everything things to work. I think it is ready for merge.

Things we may want to consider
* remove the icon scaling code (since we don't use it anymore). Alternatively we can render icons separately over the cairo surface to not downscale->upscale them. But this can be done in another pr.
* integrate the new scaling method with wayland for per-monitor-dpi

bynect:
@alebastr Where do you think should we place scale code so that the wayland backend can correctly get the scale needed? (to solve #1316)

Note that the pr is still not quite ready)

bynect:
> > If someone can test it it would be great 👍🏻
> 
> I can test it on Wayland if desired, just give me instructions on what to look for

it would be great if you could check if the input works correctly (mouse clicks for closing etc) and if you notice any graphical discrepancy let me know.
The only difference I noticed is that text is a bit less sharp but maybe it's just me?

matejdro:
Only issue I have noticed is that, with scaling enabled, dunst window has bigger horizontal offset from the edge of the screen and the window is smaller. All below screenshots extend to the top right corner of the monitor.

# This branch

Scaled monitor:

image

Interestingly enough, after I take a screenshot and Satty opens fullscreen, dunst window will become taller (even before Satty makes second notification):

image

Non-scaled monitor:

image

After I turn off scaling on the other monitor, text is clearer for the non-scaled monitor and position bug disappears:

image

# Dunst release

Scaled monitor:

image

Non-scaled monitor:

image

After I turn off scaling on the other monitor, text is clearer:

image


Mouse clicks seem to work fine. I also don't see any discrepancies regarding text quality between this branch and latest dunst release (but you can judge yourself from screenshots).

bynect:
@matejdro so is the resolution still wrong even with this branch?

Also I can look into the offset problem, probably something was not scaled appropriately

matejdro:
Above screenshots are from Hyprland.

On Sway, text look as bad as it is:

image

But interestingly enough, on Hyprland, issue appears to be different. As you can see on the screenshots from the previous posts, text is still worse when scaling is enabled, but it's not as jagged as on Sway. Maybe Hyprland applies some sort of font smoothing by default, not sure.

bynect:
Thanks very much for the feedback. I will investigate this further and make some changes 👍

bynect:
@matejdro just to let you know. the problem with wayland scaling is that we check for the screen before drawing but on wayland the protocol does not allow that with follow mode. so further changes are needed to this pr for it to work on wayland. maybe it should be done in a seperate pr

bynect:
I am not sure if this needs more work. I would probably do the scale detection refactor in another pr since it requires wayland backend changes.

I am not sure if the graphical effect of `cairo_surface_set_device_scale` is as good as before due to antialiasing...

fwsmit:
Can you show what is the visual difference between the two? I don't really understand why the old method would not work on Wayland. Not that using pango device scale is bad, but I'm not sure this should solve a scaling issue on Wayland. 
It seems like the problem on Wayland is related to not detecting the scale factor correctly which can be solved by changing the Wayland backend as you mentioned. I thought of some hacky way a while ago where you first draw the surface and then find on which display the surface was drawn and get the right display scale factor. After you might need to redraw. But there might be a better way possible now with protocol updates.

bynect:
> Can you show what is the visual difference between the two? I don't really understand why the old method would not work on Wayland. Not that using pango device scale is bad, but I'm not sure this should solve a scaling issue on Wayland. It seems like the problem on Wayland is related to not detecting the scale factor correctly which can be solved by changing the Wayland backend as you mentioned. I thought of some hacky way a while ago where you first draw the surface and then find on which display the surface was drawn and get the right display scale factor. After you might need to redraw. But there might be a better way possible now with protocol updates.

the problem is the layer protocol like you mentioned. As far I understand it doesn't allow you to know beforehand where you are drawing. The possible solution using this device scale is that you first render at 1x on the surface image (like we already do) and then you scale this surface afterwards.
This in general should be better than scaling everything but somehow the result I get is worse.
Probably skill issue on my part.
Anyway I don't see an easy way to deal with the wayland issue without thorough refactoring

fwsmit:
> the problem is the layer protocol like you mentioned. As far I understand it doesn't allow you to know beforehand where you are drawing. 

If you set the output to some value, then you know the surface will be drawn to this output. The thing is that if follow=mouse or follow=keyboard is set, we set the output to NULL, indicating to the compositor it can decide where to place the notification. This means you don't know where the notification will end up until after it's shown. When a single output it set in the configuration, it is always possible to determine beforehand on which output the notification will end up and there should be no scaling issues.

So the only case where there should be scaling issues is when `follow=keyboard/mouse` AND there are multiple outputs with different scales. We could mention this shortcoming in the documentation or work around it in a hacky way.

This PR will not solve the mentioned issue, it only makes the code a bit cleaner.

fwsmit:
if the offset problem, as mentioned in the testing above is fixed, I think this can be merged.

b0ch3nski:
> scaling issues is when `follow=keyboard/mouse` AND there are multiple outputs with different scales.
This is exactly what is happening in my case :slightly_frowning_face:

PR #1342, opened 2024-04-20
Implement dynamic height and change vertical padding

bynect:
Implement dynamic height for notification, namely a way to define a minimum height.

I also changed the behavior of vertical padding to work even when there is no icon.

----------

Summary:
* implement dynamic height (min/max)
* generalize implementation of vertical alignment to work without icon and with hide_text
* add functional-test for the changes
* add a test suite for min height
* refactored functional-tests and added them to the makefile
* added height to the docs and fixed some indentation on them


bynect:
Everything seems to work. Maybe I will add some functional tests

bynect:
I noticed that there are some differences from the old behavior so it's not ready yet

bynect:
Everything is working fine now. I also added test and functional tests.

For a quick check on what this implements use `./test vertical_align`

I also noticed some indentation error in the docs and fixed (sorry for not making another pr :o)

bynect:
I think it's best to merge this before #1337

fwsmit:
Sorry, I cannot test at the moment. Maybe someone else can do it?

zappolowski (commented):
Code seems good, but the notifications look different than current master with too much padding on the top and too little at the bottom.

Current master:
old

This branch:
new

docs/dunst.5.pod
```diff
@@ -106,7 +108,16 @@ screens exceed (e.g. 10000).
...
 =item B<height>
 
-The maximum height of a single notification.
+The height of each notification in pixels. This can be a single number to
+specify a constant height or two numbers for the minimum and maximum width. The
+notification will expand from the minimum height as neccesary.
+
+Examples:
+
+    height = 300      # constant height of 300
```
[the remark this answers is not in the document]
bynect: before the height was dynamic by default, but now we can set a minimum. what's the best way to communicate this? maybe a log info in the check settings?

bynect:
> Code seems good, but the notifications look different than current master with too much padding on the top and too little at the bottom.
> 
> Current master: old
> 
> This branch: new

I'm not sure this is the intended behavior (except if vertical alignment is bottom I guess) so I will check again. did you test with the default dunstrc?

docs/dunst.5.pod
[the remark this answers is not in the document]
zappolowski: I'm not really sure about the best approach here. Maybe an entry in the CHANGELOG marking it as breaking change suffices?

bynect:
Now it should work as intended. 

The only corner case that I am not sure how to handle is when the max height is too small to fit everything

initially it was
img-1714600820

then I made it overlap but not go outside the border
img-1714600854

but this is just an hack. the real problem is that there is not enough room in this case. I would leave it as it this...

bynect:
Ok I should have fixed the problem above

img-1714602201

test/functional-tests/test.sh
```diff
@@ -11,29 +14,28 @@ function tmp_dunstrc {
...
 }
 
 function show_age {
     echo "###################################"
     echo "show age"
     echo "###################################"
-    killall dunst
-    ../../dunst -config dunstrc.show_age &
-    ../../dunstify -a "dunst tester"  -u c "Show Age" "These should print their age after 2 seconds"
+    start_dunst dunstrc.show_age &
```
zappolowski, line 37: Missed it last time ...

When you're putting this command in background you're basically nullifying the sleep in that function, which could potentially cause issues when start up is slowed down.

test/functional-tests/test.sh
```diff
@@ -43,23 +45,22 @@ function run_script {
     echo "run script"
     echo "###################################"
     killall dunst
-    PATH=".:$PATH" ../../dunst -config dunstrc.run_script &
-    ../../dunstify -a "dunst tester" -u c \
+    PATH=".:$PATH" $DUNST -config dunstrc.run_script &
```
zappolowski, line 48: Why didn't you use `start_dunst` here?

bynect: Because PATH was exported. I don't know if it would work with the function. I will try

test/functional-tests/test.sh
bynect: I thought this was the previous behavior. I'll check again. 

Anyhow is the graphical part working now?

bynect: It was just an oversight

test/functional-tests/test.sh
bynect: I refactored a bit so this can be handled nicely

fwsmit:
How would you compare the output of the functional tests? Pixel for pixel comparison of the images might not work so well, because the font rendering might change slightly with different versions of pango or a font

bynect:
> How would you compare the output of the functional tests? Pixel for pixel comparison of the images might not work so well, because the font rendering might change slightly with different versions of pango or a font

I didn't think about that. Maybe with specific pango and cairo versions

bynect:
since this seems to work well I'll merge

PR #1348, opened 2024-05-01
[RFC] DBus add NotificationListDisplayed

### LINKED ISSUE #1242: Feature request: dunstctl close notiication based on conditions
It would be handy to be able to se dunstctly to only close notifications pertaining a single app, or all the notifications containing a specified text/pattern.

also apparently the history-rm command is not documented in the docs folder of the source code.

Thanks!

zappolowski:
A possible implementation to fix #1242.

This way one could e.g. use (fish syntax):
```fish
for id in (busctl --user call org.freedesktop.Notifications /org/freedesktop/Notifications org.dunstproject.cmd0 NotificationListDisplayed -j | jq -r '.data[][] | select(.appname.data == "Spotify") | .id.data')
    busctl --user call org.freedesktop.Notifications /org/freedesktop/Notifications org.freedesktop.Notifications CloseNotification u $id
end
```

To close all currently displayed notifications generated by Spotify.

## TODO (if accepted)

- [ ] add tests
- [ ] add subcommand to dunstctl
- [ ] update completions if needed

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -83,6 +83,9 @@ static const char *introspection_xml =
     "        <method name=\"NotificationClearHistory\"/>"
     "        <method name=\"NotificationCloseLast\" />"
     "        <method name=\"NotificationCloseAll\"  />"
+    "        <method name=\"NotificationListDisplayed\">"
+    "            <arg direction=\"out\" name=\"notifications\"   type=\"aa{sv}\"/>"
+    "        </method>"
     "        <method name=\"NotificationListHistory\">"
     "            <arg direction=\"out\" name=\"notifications\"   type=\"aa{sv}\"/>"
     "        </method>"
@@ -187,6 +190,7 @@ DBUS_METHOD(dunst_NotificationAction);
 DBUS_METHOD(dunst_NotificationClearHistory);
 DBUS_METHOD(dunst_NotificationCloseAll);
 DBUS_METHOD(dunst_NotificationCloseLast);
+DBUS_METHOD(dunst_NotificationListDisplayed);
 DBUS_METHOD(dunst_NotificationListHistory);
 DBUS_METHOD(dunst_NotificationPopHistory);
 DBUS_METHOD(dunst_NotificationRemoveFromHistory);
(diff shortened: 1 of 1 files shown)


bynect:
I like the idea.

I was wondering if there was a nice way to call closeNotification from dunstctl or the like. Maybe we can add an optional argument to `dunstctl close`?

bynect:
I suggest to implement a more generic `NotificationList` that takes a bool parameter for displayed. This way we can query all the current notifications (otherwise the one neither displayed nor in history would be inaccessible). 
this would allow for example autocompletion of `dunstctl close` with the id (and surely other nice things).

zappolowski:
I don't think, that `NotificationList bool` would make good API as the meaning of the parameter is not immediately clear. Given that `NotificationListHistory` already exists and this one adds `NotificationListDisplayed` adding another `NotificationListQueued` or `NotificationListWaiting` seems preferable from my side.

Maybe for the future one could think about implementing `NotificationList [waiting/queued|displayed|history]`, but at the moment `NotificationListHistory` is already published and thus it might be used. Changing it now would be a breaking change.

bynect:
> I don't think, that NotificationList bool would make good API as the meaning of the parameter is not immediately clear.

Well the parameter name can be specified in the xml. Maybe for a clearer name something like `NotificationListCurrent`?

But creating another method is alright

zappolowski:
> Maybe for a clearer name something like `NotificationListCurrent` to differentiate from the history one?

TBH, I don't get this proposal. `NotificationListDisplayed` states that it's listing notifications displayed. For `NotificationListCurrent` I would expect the union of displayed and waiting notifications being returned.

I also don't understand why there would be a need to differentiate from `NotificationListHistory`. We then would end up with `NotificationListDisplayed`, `NotificationListWaiting`, and `NotificationListHistory`, which form a kind of self-describing API stating what they're about to return.

> But creating another method is alright. So we would just need to combine the two where needed

For both use cases described in #1242 and #1346 just the displayed notifications are of interest and thus `NotificationListWaiting` (to use the term used in the code already) would not be strictly needed (though, I'd add it anyhow as it's just a couple of lines).

bynect:
> But creating another method is alright

It was just an idea. You can go ahead with displayed and waiting as you said

bynect:
is this in a mergeable state @zappolowski ?

PR #1350, opened 2024-05-02
[RFC] config reload mechanism

bynect:
I wanted to add a way to reload the config files. part of the code is taken from #968.

In respect to #968 I added the ability to pass a list of config files to use to the dbus method. If none are passed it will use the old config list.

I noticed that the cmdline parser ignores multiple `-conf` options (without any acknowledgment of the fact). So I changed the behavior to accept a list of files.

-----

# Summary

* Add dbus method `ConfigReload`
* Refactor cmdline parsing to account for multiple `-conf` values
* Cleanup wayland output on deinit
* Update tests and documentation
* Add `dunstctl reload` and completions
* Keep original notification state for reload


bynect:
mostly done. I'll do some more checks and maybe add a functional test

main.c
```diff
@@ -1,6 +1,6 @@
 #include "src/dunst.h"
 
-int main(int argc, char *argv[])
+int main(int argc, const char *argv[])
```
fwsmit, line 3: Why only make argv const? I reckon argc also could be const

src/dunst.c
```diff
@@ -229,10 +255,16 @@ int dunst_main(int argc, char *argv[])
...
+        cmdline_usage_append("-conf/-config", "string", "Path to configuration file");
+
+        int start = 1, count = 1;
+        while (cmdline_get_string_offset("-conf/-config", NULL, start, &start))
+                count++;
+
+        config_paths = g_malloc0(sizeof(char *) * count);
+        start = 1, count = 0;
+
+        while ((config_paths[count++] = cmdline_get_string_offset("-conf/-config", NULL, start, &start)));
```
fwsmit, line 267: This is written a bit too compact, making it hard to read. Could you expand this to multiple lines?

src/queues.h
```diff
@@ -179,6 +179,10 @@ gint64 queues_get_next_datachange(gint64 time);
  */
 struct notification* queues_get_by_id(int id);
 
+/**
+ * Reapply all rules to the queue (used when reloading configs)
```
fwsmit, line 183: What if a property of a notification is previously changed by a rule? Then reapplying the rule will apply to the changed notification, not the original. This might result in weird behavior. To make it work correctly you have to store the original notification's contents

src/notification.c
```diff
@@ -317,6 +317,8 @@ void notification_unref(struct notification *n)
         notification_private_free(n->priv);
 
         if (n->script_count > 0) {
+                for (int i = 0; i < n->script_count; i++)
+                        g_free(n->scripts[i]);
```
fwsmit, line 321: I feel like there is a glib function to do this

fwsmit (changes requested):

fwsmit:
Thanks for picking this up! I never got it to work properly when I tried it. But when the code is structured well, it should not be too hard to implement

src/dunst.c
```diff
@@ -229,10 +255,16 @@ int dunst_main(int argc, char *argv[])
...
-        cmdline_config_path =
-            cmdline_get_string("-conf/-config", NULL,
-                               "Path to configuration file");
+        cmdline_usage_append("-conf/-config", "string", "Path to configuration file");
+
+        int start = 1, count = 1;
+        while (cmdline_get_string_offset("-conf/-config", NULL, start, &start))
+                count++;
+
+        config_paths = g_malloc0(sizeof(char *) * count);
```
fwsmit, line 264: Is this freed properly somewhere?

main.c
bynect: It was due to const correctness somewhere along the chain. Also a int paramater doesn't really matter, but a pointer can be modified otherwise (and argv should not be modified)

src/dunst.c
bynect: Sure

src/queues.h
bynect: That's true, but for now I think this is the correct behavior. I also added it to the documentation. Maybe we could add an option `keep_original` or something to maintain a copy of the original notification if the user so wants. Doing so would add to the memory footprint

src/notification.c
bynect: I didn't find it

bynect: I didn't find it

src/dunst.c
bynect: In the teardown function. And when it is overridden in reload

src/notification.c
bynect: modified to use gstrfreev

bynect:
> Thanks for picking this up! I never got it to work properly when I tried it. But when the code is structured well, it should not be too hard to implement

I made some updates. The only thing I am a bit unsure is the reapply of rules

src/queues.h
fwsmit: I would definitely not make an option for this. The behavior will probably be acceptable in most cases, but the more correct thing to do is store the original notification and apply the rule to that.

Say a rule matches a notification with a timeout of 10 seconds with match_dbus_timeout and sets the timeout to 20 seconds and the color to yellow. Let's say you change your config to change the color to green instead and reload. You expect the notification to become green, but it stays yellow, because it's timeout has been changed to 20, making it not match the rule anymore.

Of course this is not the end of the world, but it is not the correct behavior. I don't see many rules that would make it a very common occurrence, so it would be fine by me to keep it like this. But it would be good to document and make an issue about it, so it is not forgotten.

bynect: @fwsmit I would prefer doing that in another pr if possible since this is already quite big. How are the current changes looking?

bynect:
Basically I added a rule struct inside the notification that is allocated and filled when we try to apply a rule to change a value. Then when we reload we reapply that rule to "revert" the original state.
it should work but I haven't tested much (it is kind of involved)

@fwsmit does this solve the problem you said in the comments?

bynect:
arch ci is not working ...

src/queues.c
```diff
@@ -617,6 +617,22 @@ struct notification* queues_get_by_id(int id)
...
 
+void queues_reapply_all_rules(void)
+{
+        GQueue *recqueues[] = { displayed, waiting, history };
+        for (int i = 0; i < sizeof(recqueues)/sizeof(GQueue*); i++) {
+                for (GList *iter = g_queue_peek_head_link(recqueues[i]); iter;
+                     iter = iter->next) {
+                        struct notification *cur = iter->data;
+                        if (cur->original) {
+                                //rule_print(cur->original);
```
zappolowski, line 628: Left over from debugging?

src/rules.c
```diff
@@ -14,66 +14,106 @@
 
 GSList *rules = NULL;
 
+#define APPLY(nprop, rprop, defval) \
```
zappolowski, line 17: Personally, I would scope this macro just for `rule_apply`. This way, it's also more obvious what `r`, `n` and `save`.

You could also move the initial condition to the macro as it always seems to be
```c
if (r->rprop != defval)
```

Do we have any cases where `nprop` and `rprop` differ?

src/settings.c
```diff
@@ -289,27 +268,41 @@ static void process_conf_file(const gpointer conf_fname, gpointer n_success) {
...
         if (0 == n_loaded_confs)
-                LOG_I("No configuration file found, using defaults");
+                LOG_M("No configuration file found, using defaults");
 
-        for (GSList *iter = rules; iter; iter = iter->next) {
-                struct rule *r = iter->data;
-                print_rule(r);
-        }
+        // Maybe add an option to display the rules
+        //for (GSList *iter = rules; iter; iter = iter->next) {
```
zappolowski, line 301: This code here should be deleted as the referenced function doesn't exist anymore and thus it will be more confusing than helping.

src/wayland/wl.c
```diff
@@ -332,10 +332,10 @@ void wl_deinit(void) {
         // could have been aborted half way through, or the compositor doesn't
         // support some of these features.
         if (ctx.layer_surface != NULL) {
-                zwlr_layer_surface_v1_destroy(ctx.layer_surface);
+                g_clear_pointer(&ctx.layer_surface, zwlr_layer_surface_v1_destroy);
```
zappolowski, line 335: Why was this needed?

src/queues.c
bynect: I left it half intentionally because we might want to give some feedback about what is saved during a reload

src/settings.c
bynect: Yes, I left it only because we have no way to print rules currently. maybe we should add an option to do it?

src/wayland/wl.c
bynect: this was taken from the original commit made by friso

src/rules.c
bynect: > Do we have any cases where `nprop` and `rprop` differ?

unfortunately yes. But I made a secondary macro to fix this

docs/dunstctl.pod
```diff
@@ -90,6 +90,13 @@ to temporarily activate or deactivate specific rules.
 
 Exports all currently configured rules (optionally JSON formatted).
 
+=item B<reload> [dunstrc ...]
+
+Reload the settings of the running dunst instance. You can optionally specify
+which configuration files to use. Otherwise, the previous one will be used.
```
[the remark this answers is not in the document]
bynect: drop ins are searched only if nothing is specified. I suppose that if you start dunst without -config and reload it will read drop ins on reload. Otherwise if you specify a specific file drop ins are not loaded.



> If you don't specify a config, I think the default config should be used, not the previous one.

By default you mean the dunstrc in the xdg dirs?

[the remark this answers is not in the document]
bynect: sure, I just need to save only the original configs and not every time we reload.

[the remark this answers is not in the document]
bynect: for drop-ins I don't see an easy way to do it because one of the use cases is  testing specific dunstrc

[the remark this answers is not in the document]
fwsmit: You just need to store the original command line arguments, not the config contents.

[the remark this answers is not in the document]
bynect: > You just need to store the original command line arguments, not the config contents.

yes, I already did something similar

bynect:
@fwsmit @zappolowski I finished. Also updated the docs. Can I merge this? wanted to release a version in this month what do you think?

fwsmit:
Looks good to me!

fwsmit (approved):

bynect:
I'll merge then

Fxzzi:
seems like running `dunstctl reload` whilst a notification on screen causes dunst to stop working.

bynect:
> seems like running `dunstctl reload` whilst a notification on screen causes dunst to stop working.

How so? It should just reapply the config to the notification

Dreuzz:
> dunstctl reload /home/dreuzzz/.cache/wal/dunst
Error org.freedesktop.DBus.Error.UnknownMethod: No such method “ConfigReload”
Failed to communicate with dunst, is it running? Or maybe the version is outdated. You can try 'dunstctl debug' as a next debugging step.

bynect:
> > dunstctl reload /home/dreuzzz/.cache/wal/dunst
> 
> Error org.freedesktop.DBus.Error.UnknownMethod: No such method “ConfigReload” Failed to communicate with dunst, is it running? Or maybe the version is outdated. You can try 'dunstctl debug' as a next debugging step.

this means that the running version of dunst is not updated and doesn't have the reload patch.

Dreuzz:
> dunst -v
Dunst - A customizable and lightweight notification-daemon v1.11.0-76-gc1cc6d1
Compiled on 2024-09-29 with the following options:
X11 support: enabled
Wayland support: enabled
SYSCONFDIR set to: /usr/local/etc/xdg
Compiler flags: -g -std=gnu11 -pedantic -Wall -Wno-overlength-strings -Os -pthread -MMD -MP
Linker flags: -lm -lrt -lgio-2.0 -lgdk_pixbuf-2.0 -lglib-2.0 -lpangocairo-1.0 -lpango-1.0 -lgobject-2.0 -lharfbuzz -lcairo -lwayland-client -lwayland-cursor -lX11 -lXinerama -lXext -lXrandr -lXss

bynect:
> > dunst -v
> 
> Dunst - A customizable and lightweight notification-daemon v1.11.0-76-gc1cc6d1 Compiled on 2024-09-29 with the following options: X11 support: enabled Wayland support: enabled SYSCONFDIR set to: /usr/local/etc/xdg Compiler flags: -g -std=gnu11 -pedantic -Wall -Wno-overlength-strings -Os -pthread -MMD -MP Linker flags: -lm -lrt -lgio-2.0 -lgdk_pixbuf-2.0 -lglib-2.0 -lpangocairo-1.0 -lpango-1.0 -lgobject-2.0 -lharfbuzz -lcairo -lwayland-client -lwayland-cursor -lX11 -lXinerama -lXext -lXrandr -lXss

you are using master branch right? You have to make sure that the program actually running is this one. sometimes the build systems/packages etc start a different binary

B1ack3ye:
On my Arch installation, `dunstctl reload` is not working:

```
[b1ack3ye@Black-PC ~]$ dunst -v
Dunst - A customizable and lightweight notification-daemon 1.11.0 (2024-04-15)
[b1ack3ye@Black-PC ~]$ dunstctl reload
dunstctl: unrecognized command 'reload'. Please consult the usage.
[b1ack3ye@Black-PC ~]$ dunstify "Test"
[b1ack3ye@Black-PC ~]$ dunstctl reload
dunstctl: unrecognized command 'reload'. Please consult the usage.
```

zappolowski:
@B1ack3ye this feature is not yet released. If you want to it you either have to build it yourself or use `dunst-git` from AUR.

Fxzzi:
> > seems like running `dunstctl reload` whilst a notification on screen causes dunst to stop working.
> 
> How so? It should just reapply the config to the notification

yeah, which is what I'm thinking, however afterwards no notifications show up. For now I just `pkill dunst` and it starts itself up again next time a notification is sent.

bynect:
> > > seems like running `dunstctl reload` whilst a notification on screen causes dunst to stop working.
> > 
> > 
> > How so? It should just reapply the config to the notification
> 
> yeah, which is what I'm thinking, however afterwards no notifications show up. For now I just `pkill dunst` and it starts itself up again next time a notification is sent.

do you mean when reloading all the notifications are closed? or when new notifications come nothing show up? it would be very helpful to provide a debug log (dunst -verbosity debug)

Fxzzi:
> do you mean when reloading all the notifications are closed? or when new notifications come nothing show up?

Both. current notifications close, and no new notifications show up.

> it would be very helpful to provide a debug log (dunst -verbosity debug)

sure, here u go. https://paste.rs/xSpVi.txt

in this log, i simply launched dunst with verbosity, triggered my volume notification, and then did a `dunstctl reload`

bynect:
> > do you mean when reloading all the notifications are closed? or when new notifications come nothing show up?
> 
> Both. current notifications close, and no new notifications show up.
> 
> > it would be very helpful to provide a debug log (dunst -verbosity debug)
> 
> sure, here u go. https://paste.rs/xSpVi.txt
> 
> in this log, i simply launched dunst with verbosity, triggered my volume notification, and then did a `dunstctl reload`

Are you using wayland right? Since I can only test on xorg maybe this is a wayland only bug. could you try using dunst in Xwayland mode to see if this works there?

Fxzzi:
> > > do you mean when reloading all the notifications are closed? or when new notifications come nothing show up?
> > 
> > 
> > Both. current notifications close, and no new notifications show up.
> > > it would be very helpful to provide a debug log (dunst -verbosity debug)
> > 
> > 
> > sure, here u go. https://paste.rs/xSpVi.txt
> > in this log, i simply launched dunst with verbosity, triggered my volume notification, and then did a `dunstctl reload`
> 
> Are you using wayland right? Since I can only test on xorg maybe this is a wayland only bug. could you try using dunst in Xwayland mode to see if this works there?

yep, I'm running on wayland.

Fxzzi:
Finally had some time to test this (I also forgot about this issue lol)

`force_xwayland = true` in config does allow reload to work, however xwayland isn't useable, the notification is always misplaced and also its just bad practice.

bynect:
> Finally had some time to test this (I also forgot about this issue lol)
> 
> `force_xwayland = true` in config does allow reload to work, however xwayland isn't useable, the notification is always misplaced and also its just bad practice.

did you try with master? I discovered some nasty mem errors that may be related

Fxzzi:
> did you try with master? I discovered some nasty mem errors that may be related

i did indeed see your pr and tested to check if it made any change. sadly it hasn't made any change.

bynect:
> > did you try with master? I discovered some nasty mem errors that may be related
> 
> i did indeed see your pr and tested to check if it made any change. sadly it hasn't made any change.

I see. this needs more testing but for now I have to release anyway with this bug because of #1411

Since this pr is now closed could you please create an issue with the error description so I can keep track of it? thanks

PR #1351, opened 2024-05-02
Add an optional id to `dunstctl close`

bynect:


zappolowski (commented):
Looks good. It would be nice to adjust the completions as well (at least the summary strings) for zsh and fish as they're misleading otherwise.

bynect:
> Looks good. It would be nice to adjust the completions as well (at least the summary strings) for zsh and fish as they're misleading otherwise.

right, forgot about that 👍🏻

bynect:
@zappolowski I fixed some typos and updated the completions. Unfortunately we can't autocomplete the id (see my suggestion in #1348).

completions/_dunstctl.zshcomp
zappolowski, line 1: This comment applies to both completions: the descriptions for each subcommand were taken verbatim from the output of `dunstctl help` and I think it should stay this way.

Thus I suggest to adjust the output of help:
* `is-paused`: `>` → `greater than`
* `history-pop`: remove trailing period
* `help`: just `Show help` should be fine

bynect: Sure. I noticed that the help message itself had some typos.

bynect:
done.

unrelated question, should we add a completion for dunstify later?

zappolowski (approved):

zappolowski:
> unrelated question, should we add a completion for dunstify later?

Yes, this would make sense. I just added it for fish as that's what I'm using.

PR #1353, opened 2024-05-08
Update config man page

d2718nis:
Remove the message about shortcuts deprecation, as they were reintroduced to the configuration as part of the global section in commit 17822228


d2718nis:
Hmm, I think my second commit is misleading: there are multiple `icon` related attributes and that's what was meant. But then again I'm able to modify the `format` attribute per urgency section as well, so something is not quite right?

zappolowski (approved):

Issue #1364, opened 2024-06-25
WARNING: no icon found in path: 'dialog-information' is making journalctl hard to use.


misterhackerman:
### Issue description

<details>
<summary>journalctl log</summary>
Jun 25 11:21:21 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:31 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:36 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:37 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:37 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:37 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:37 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:37 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:37 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:37 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:38 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:38 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:38 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:38 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:38 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:38 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
Jun 25 11:21:38 M4CH1N3 dunst[31192]: WARNING: No icon found in path: 'dialog-information'<br>
[...]

Issue #1358, opened 2024-05-28
How to disable icons in `dunstify`


eeeXun:
### Issue description

I set `enable_recursive_icon_lookup` to true in my `dunstrc`. After version `1.11.0`, the icon is alway displayed in `dunstify` command. Like

```
dunstify "Hello"
```
![image](https://github.com/dunst-project/dunst/assets/58657914/0ee825b0-dbce-46e5-8ff7-61532c0c2d07)

In some cases, I don't want to show the icon in `dunstify` command. I've checked the `dunstify --help` menu, there doesn't seem to be a flag, like `--no-icon`, to disable the icon.

### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon 1.11.0 (2024-04-15)
`
- Install type: `pacman`
- Window manager / Desktop environment: `dwm`
- Distro: `ArchLinux`

<details>
<summary>Minimal dunstrc</summary>

```ini
[global]
    enable_recursive_icon_lookup = true
```
</details>

9mdv:
Changing `enable_recursive_icon_lookup` to `false` disabled the icon for me.

bynect:
At the moment the icon have some default value (for historical reason I guess)
(The code responsible for this is at https://github.com/dunst-project/dunst/blob/20033b86595c3967a20a476065443e515e39ac63/src/settings_data.h#L368. )

Can you try doing 
```
[urgency_low]
icon =
default_icon = 
```

Otherwise I you just don't want icon you can set `icon_position=off`

eeeXun:
Oh, I don't know that's the default behavior.

And I found out that my problem was not caused by dunst `1.11.0`, but by updating `adwaita-icon-theme` from `46.0` to `46.2`. (Maybe the version before `46.2` is missing some icons. So I haven't encountered this problem before.)

> there are two ways I can think of. One is to create a rule for a custom category to disable icons and using that category from dunstify
> 
> the other is passing an invalid icon path to `dunstify -i GIBRISH`

Yes, these could be the workaround for me. But could I request a feature of `--no-icon` flag or something like that in `dunstify` command? If this is not planned, feel free to close this issue.

NicolasWebDev:
I have the same issue you're mentioning @eeeXun , since a system upgrade yesterday a default icon began to appear in my notifications. It must be a theme package that got upgraded because dunst itself was not upgraded.

[...]

PR #1365, opened 2024-07-01
Remove default icons and refactor icon-lookup

bynect:
As requested by #1364 and #1358. I also refactored the search function to provide more informative logging


bynect:
some  cleanup

test/icon-lookup.c
```diff
@@ -55,20 +46,6 @@ TEST test_find_icon(void)
         PASS();
 }
 
-TEST test_find_path(void)
```
fwsmit, line 58: Why did you remove these tests? Did anything in the search behavior change other than the default icons? If so that should also be documented and discussed

bynect: find_icon_path and get_path_from_icon had basically half of the code duplicate. so I removed some code from the first since in the first place it is used only by get_path_from_icon

bynect: the absolute path and uri handling were the exact same

fwsmit: Ah okay. Makes sense

bynect:
@fwsmit does it looks good? can I merge?

fwsmit:
> @fwsmit does it looks good? can I merge?

Yeah, looks good. You can go ahead and merge

fwsmit (approved):

bynect:
> > @fwsmit does it looks good? can I merge?
> 
> Yeah, looks good. You can go ahead and merge

perfect 👍🏻

PR #1370, opened 2024-08-16
Add color gradients

### LINKED ISSUE #1273: Feature: Color gradient

bynect:
Supersedes #1286


bynect:
I added the gradient only for the `highlight` at the moment, but it should be easy to add to the others. the cairo pattern is cached at the start. for the rest it is basically the same as #1286 but with a custom gradient type instead of a string list

bynect:
this is what it looks like
img-1723929625

bynect:
@zappolowski @fwsmit should be ready. I already closed the old pr

bynect:
ok fixed 🤣

bynect:
@zappolowski @fwsmit can I merge this?

cwrau:
Amazing! Is there a timeline for when the next release will happen? 😊

bynect:
> Amazing! Is there a timeline for when the next release will happen? 😊

We don't have a timeline but since it's been a while since the last a new release is almost ready

PR #1376, opened 2024-09-03
Use TESTDIR env var

bynect:
Should resolve the issues pointed by @zappolowski in #1373


bynect:
@apprehensions is this solution ok for you?

zappolowski (approved):

PR #1378, opened 2024-09-04
feat: exit with `1` when paused for `is-paused`

cwrau:
That way the user can more easily use this in a script

instead of

```bash
if [[ "$(dunstctl is-paused)" == "true" ]]; then
  {...}
fi
```

they can do

```bash
if dunstctl is-paused -e; then
  {...}
fi
```

### DIFF
--- a/dunstctl
+++ b/dunstctl
@@ -30,7 +30,7 @@ show_help() {
 	                                    notification with given ID
 	  history-rm ID                     Remove the notification from
 	                                    history with given ID.
-	  is-paused [-e]                    Check if pause level is greater than 0, optionally with exit code instead of text output
+	  is-paused [-e|--exit-code]        Check if pause level is greater than 0, optionally with exit code instead of text output
 	  set-paused true|false|toggle      Set the pause status
 	  get-pause-level                   Get the current pause level
 	  set-pause-level level             Set the pause level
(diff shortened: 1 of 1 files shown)


dunstctl
```diff
@@ -117,7 +117,20 @@ case "${1:-}" in
...
+			exit=true
+		fi
+		property_get paused | {
+			read -r _ _ paused
+			if [ "$exit" = "false" ]; then
+				printf "%s\n" "${paused}"
+			else
+				if [ "${paused}" = "false" ]; then
+					exit 1
+				fi
```
zappolowski, line 131:
```suggestion
			elif [ "${paused}" = "false" ]; then
				exit 1
```

dunstctl
```diff
@@ -117,7 +117,20 @@ case "${1:-}" in
 		fi
 		;;
 	"is-paused")
-		property_get paused | ( read -r _ _ paused; printf "%s\n" "${paused}"; )
+		exit=false
+		if [ "${2:-}" = "-e" ]; then
```
zappolowski, line 121: The only other command taking a flag uses a long form. So, maybe `--exit-code` might be a better name (or maybe allow for long and short forms here).

dunstctl
```diff
@@ -30,7 +30,7 @@ show_help() {
 	                                    notification with given ID
 	  history-rm ID                     Remove the notification from
 	                                    history with given ID.
-	  is-paused                         Check if pause level is greater than 0
+	  is-paused [-e]                    Check if pause level is greater than 0, optionally with exit code instead of text output
```
zappolowski, line 33:
```suggestion
	  is-paused [-e|--exit-code]        Check if pause level is greater than 0, optionally with exit code instead of text output
```
The long form should be documented as well.

cwrau: Ah, forgot that 👍

zappolowski (approved):

PR #1381, opened 2024-09-14
Fix broken link to freedesktop's notification specs in man page

Phantomwise:
Fix the broken link in the `man 5 dunst` page (link to the page describing categories in freedesktop's notification specs)
https://specifications.freedesktop.org/notification-spec/latest/ar01s06.html gives a 404 error.
https://specifications.freedesktop.org/notification-spec/latest/categories.html seems to be the correct link


bynect:
thanks

PR #1383, opened 2024-09-23
Resolve ~ and $HOME in icon_path

### LINKED ISSUE #1371: Resolve `~` or `$HOME` in `icon_path`
### Issue description

dunst doesn't resolve `~` or `$HOME` in `icon_path`. It was [claimed](https://github.com/dunst-project/dunst/issues/350#issuecomment-477550687) to be fixed in `1.4` by @tsipinakis and [here](https://github.com/dunst-project/dunst/issues/520#issuecomment-398317087) by @bebehei  so I'm little puzzled.

### Installation info

- Version: `1.11.0 (2024-04-15)`
- Install type: `package`
- Window manager / Desktop environment: `Hyprland`
- Distro: `Artix`

<details>

```ini
enable_recursive_icon_lookup = false
# doesn't work
icon_path = ~/.config/dunst/icons/
# also doesn't work
#icon_path = $HOME/.config/dunst/icons/
# works
#icon_path = /home/<USER>/.config/dunst/icons/
```
</details>

Related: #350 #520

zappolowski:
`icon_path` was used verbatim which doesn't resolve e.g. `~` or `$HOME`. As `string_to_path` already exists, we can use it for the resolution here as well.

This fixes #1371.

### DIFF
--- a/src/icon.c
+++ b/src/icon.c
@@ -242,7 +242,7 @@ char *get_path_from_icon_name(const char *iconname, int size)
                 end = strchr(start, ':');
                 if (!end) end = strchr(settings.icon_path, '\0'); /* end = end of string */
 
-                current_folder = g_strndup(start, end - start);
+                current_folder = string_to_path(g_strndup(start, end - start));
 
                 for (const char **suf = suffixes; *suf; suf++) {
                         gchar *name_with_extension = g_strconcat(iconname, *suf, NULL);


bynect:
ok for me. it works with multiple paths right?

what about using TYPE_PATH for icon_path tough?

zappolowski:
> it works with multiple paths right?
Yes, every path (separated by `:`) is processed on its own.

> what about using TYPE_PATH for icon_path tough?

I don't think this would work, as `TYPE_PATH` is used for single argument which is then expanded. What might work instead would be the introduction of a new type e.g. named `TYPE_PATH_ARRAY`, which then could be processed on reading the configuration (opposed to expansion on usage like it's currently done).

bynect:
well, I guess this is all right

bynect (approved):

danielwerg:
@zappolowski works as expected!

---

On a side note, for some reason, this branch dunst (and master I assume) uses `global.height` as min height instead of max height (only expand when content size demands) unlike v1.11.0.
Is this intentional or should I open an issue?

bynect:
> @zappolowski works as expected!
> 
> On a side note, for some reason, this branch dunst (and master I assume) uses `global.height` as min height instead of max height (only expand when content size demands) unlike v1.11.0. Is this intentional or should I open an issue?

See #1342. Now global.height is a range. if you specify a single number the max = min and the size is fixed. Otherwise you can specify (min, max). Just like the width value

danielwerg (approved):

Issue #1386, opened 2024-10-16
dunstctl history does not work


hholst80:
```
root@xps ~# dunstctl count
              Waiting: 0
  Currently displayed: 0
              History: 4
root@xps ~# dunstctl history
Dunst is not running.
root@xps ~ [1]# 
```

```
root@xps ~# dunst -v
Dunst - A customizable and lightweight notification-daemon 1.11.0 (2024-04-15)
root@xps ~# 
```

bynect:
Can you confirm if dunst is actually running? Ps aux | grep dunst

Also can you provide a 'dunst -verbosity debug' output when you ask history on another terminal (same user)?

hholst80:
dunst is running, but dunstctl history does not trigger any output on the terminal.

```
# dunst -verbosity debug
DEBUG: [   load_settings:0272] Setting defaults
DEBUG: [     set_setting:0441] [urgency_low] Trying to set icon to dialog-information
DEBUG: [     set_setting:0441] [urgency_normal] Trying to set icon to dialog-information
DEBUG: [     set_setting:0441] [urgency_critical] Trying to set icon to dialog-warning
DEBUG: [     set_setting:0441] [global] Trying to set format to <b>%s</b>
%b
DEBUG: [     set_setting:0441] [global] Trying to set corners to all
DEBUG: [string_parse_enum:0034] Setting enum to 15 (all)
DEBUG: [     set_setting:0441] [global] Trying to set progress_bar_corners to all
DEBUG: [string_parse_enum:0034] Setting enum to 15 (all)
DEBUG: [     set_setting:0441] [global] Trying to set icon_corners to all
DEBUG: [string_parse_enum:0034] Setting enum to 15 (all)
DEBUG: [     set_setting:0441] [*] Trying to set frame_color to #888888
DEBUG: [     set_setting:0441] [experimental] Trying to set per_monitor_dpi to false
DEBUG: [string_parse_enum:0034] Setting enum to 0 (false)
DEBUG: [     set_setting:0441] [global] Trying to set force_xinerama to false
DEBUG: [string_parse_enum:0034] Setting enum to 0 (false)
DEBUG: [     set_setting:0441] [global] Trying to set force_xwayland to false
DEBUG: [string_parse_enum:0034] Setting enum to 0 (false)
DEBUG: [     set_setting:0441] [global] Trying to set font to Monospace 8
DEBUG: [     set_setting:0441] [global] Trying to set sort to true
[...]

PR #1389, opened 2024-10-31
Check DBUS_SESSION_BUS_ADDRESS and give more meaningful errors

bynect:
Addresses #1386

### DIFF
--- a/dunstctl
+++ b/dunstctl
@@ -170,7 +170,7 @@ case "${1:-}" in
 						jq --raw-output '.data[][] | ["[\(.name.data)]"], [to_entries[] | select(.key != "name") | "    \(.key) = \(.value.data)"] | join("\n")'
 					fi
 				} \
-					|| die "Dunst is not running or unreachable."
+					|| die "Dunst is unreachable or the version is too old."
 			;;
 			*)
 				die "Unknown format \"${2}\". Please use either \"--json\" or no option at all."
--- a/dunstctl
+++ b/dunstctl
@@ -10,8 +10,9 @@ DBUS_IFAC_FDN="org.freedesktop.Notifications"
 
 die() {
 	printf "%s\n" "${1}" >&2
-	[ -z "$DBUS_SESSION_BUS_ADDRESS" ] \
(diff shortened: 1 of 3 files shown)


bynect:
I also updated the wiki

dunstctl
```diff
@@ -8,7 +8,12 @@ DBUS_IFAC_DUNST="org.dunstproject.cmd0"
 DBUS_IFAC_PROP="org.freedesktop.DBus.Properties"
 DBUS_IFAC_FDN="org.freedesktop.Notifications"
 
-die(){ printf "%s\n" "${1}" >&2; exit 1; }
+die() {
+	printf "%s\n" "${1}" >&2
+	[ -z "$DBUS_SESSION_BUS_ADDRESS" ] \
+		&& echo "DBUS_SESSION_BUS_ADDRESS is blank. Is the dbus session configured correctly?"
```
bebehei, line 14: You're having `-e` set int the shell. Testing something for negative result and then concatenating the resulting with `&&` will make the shell totally abort without even printing an error message.

```suggestion
	if [ -z "$DBUS_SESSION_BUS_ADDRESS" ]; then
		echo "DBUS_SESSION_BUS_ADDRESS is blank. Is the D-Bus session configured correctly?" >&2
	fi
```

bebehei: Alternatively you can use `||`:

```suggestion
	[ -n "$DBUS_SESSION_BUS_ADDRESS" ] \
		|| echo "DBUS_SESSION_BUS_ADDRESS is blank. Is the dbus session configured correctly?"
```

But I've learnt using an `if` is much better, because nobody sees visibly this intricate behavioral difference.

bynect: thanks

bebehei:
I don't see any problems. But I'm not confident anymore using "not seeing any problems" as a good indicator for having good code in this project.

bynect:
> I don't see any problems. But I'm not confident anymore using "not seeing any problems" as a good indicator for having good code in this project.

I'll just think of this as a LGTM moment then 😆

PR #1397, opened 2024-11-11
Place dunst.service in session.slice

Atemu:
By default, systemd would place it in `app.slice` which doesn't make sense; that's where apps go, not daemons that are part of the desktop session.

`session.slice` makes a lot more sense for a daemon that's part of the desktop session.

Related: https://github.com/flatpak/xdg-desktop-portal-gtk/pull/504

### DIFF
--- a/dunst.systemd.service.in
+++ b/dunst.systemd.service.in
@@ -7,3 +7,4 @@ PartOf=graphical-session.target
 Type=dbus
 BusName=org.freedesktop.Notifications
 ExecStart=##PREFIX##/bin/dunst
+Slice=session.slice


bynect:
Could you link some docs about the slice option? I found this https://www.freedesktop.org/software/systemd/man/latest/systemd.slice.html where it says

> By default, service and scope units are placed in system.slice, virtual machines and containers registered with [systemd-machined(8)](https://www.freedesktop.org/software/systemd/man/latest/systemd-machined.html#) are found in machine.slice, and user sessions handled by [systemd-logind(8)](https://www.freedesktop.org/software/systemd/man/latest/systemd-logind.html#) in user.slice. See [systemd.special(7)](https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#) for more information

Maybe I am misinterpreting this but shouldn't the service be added by default to system.slice?

Atemu:
The upstream docs are confusing on this but this is a *user* service, not a *system* service. I could not find it documented but user services appear to be placed in app.slice by default.

bynect:
I found this https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html
which mentions user.slice and session.slice
The docs of systemd really are 💩 though

Atemu:
It isn't for me. I use the dbus-broker implementation though:

```
$ scu cat dbus.service 
# /etc/systemd/user/dbus-broker.service
[Unit]
Description=D-Bus User Message Bus
Documentation=man:dbus-broker-launch(1)
DefaultDependencies=false
After=dbus.socket
Before=basic.target shutdown.target
Requires=dbus.socket
Conflicts=shutdown.target

[Service]
Type=notify
Sockets=dbus.socket
ExecStart=/nix/store/2zi5iw7p7hvgc4vryfhr6caj19mbnk3q-dbus-broker-36/bin/dbus-broker-launch --scope user
ExecReload=/nix/store/7bvzsk07az1d9hnpy9d5cpvnizbzmdcd-systemd-256.7/bin/busctl --user call org.freedesktop.DBus /org/freedesktop/DBus org.freedesktop.DBus ReloadConfig
Slice=session.slice

[Install]
Alias=dbus.service

# /nix/store/djigkrxi03y95ab4whkr607snwkwxpia-user-units/dbus-broker.service.d/overrides.conf
[Unit]
X-Restart-Triggers=/nix/store/0nifcbjr1ps3p08rvcggpprny9415110-X-Restart-Triggers-dbus-broker

[Service]
Environment="LOCALE_ARCHIVE=/nix/store/b8bmyz7mkfvvlrrrbr2py9b7wlv4xjbh-glibc-locales-2.40-36/lib/locale/locale-archive"
Environment="PATH=/nix/store/sf6y4arqcm100rnnl3dhpg732i774zp6-coreutils-9.5/bin:/nix/store/psjqqbj19n3fqssn38hgz4cv7b7a9alp-findutils-4.10.0/bin:/nix/store/nysgss3ggh3dv7mz08n28h29l>
Environment="TZDIR=/nix/store/5z7yfmp96ijpx0lhy5gpj65i38972330-tzdata-2024b/share/zoneinfo"
X-ReloadIfChanged=true
```

bynect:
Well I will go ahed and merge then

Atemu:
Thanks!

Issue #1406, opened 2024-12-07
1.12.0 fails to render body on notifications when there are tall Unicode characters only in title


Architector4:
### Issue description

My test case is this command:

    dunstify "⛧" "hi"

On Dunst 1.11.0, this shows two lines perfectly:

![image](https://github.com/user-attachments/assets/5f4ba47e-fa82-48b5-b533-b2db3870edab)

On Dunst 1.12.0, this breaks:

![image](https://github.com/user-attachments/assets/a72a7da6-5a7d-488a-86ad-f0d68a1b3943)

The extra line can still be shown if the minimum/constant height is set to a big enough value, but that's obviously suboptimal.

It doesn't break if both title and body have such a character, nor if only the body has one, nor if the body text is long enough.

### Installation info

- Version: `1.12.0 (2024-11-30)`
- Install type: `package`
- Window manager / Desktop environment: `i3wm`
- Distro: `Arch Linux`

Two screenshots above are done with no `dunstrc` in place, i.e. on completely default configuration.

bynect:
I guess it has to do with the dynamic height changes.

Does it occur with unicode special chars in general or only with tall ones?

Architector4:
Only with tall ones, it seems. Greek and Cyrillic letters and random arrows and weird spaces go through just fine.

Seems it also breaks with emojis too; a title of just "😁" causes the issue as well.

bynect:
> Only with tall ones, it seems. Greek and Cyrillic letters and random arrows and weird spaces go through just fine.
> 
> Seems it also breaks with emojis too; a title of just "😁" causes the issue as well.

Could you try a few things

* add a minimum height of 1 like height = (1,300)
* test different values of vertical alignment
* try the provided dunstrc

Architector4:
Side note, a usable hack workaround for anyone else experiencing this may be to change `format` string like this:

    format = "<b>%s</b>\n%b\n "

Compared to default setting, this appends a newline and a word joiner (special Unicode character; it's only significance here is that it's rendered invisible, and doesn't get trimmed like a space in its place seemingly does).

This results in one extra line worth of empty space at the end of every notification, but at least it shows all the text now lol

bynect:
I will look into possible solutions as soon as I can

bynect:
I can replicate this. however still no idea why this happened

Architector4:
[...]

PR #1412, opened 2024-12-10
Add backward compatibility for height + notice

bynect:
As solicited by @stapelberg in the discussions this will be a (temporary) solution for backward compatibility of `height`.
This will *not* make it into v2, but for now is fine.
I will need to release a 1.12.1 anyway due to the graphical bug #1406

### DIFF
--- a/src/option_parser.c
+++ b/src/option_parser.c
@@ -237,12 +237,12 @@ int string_parse_color(const char *s, struct color *ret)
 
 int string_parse_gradient(const char *s, struct gradient **ret)
 {
-        struct color colors[10];
+        struct color colors[16];
         size_t length = 0;
 
         gchar **strs = g_strsplit(s, ",", -1);
         for (int i = 0; strs[i] != NULL; i++) {
-                if (i > 10) {
+                if (i > 16) {
                         LOG_W("Do you really need so many colors? ;)");
                         break;
                 }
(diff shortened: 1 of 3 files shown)


fwsmit:
Thank you for making the change for compatibility. I agree with the users that making the configuration backward compatible is very important. I would even suggest to keep the current configuration working for version 2, since it would cause a lot of work for users to move their configuration over to the new version. Instead you could maybe add a version specifier to the configuration file like "#dunst config V2" and only fix the inconsistencies there.

stapelberg:
Thank you for making this change! This will save time for many users, I expect :)

fwsmit:
> > I would even suggest to keep the current configuration working for version 2
> 
> well, if there are some backward incompatible changes to be made it has to be a major revision.
> I think that if a few changes to the config will be needed by v2 this can be added to the bucket.
> But we'll see when time comes

Well see yeah. For now i think its good to just bring a new release with these fixes

bynect:
> > > I would even suggest to keep the current configuration working for version 2
> > 
> > 
> > well, if there are some backward incompatible changes to be made it has to be a major revision.
> > I think that if a few changes to the config will be needed by v2 this can be added to the bucket.
> > But we'll see when time comes
> 
> Well see yeah. For now i think its good to just bring a new release with these fixes

I want to fix one last thing than I can make 12.1

PR #1413, opened 2024-12-10
Fix gradient ownership in rule

bynect:
Caused occasional double frees

### DIFF
--- a/src/dunst.c
+++ b/src/dunst.c
@@ -223,6 +223,7 @@ void reload(char **const configs)
         setup_done = false;
         draw_deinit();
 
+        settings_free(&settings);
         load_settings(configs);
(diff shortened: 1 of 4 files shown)


bynect:
@fwsmit when debugging some double frees I noticed that rules where not cleaned up when reloading. so I changed it to remove all old rules. which should be the best thing to do right?

bynect:
strange codecov error unrelated to code

fwsmit:
Yeah, rules from an old config should be removed

bynect:
> Yeah, rules from an old config should be removed 

@fwsmit can i make the release now?

fwsmit:
Yeah, go ahead 👍

PR #1418, opened 2024-12-14
feat(completions): Added history-clear completion for zsh

KiLLeRRaT:
Simple, just added history-clear to the zsh completion.


bynect:
thanks. since you are already there could you also add `history-rm`?

KiLLeRRaT:
> thanks. since you are already there could you also add `history-rm`?

Done :)

bynect:
Thanks 👍🏻

PR #1424, opened 2024-12-27
Fix some warnings (gint formatting, size_t/int mix, etc)

bynect:
Related to #1372

### DIFF
--- a/src/icon.c
+++ b/src/icon.c
@@ -268,6 +268,12 @@ char *get_path_from_icon_name(const char *iconname, int size)
         return path;
 }
 
+static void icon_destroy(guchar *pixels, gpointer data)
+{
+        (void)data;
+        g_free(pixels);
+}
+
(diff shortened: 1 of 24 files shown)


src/icon.c
```diff
@@ -351,7 +357,7 @@ GdkPixbuf *icon_get_for_data(GVariant *data, char **id, double dpi_scale, int mi
                                           width,
                                           height,
                                           rowstride,
-                                          (GdkPixbufDestroyNotify) g_free,
```
bynect, line 354: @fwsmit I replaced this g_free with an actual function taking two arguments since it just seemed wrong. Was this supposed to free both the pixels and data_pb?

bynect:
Int formatting was a problem on Openbsd and on x86 architecture. I tried to fix some other warnings like unused param and size_t instead of int

PR #1428, opened 2025-01-01
Set font in tests

### LINKED ISSUE #1404: `test_calculate_dimensions_height_min` test fails
### Issue description

The following test failure is shown when trying to `make test`.

```
FAIL test_calculate_dimensions_height_min: dim.h == expected_height (test/draw.c:330) (133 ticks, 0.000 sec)
```

### Installation info

- Version: `1.12.0`

bynect:
This should fix #1404.

This is kind of an hack, but it should work for the fontconfig + pango problem.
Further testing is required


fwsmit:
Can this not be solved by installing the font in the relevant docker images. I'm guessing the issue is that those images dont have a good font installed that can be picked up by pango

bynect:
> Can this not be solved by installing the font in the relevant docker images. I'm guessing the issue is that those images dont have a good font installed that can be picked up by pango

yes that's the problem. however automated test done downstream by various distros doesn't pull a font. this is more of an hack, but I don't know if they can add some soft dependency on fonts in their ci.

bynect:
@dcermak @fossdd as distro packager does this solution seem right to you? or is a soft dependency on fonts better ?

fwsmit:
A distro test system should be able to configure dependencies for software I would think. Do distros use the dunst test suite to test their packages?

bynect:
> A distro test system should be able to configure dependencies for software I would think. Do distros use the dunst test suite to test their packages?

Yes but the problem is that we introduced a test that fails when no font is present. So all the old packaging started failing when updating to v1.12

the automatic ci runs for the distros started failing because they did not explicitly add the font dependency

fossdd:
For Alpine Linux it's totally fine to add a font to our test dependencies. Sounds easier to maintain than shipping a font in the tarball. It makes sense documenting this somewhere, like in HACKING.md.

dcermak:
I do agree with @fossdd, it's no problem to pull in a font as a build-time dependency, neither in Fedora nor in openSUSE. I would discourage bundling a font and instead document this requirement for the test suite

bynect:
Ok let's do it that way then.
I'll close this proof of concept and add some lines to the dependencies documentation

PR #1429, opened 2025-01-01
Add --category to dunstify

bynect:
I am not sure if the difference with notify-send for `-c` is worth keeping just for printing capabilities...


fwsmit:
I don't think the capabilities printing is used a lot in scripts, so it's okay to change the behavior

fwsmit (approved):

PR #1430, opened 2025-01-02
Add format options "%c" "%U"

bynect:


bynect:
@fwsmit what do you think about this and #1429

fwsmit:
I'm not sure what you would use the urgency string for. An all capital format doesn't seem like something you would want in a notification

bynect:
> I'm not sure what you would use the urgency string for. An all capital format doesn't seem like something you would want in a notification

Yes, but I did not want to change the function we already have. Does it have to be uppercase or can I lower it?

PR #1431, opened 2025-01-08
Add tests for additions in #1426

BinaryInk:
Self-explanatory; added three assertions to test for values for the properties `urgency`, `stack_tag`, and `urls` are being outputted via `dunstctl history`


bynect:
👍🏻

PR #1435, opened 2025-01-22
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH` to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

This date call works with different flavors of date (GNU, BSD, other).
Alternatively the date could be dropped completely to simplify the code.

This patch was done while working on reproducible builds for openSUSE.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,9 +74,15 @@ debug: all
 
 ${OBJ} ${TEST_OBJ}: Makefile config.mk
 
+DATE_FMT = +%Y-%m-%d
+ifdef SOURCE_DATE_EPOCH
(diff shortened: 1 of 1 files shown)


bynect:
The date is not really necessary but it was nice to have.

I understand the need for reproducible builds, so I approve these changes.

If possible I would like to keep support for BSD.

bmwiedemann:
I updated the patch to also support BSD.

PR #1436, opened 2025-01-22
Add clickable menu

Qingwu-Li:
Hello fellow geeks,
I tring to implemented a built-in clickable menu feature,
Tested with wayland,


bynect:
hello. I like your idea and I also wanted to implement the buttons.

I tried your branch and noticed some problems with the text alignment

img-1737672610

I think the best way would be to either specify a maximum amount of buttons allowed (then dmenu is used for extra action) or to wrap the buttons to a newline. squeezing all actions in a single line is impossible when there are a lot after all

gujie-leica:
We revised the settings parameter in the built-in menu and added support for a rounded corner menu in the built-in menu two commits ago.

bynect:
I will try the new changes 👍

gujie-leica:
Now we add key navigation inside of notification:

- left, right, up and down arrow keys select the button (action)
- enter key executes action
- add “focus“ when selected, the first key is the default focus key

bynect:
After this change

```
                        cairo_move_to(c, round(text_x * scale), round(text_y * scale));
```

It almost works correctly, with a minor offset to the text.

img-1746477429

gujie-leica:
> 
> I think splitting the keyword part in a separate pr is better since this one already quite big.

Thank you for your reply. Basically, the drawing and navigation parts are separate, but they share some common changes in input.c. I will update it in the coming days.

bynect:
is this ready for review?

gujie-leica:
> is this ready for review?

Yes, I think so, I just rebase to resolve conflicts.  Sorry we're busy recently, didn't reponse in time.

bynect:
hello I finally tested it properly. codewise it seems ready to merge.
 I would like to add some considerations:
 - I'm not sure I would enable the menu by default since it is a new feature and it would change behavior
 - When the menu is active the old dmenu way of doing actions is still active. probably this is right, but we might consider disabling it when there are few actions
 - for consistency I would probably rename built_in_menu to enable_menu or button_menu

bynect:
also I think that the way the background can be improved. instead of having menu_frame_fill
something like `separator_color` can be used (which can be either `frame` or a color).
personally I would also add a way to change text color inside the menu. but that's something I can add myself.

bynect:
let me know what do you think @Qingwu-Li

bynect:
img-1756027602


long action names should be ellipsized or truncated to avoid this


img-1756027874

a big number of items is handled correctly 👍🏻 
but maybe a limit to the number of button should be added to keep in mind height limitation of the notification
so the excess buttons should be instead handled by dmenu

Qingwu-Li:
> let me know what do you think @Qingwu-Li


I will have a look on it!

src/menu.c
```diff
@@ -392,4 +392,90 @@ static gpointer context_menu_thread(gpointer data)
 
         return NULL;
 }
+
```
fwsmit, line 395: Right now the new code in menu.c is not integrated with the old code. This could lead to confusion with other similar functions in this file. I would suggest to either
1. create a separate dmenu.c for all dmenu stuff and renaming those functions to reflect the fact that there are multiple kinds of menu's
or 2.  Merge some of the new functions with old ones and rename the other existing functions to avoid confusion.

fwsmit (commented):
I have some suggestions that could improve the code, avoiding issues in the future. Other than that it looks good

fwsmit:
Also, could you add some tests, for example for `calculate_menu_height`

ClairDePrince:
Any update?

fedang:
a few touch-ups to the layout calculation and a few tests and this should be mergeable

Issue #1443, opened 2025-02-22
dunstify Crashes with Segmentation Fault when using  -r


DanielLeviLucas:
### Issue description:

- The dunstify command crashes with a segmentation fault when the -r (replace) option is used.
- Example command:
    ```bash
    ➜ dunstify -r "9999" "foo" "bar"
    zsh: segmentation fault (core dumped)  dunstify -r "9999" "foo" "bar"
    ```
    ```bash
    ➜ coredumpctl list dunstify

    Sat 2025-02-22 18:29:41 IST   21990 1000 1000 SIGSEGV present  /usr/bin/dunstify 396.4K
    ```
### Installation info
- Version: 1.12.1 
    Dunst - A customizable and lightweight notification-daemon 1.12.1 (2024-12-17)
    Compiled on 2024-12-22 with the following options:
    X11 support: enabled
    Wayland support: enabled
    SYSCONFDIR set to: /etc
    Compiler flags: -g -std=gnu11 -pedantic -Wall -Wno-overlength-strings -Os -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -g -ffile-prefix-map=/build/dunst/src=/usr/src/debug/dunst -flto=auto -pthread -MMD -MP
    Linker flags: -lm -lrt -Wl,-O1 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,-z,pack-relative-relocs -flto=auto -lgio-2.0 -lgdk_pixbuf-2.0 -lpangocairo-1.0 -lpango-1.0 -lgobject-2.0 -lglib-2.0 -lharfbuzz -lcairo -lwayland-client -lwayland-cursor -lX11 -lXinerama -lXext -lXrandr -lXss
- Install type: arch extras/dunst
- Window manager / Desktop environment: qtile
- Distro: Arch Linux
- libnotify 0.8.4-1

bynect:
it may be related to #1442 
what version of libnotify >=0.8.4 do you have?

DrWarpMan:
> it may be related to [#1442](https://github.com/dunst-project/dunst/issues/1442) what version of libnotify >=0.8.4 do you have?

I am also experiencing this, I have `0.8.4-1`.

[...]

PR #1446, opened 2025-02-25
Use properties instead of private fields in dunstify. #2

Nocturnalx:
fixing #1443
doing similar to #1444 but needed to cast to G_OBJECT first

### DIFF
--- a/dunstify.c
+++ b/dunstify.c
@@ -173,53 +173,16 @@ void parse_commandline(int argc, char *argv[])
     }
 }
 
-typedef struct _NotifyNotificationPrivate
-{
-        guint32         id;
-        char           *app_name;
-        char           *summary;
-        char           *body;
-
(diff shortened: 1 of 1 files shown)


bynect:
Shouldn't it use a GValue?

bynect:
I ended up merging  #1444, thanks anyway

Nocturnalx:
> I ended up merging #1444, thanks anyway

oh yeah you're right, no badge PepeHands but np happy to help :p

liskin:
> Shouldn't it use a GValue?

Why?

This PR actually works, #1444 fails with
```
(dunstify:11400): GLib-GObject-CRITICAL **: 15:04:20.493: g_value_get_int: assertion 'G_VALUE_HOLDS_INT (value)' failed
```

Issue #1445, opened 2025-02-24
duntstify -i icon doesnt show with /tmp/image.png


gh0stzk:
Hello maybe this related to https://github.com/dunst-project/dunst/issues/1408

Im using arch linux with **dunst 1.12.1-1** and **libnotify 0.8.4-1**

this code not work
`dunstify -i /tmp/cover.png "Playing..." "$(mpc --format '%title% \n%artist%' current)"`

But if i chenge to -I it works.

`dunstify -I /tmp/cover.png "Playing..." "$(mpc --format '%title% \n%artist%' current)"`

I need to change all my scripts or is a bug? i see a commit to fix it but apparently still the bug.

But reading the manual i think i need to change my scripts:

```
       -i, --icon=ICON
           Specifies an icon to display with the notification.

       -I, --raw_icon=PATH
           Path to the icon to be sent as raw image data.
```

And i already see and issue with --replace=ID segmentation fault bug, i have this bug too.

gh0stzk:
Yes Sr. this is the part of icons conf i have in my dunst config file.

```
enable_recursive_icon_lookup = true
icon_theme = "TokyoNight-SE, Adwaita"
icon_position = left
min_icon_size = 60
max_icon_size = 90
icon_corner_radius = 6
```

bynect:
Thanks. I'll look into it. Meanwhile if you can could you try if dinst 1.11 worked correctly?

gh0stzk:
i assume is with all paths because i have this little script to change brightness and show notification and the notifications breaks days ago: 

this is part of the code: with the -I works fine.

```
DIR="$HOME/.config/bspwm/src/assets"
icon="$DIR/brightness.svg"

dunstify "Brightness $brightness%" -i "$icon" -r 5555 -u normal \
             -h "int:value:$brightness"
```

Yeah let me downgrade dunst and i will reply soon.

gh0stzk:
i downgraded to dunst-1.11.0-1

 And my image doesnt show but now an informative icon appears

![Image](https://github.com/user-attachments/assets/5c706ce7-acff-4373-995c-b4a8c7111bfc)

But im pretty sure is because the config file.. i cant find the commit before when i changed the config to work with dunst 1.12

Beyond that, dunstify -i /path/ should work I think, but it shows an alternative icon.

bynect:
> Beyond that, dunstify -i /path/ should work I think, but it shows an alternative icon.

Could you clarify this point.

[...]

PR #1449, opened 2025-02-26
Fix dunstify icon and ids

bynect:
fixing #1445

### DIFF
--- a/dunstify.c
+++ b/dunstify.c
@@ -27,21 +27,21 @@ static gchar **rest = NULL;
 
 static GOptionEntry entries[] =
 {
-    { "appname",      'a', 0, G_OPTION_ARG_STRING,       &appname,        "Name of your application", "NAME" },
-    { "urgency",      'u', 0, G_OPTION_ARG_STRING,       &urgency_str,    "The urgency of this notification", "URG" },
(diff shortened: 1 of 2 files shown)


dunstify.c
```diff
@@ -40,6 +41,7 @@ static GOptionEntry entries[] =
     { "replace",      'r', 0, G_OPTION_ARG_INT,          &replace_id,     "Set id of this notification.", "ID"},
     { "close",        'C', 0, G_OPTION_ARG_INT,          &close_id,       "Close the notification with the specified ID", "ID"},
     { "block",        'b', 0, G_OPTION_ARG_NONE,         &block,          "Block until notification is closed and print close reason", NULL},
+    { G_OPTION_REMAINING, 0, 0, G_OPTION_ARG_FILENAME_ARRAY, &rest, NULL, NULL },
```
Narrat, line 44: Just a style note: It isn't aligned to the rest of the options if that is of concern :)

bynect: Ops my bad. Do you think the other changes are sound?

bynect:
@liskin is it okay now?

dunstify.c
Narrat: I'm in no position to tell :) Would have to take a deeper look at the dunst code.

liskin:
> @liskin is it okay now?

The get/set id code looks okay now. Thanks!

PR #1457, opened 2025-03-10
filter prefix maps from the embedded build flags

pslldq:
To aid reproducible compilation there exist three compiler/linker commands to map build time dependent paths to generic paths: -fdebug-prefix-map, -fmacro-prefix-map and -ffile-prefix-map These allow compilations from different paths to generate the same paths in the resulting executable.

Including these flags in the executable defeats the purpose of leaking build paths into the executable. Therefore just filter them out.


bynect:
Thanks for pointing this out

PR #1458, opened 2025-03-17
wayland: improve deinit and init

### LINKED ISSUE #1434: `dunstctl reload` from #1350 stops dunst from working on native waylandd
Hey,

As I mentioned in #1350 , the `dunstctl reload` mechanism does not seem to work correctly when running dunst through wayland native. It seems to reload fine if a notification is not visible on screen, but if a reload occurs whilst a notification is visible, no new notifications will show up for some reason. They still show in the verbose logs.

It does however work fine on X11 and XWayland.

As mentioned in the comments there, I tried the latest master with no change.

The logs after running `dunst -verbosity debug`, and triggering the bug are as follows: https://paste.rs/yrEGc.txt

If any more info is required please do not hesitate to let me know. I'll be happy to provide anything else.

pslldq:
Improve the wayland deinit and init process to fix problems after running deinit followed by init in case of a reload. Furthermore destroy otherwise leaked objects and clear associated pointers to prevent potential use after free bugs.

This change fixes dunst from breaking sometimes, when a reload is triggered while a notification is displayed. This issue happed because of either ctx.dirty not being cleared, which causes an set_dirty to do an early exit without scheduling a frame to the surface. Or because the frame_callback was not being cleared, which prevented further surface commits, as the callback would never be triggered after the surface was destroyed by the deinit process.

Fixes #1434

### DIFF
--- a/src/wayland/wl.c
+++ b/src/wayland/wl.c
@@ -220,6 +220,7 @@ static const struct wl_registry_listener registry_listener = {
 };
 
 bool wl_init(void) {
+        ctx.dirty = false;
         wl_list_init(&ctx.outputs);
         wl_list_init(&ctx.seats);
         wl_list_init(&toplevel_list);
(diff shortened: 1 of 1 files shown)


bynect:
thank you sven for your work ❤️

PR #1461, opened 2025-03-20
Add an .editorconfig file

### LINKED ISSUE #1455: Introduce .editorconfig
I've noticed there is a specific way as to how dunst is written, and I'd like to see a .editorconfig to be made to accommodate this.

An .editorconfig can better maintain a consistent coding style within dunst :)

pslldq:
Add an `.editorconfig` mostly trying to match the current styles used in the repository. Consistently indented inconsistent files and moved relevant source files to use spaces instead of tabs. Added necessary checking to the CI using `editorconfig-checker`. As it doesn't support "fancy indentations" (e.g. align arguments), the indentation check was disabled.

Fixes #1455

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,55 @@
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = space
+indent_size = 8
+tab_width = 8
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[docs/*.pod]
+indent_size = 4
+
+[*.awk]
+indent_size = 4
+
+[{test/greatest.h,test/functional-tests/test.sh,contrib/notification-history.sh,contrib/dunst_xr_theme_changer.sh}]
+indent_size = 4
+
+[dunstrc*]
+indent_size = 4
+
+# Just for historical purposes to avoid
+# manually reformatting whole files
+[src/wayland/pool-buffer.*]
+indent_style = tab
+
+# looks to be external code
+[src/wayland/libgwater-wayland.*]
+indent_size = 4
+
+[completions/*]
+indent_size = 4
+
+[.valgrind.suppressions]
+indent_size = 3
+
+[*.yml]
+indent_size = 2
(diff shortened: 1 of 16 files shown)


bynect:
I would rather prefer 4 space indentation in c files. @fwsmit would that be okay?

fwsmit:
I don't have a problem with that. It's only that if you change it you will polute the git blame history with indent changes everywhere. So I would suggest leaving it. Any code that is indented so many times as to become an issue with 8 space indent should probably be rewritten anyways to have better code structure

bynect:
I am inclined to agree. Ideally the indentation should be adjusted on future changes to those files as to prevent git history pollution. I am not sure its possible with an automated tool however

pslldq:
> I am inclined to agree. Ideally the indentation should be adjusted on future changes to those files as to prevent git history pollution. I am not sure its possible with an automated tool however

For this (and C files) [clang-format](https://clang.llvm.org/docs/ClangFormat.html) is the tool of choice. Although it needs a bunch of configuration, as it will reformat the whole code (where do braces go, case statement indented, max line length, etc.)

> I would rather prefer 4 space indentation in c files. @fwsmit would that be okay?

ah ok, I'll update the PR to remove the 4 -> 8 indention changes the 3 or 4 c files, so that my indentation commit only affects files with mixed spaces.

pslldq:
updated the PR to just reformat files with mixed spaces

bynect:
this looks good. probably the way to go is to transition slowly to 4 spaces everywhere by removing the exceptions one by one going forward

pslldq:
@bynect I'm more on the autoformatting side to get it consistent in one big commit (and make a bunch of other stuff equally formatted between source files). See #1464

pslldq:
Whats the plan to continue on this? I could also remove all whitespace fixups, as these are not checked by the editorconfig-checker (because enabling the check would cause false positives for all aligned indents in the C code).

bynect:
For now I would prefer starting with this one. For autoformat we'll decide what to do then

PR #1464, opened 2025-03-31
Introduce clang-format

pslldq:
Automatically format the code with clang-format to always have a consistent style.

Options used:
- Indent width of 4 as of https://github.com/dunst-project/dunst/pull/1461#issuecomment-2741511566
- Only move function body to the same line, if it's empty (e.g. `void foo() {}`, but not `void foo() { bar(); }`)
- Add braced below functions, but directly after conditional/loop statements (Mozilla style)
- Compact short case labels on a single line (primarely for `case A: return foo;`)
- Also move binary operator to the start of the next line (if it is broken up)
- Either have all parameters/arguments on the same line or split it up into one line for each
- Alignment stuff...

@bynect @fwsmit Would like to hear your thoughts about this style. Some additional desires or maybe remove some of the options?

Also feel free to point out potential ways, where the formatting really breaks the understanding. If they cannot be formatted in a better way, then they always can be excluded using comments.

TODO myself: Add some github action to check the code matches the formatting.


bynect:
Is this alongside the editorconfig pr or supersedes it?

The config seems sensible but my problem is that such a big commit would dirty the git history

pslldq:
> Is this alongside the editorconfig pr or supersedes it?

no, this is alongside the editorconfig PR. The PR is just included to see, if the formatting trips up something in the editorconfig checker.

The main purpose of the editorconfig is to avoid having to manually adjust the editor each time (use tabs/spaces, indention width, ...). It's like vim modelines, but supported by much more editors by default. Checking the editorconfig is just to prevent some basic stuff.

clang-format on the other hand precisely defines how C/C++ files should be formatted and therefore removes manual formatting and potential discussion entirely.

> The config seems sensible but my problem is that such a big commit would dirty the git history

You can partially work around with a [`.git-blame-ignore-revs`](https://www.michaelheap.com/git-ignore-rev/) file. Or the formatting could only be applied to files changed in future git commits (with `git clang-format`). Latter is nice, but will hurt for the 8->4 space indentation, as only newly touched code is formatted with 4 spaces.

pslldq:
General thoughts how to proceed? As described above, there are multiple ways to handle the potential reformatting. And the still missing Github CI Action in this PR would depend on this choice.

I also wouldn't mind if this PR is closed, but in this case I want to avoid the effort of implementing these CI Actions.

bynect:
For now let's use the editorconfig. I am not opposed to using autoformat but I have some things to still think through.

Also since we are going to reformat everything we may decide to use a slightly different style.

PR #1473, opened 2025-04-28
Change the caching of icons

bynect:


bynect:
#586

bynect:
this is not working out the way I intended. maybe icon_id should be scrapped altogether..

bynect:
@fwsmit if you have the time could you check if these changes are sound? the testing does work, but I did change some pretty critical parts

PR #1474, opened 2025-05-05
Fix version mismatch between Makefile and meson

### LINKED ISSUE #1462: Meson based build creates invalid systemd unit file
Decided to build a new package with the new build system in place.
And after a restart I noticed that `dunst` won't display anymore.
```
$ notify-send test
GDBus.Error:org.freedesktop.DBus.Error.NameHasNoOwner: Could not activate remote peer 'org.freedesktop.Notifications': activation request failed: unit is invalid
```

Looking at the user unit file it contains two exec lines (and both of them are wrong for the generated package):
```
[Unit]
Description=Dunst notification daemon
Documentation=man:dunst(1)
PartOf=graphical-session.target

[Service]
Type=dbus
BusName=org.freedesktop.Notifications
ExecStart=##PREFIX##/bin/dunst
ExecStart=bin/dunst
Slice=session.slice
```

So ~two~ one issue~s~:
1) Both lines already there in [dunst.systemd.service.in](https://github.com/dunst-project/dunst/commit/566a4539f5a6c75d582ad07a726c45410a3418d7#diff-e6a6f2146ddd76ab8f704c27acfd97e70835561d7ae47943729b7e9475325de4).
I suppose it is due to keep supporting both build systems? But doesn't help if the unit file gets invalidated

~2) If `--bindir` is unset it should respect the value of `--prefix` as a fallback (like it is done in the [Makefile](https://github.com/dunst-project/dunst/blob/e4eedd9853a92c277d81da78bbc9ef7ed6756d1e/config.mk#L3C1-L3C24). Could be circumvented with explicitly setting `bindir`, but shouldn't be necessary. (And not that much of a problem, as /bin symlinks to /usr/bin, but still?)~
Edit: I must have mixed up the files. Looking at the installed user unit file contains the right path to the command

### How was it built?
With a `PKGBUILD` to create a package for an Arch distribution. Essential commands used in said PKGBUILD:
```shell
arch-meson ${_pkgname} build
meson compile -C build
```

`arch-meson` being a wrapper and setting the following options:
```
  --prefix        /usr \
  --libexecdir    lib \
  --sbindir       bin \
  --buildtype     plain \
  --auto-features enabled \
  --wrap-mode     nodownload \
  -D              b_pie=true \
```

bynect:


bynect:
unfix #1462

PR #1475, opened 2025-05-05
Fix meson configuration files

### LINKED ISSUE #1462: Meson based build creates invalid systemd unit file

bynect:
fix #1462

### DIFF
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -15,7 +15,7 @@ custom_target(
     command: [
         pod2man,
         '--name=dunst',
-        '--center=Dunst Reference',
+        '--center="Dunst Reference"',
         '--section=1',
         pod2man_version_arg,
         '@INPUT@',
@@ -32,7 +32,7 @@ custom_target(
     command: [
         pod2man,
         '--name=dunst',
-        '--center=Dunst Reference',
+        '--center="Dunst Reference"',
         '--section=5',
         pod2man_version_arg,
         '@INPUT@',
@@ -48,8 +48,25 @@ custom_target(
     output: 'dunstctl.1',
     command: [
         pod2man,
-        '--name=dunst',
-        '--center=dunstctl Reference',
+        '--name=dunstctl',
+        '--center="Dunstctl Reference"',
+        '--section=1',
+        pod2man_version_arg,
+        '@INPUT@',
+        '@OUTPUT@',
+    ],
+    install: true,
+    install_dir: man1,
+)
+
+custom_target(
+    'dunstify_pod2man',
+    input: 'dunstify.pod',
+    output: 'dunstify.1',
+    command: [
+        pod2man,
+        '--name=dunstify',
+        '--center="Dunstify Reference"',
         '--section=1',
         pod2man_version_arg,
         '@INPUT@',
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -15,7 +15,7 @@ custom_target(
     command: [
         pod2man,
         '--name=dunst',
-        '--center="Dunst Reference"',
+        '--center=Dunst Reference',
         '--section=1',
         pod2man_version_arg,
         '@INPUT@',
@@ -32,7 +32,7 @@ custom_target(
     command: [
         pod2man,
         '--name=dunst',
-        '--center="Dunst Reference"',
+        '--center=Dunst Reference',
         '--section=5',
         pod2man_version_arg,
         '@INPUT@',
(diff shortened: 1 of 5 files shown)


docs/meson.build
```diff
@@ -15,7 +15,7 @@ custom_target(
     command: [
         pod2man,
         '--name=dunst',
-        '--center=Dunst Reference',
+        '--center="Dunst Reference"',
```
eli-schwartz, line 18: Since meson uses arrays and safely quotes them, this will result in invoking pod2man with:

```
pod2man --name="dunst" --center="\"Dunst Reference\"" --section="1" ....
```

which may or may not be what you want...

bynect: oops didn't know that the arguments were actually parsed. thanks for the tip

bynect: eli do you know perhaps of a way to include the cflags used by meson in the executable itself?
dunst -version did display the compilation flags when make is used but I can't find a way to replicate that

eli-schwartz: Sounds similar to https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonbuild_options

If the cflags are passed via

```
meson setup builddir/ -Dc_args="$CFLAGS"
```

Then those should be getting recorded in `meson.build_options()`.

You could also directly use:
```meson
cflags_str = ' '.join(get_option('c_args'))
```

With zero guarantees about their quotedness.

bynect: > Sounds similar to https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonbuild_options
> 
> If the cflags are passed via
> 
> ```
> meson setup builddir/ -Dc_args="$CFLAGS"
> ```
> 
> Then those should be getting recorded in `meson.build_options()`.
> 
> You could also directly use:
> 
> ```meson
> cflags_str = ' '.join(get_option('c_args'))
> ```
> 
> With zero guarantees about their quotedness.

it seems like this could work only for args passed directly by the user. is there no way to get all the arguments used by the compiler actually? like all defines and includes and whatnots included in the project?

bynect: in the makefile we do it here: https://github.com/dunst-project/dunst/blob/669c586d00548c14c54083d3ef5e475805294071/Makefile#L82

bynect: I will admit that this feature was added more for the coolness factor. I wanted it to be like gcc/vim.
But if meson can't do it I'll just remove it at some point as the Makefile is phased out :(

eli-schwartz: Vim does show the individual CFLAGS / LDFLAGS, though I'm not exactly sure why I'd care. As a vim user, I've never looked at that information once.

GCC shows the configure options, so, equivalent to `meson.build_options()`, and that is something I've definitely looked at before and found quite helpful to see which features were enabled at configure time.

PR #1480, opened 2025-05-20
Add navigation for clickable menu

gujie-leica:
Hello fellow geeks,

I add key navigation inside of notification:

- left, right, up and down arrow keys select the button (action)
- enter key executes action
- add “focus“ when selected, the first key is the default focus key

I tested it in Wayland, hope this makes the built-in menu more user-friendly.

This PR is based on #1436, which introduces a built-in clickable menu feature.

Additionally, since this functionality depends on libxkbcommon, I have added it to the CI Docker image in dunst-project/docker-images#16.


gujie-leica:
V2: rebase code.

bynect:
i added libxkbc to the dockerfiles. ci looks good, the only problem is something with doxygen

src/input.c
```diff
@@ -55,6 +56,35 @@ struct notification *get_notification_at(const int y) {
...
+                                break;
+                            }
+                        }
+
+                        m->n = n;
+                        menu_activate(m);  // activate the selected menu item
+                        return true;
+                    }
+                }
+            }
```
bynect, line 84: somehow the brace is unaligned

config.mk
```diff
@@ -67,6 +67,7 @@ pkg_config_packs := gio-2.0 \
 ifneq (0,${WAYLAND})
 pkg_config_packs += wayland-client
 pkg_config_packs += wayland-cursor
+pkg_config_packs += xkbcommon
```
bynect, line 70: is xkb only used in wayland?

src/wayland/wl_seat.h
```diff
@@ -41,6 +41,10 @@ struct dunst_seat {
                         int32_t x, y;
                 } pts[MAX_TOUCHPOINTS];
         } touch;
+
+        struct {
+                struct wl_keyboard *wl_keyboard;
```
bynect, line 46: if you know this well, do you think this wayland keyboard interface can also be used for the shortcuts that dunst offers in Xorg? (https://dunst-project.org/documentation/#Keyboard-shortcuts-X11-only)

bynect (commented):
i didn't write a lot because most of the code is for the menu. I will also review the other pr

src/input.c
gujie-leica: > somehow the brace is unaligned

I revised it. Thank you.

config.mk
gujie-leica: > is xkb only used in wayland?

In the current code, `libxkbcommon` is used only for Wayland, as X11 has its own built-in XKB extension and doesn't require libxkbcommon.

src/wayland/wl_seat.h
gujie-leica: > if you know this well, do you think this wayland keyboard interface can also be used for the shortcuts that dunst offers in Xorg? (https://dunst-project.org/documentation/#Keyboard-shortcuts-X11-only)

In my opinion, menu navigation utilizes simple arrow keys and 'Enter' within the notification context, which is fundamentally different from global shortcuts that require modifier combinations, such as 'modifier' +Key. We provide intuitive UI navigation similar to dialog boxes or web forms, while the existing shortcuts are system-wide hotkeys for operations like closing all notifications or viewing history.

gujie-leica: > if you know this well, do you think this wayland keyboard interface can also be used for the shortcuts that dunst offers in Xorg? (https://dunst-project.org/documentation/#Keyboard-shortcuts-X11-only)

In my opinion, menu navigation utilizes simple arrow keys and 'Enter' within the notification context, which is fundamentally different from global shortcuts that require modifier combinations, such as 'modifier' +Key. We provide intuitive UI navigation similar to dialog boxes or web forms, while the existing shortcuts are system-wide hotkeys for operations like closing all notifications or viewing history.

pslldq: @bynect This is only the standard keyboard interface for wayland applications to get keyboard events. Usually the compositor will only pass the keyboard events to the currently focused application. Given that dunst doesn't want (and probably can't) steal the focus, it'll only work for a limited set of use cases.

For proper global shortcuts the direction seems to be the [Global Shortcuts Portal API from Flatpack](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GlobalShortcuts.html). GNOME already implements it. But for wl-roots it's still an open issue: https://github.com/emersion/xdg-desktop-portal-wlr/issues/240

bynect: that's what I thought. thank you for confirmation 👍🏻

fwsmit:
This PR is based on #1436, containing a lot of the same code. However it uses totally different commits. This means that if #1436 is merged, this PR would likely have a lot of merge conflicts. I suggest either dropping #1436, because this all of its functionality, or properly rebasing on this PR (which would involve some git trickery).

bynect:
I originally proposed to split the menu button and this pr since this added libxkd dependency. I think the best course is to merge the other pr first and then rebase this one on top of it.

fwsmit:
> I originally proposed to split the menu button and this pr since this added libxkd dependency. I think the best course is to merge the other pr first and then rebase this one on top of it.

Yeah, that could work

PR #1486, opened 2025-06-03
Fix a Use-After-Free issue

oss-belobog:
Hello!

We are casually auditing some open-source projects we are using and we found this straightforward use-after-free vulnerability.

When a `get_rule` fails and returns `NULL`, the `name` is free()-ed (line 631) but soon reused for logging (line 638), which is dangerous. This patch fixes this by always free-ing the variable only if it is no longer used.  Even if `get_rule` will unlikely fail at this moment, this patch enhances the code to avoid potential misuse in the future.

Thanks for the brilliant project anyway!


bynect:
thank you

PR #1489, opened 2025-06-21
feat: add mouse scroll action scripting to notificaiton

### LINKED ISSUE #1490: Add mouse action scripting to dunst notificaitons.
Having a few selected actions like, open_url, close_all ... is nice.
But I want do more than that on my notifications like:-
- if volume notification is their, use mouse wheel to increase or decrease my volume
- if brightness notification is their, use mouse wheel to increase or decrease my brightness
- if I get a notification, and on the basis of which mouse is clicked, I should be able to run my own script.
> for example:- 
I have setup a notificaiton system that sends me notification if price of something is low.
then being able to attach a script that gets triggred when I do different mouse click would be great.
If I press left click run script that buys that item,
if I press right click don't buy that,
If I scroll up, increase the bidding
if I scroll down, decrease the bidding

leyuskckiran1510:
#### how to use it
- add mouse action `mouse_scroll_forward = forward_script` as  and `mouse_scroll_backward = back_script`  [optional as it's by default]
- now add your costume rule to trigger for notification in your `dunstrc`;
```bash
# a example 
[volume]
    sumamry = "changevolume"
    script_mouse_forward = change_vup
    script_mouse_back = change_vdown
```
[note:- script_mouse_forward/back takes single word script to run, you cannot put long script their, make a file of the script you want to execute, and give it a simple single word name and add that to path, do not forget to add executable flag  `chmod +x my_script` ]
- now you can do something like `dunstify "changevolume" "volume=10%"`
- now if you hover to that notification and scroll with mouse wheel, your script should be triggred

#### why we need this
- this allows us to control linear actions, through our mouse wheel, instead of just clicking on mouse.
- this opens ups the possibility of triggering actions, when needed, for example if you have a setup which warns if volume it to high. You can just use mouse wheel to change the volume.

## TODOS:
- if this concept is liked my many, we can also add feature like script_mouse_click1/2/3
- so that we can execute different set of scripts for different mouse action

Closes #1490


leyuskckiran1510:
demo video

https://github.com/user-attachments/assets/ed40c696-0056-45da-8dca-0523741c5e04

bynect:
hello. thank you for the contribution. I like the idea however there are some things we might improve. I am busy right now so I will just leave some initial review. also please fix the  errors

src/settings_data.h
```diff
@@ -312,6 +312,8 @@ static const struct string_to_enum_def mouse_action_enum_data[] = {
         {"context",        MOUSE_CONTEXT },
         {"context_all",    MOUSE_CONTEXT_ALL },
         {"open_url",       MOUSE_OPEN_URL },
+        {"forward_script", MOUSE_FORWARD_SCRIPT },
```
bynect, line 315: maybe there is a better name for this scripts? Ideally we could make a custom script that can be used by other buttons as well

leyuskckiran1510: okay, so just `mouse_script`  for the mouse action, and on rule define the different mouse action roles right ?
correct me if I am wrong,

src/notification.h
```diff
@@ -19,6 +18,12 @@ enum icon_position {
         ICON_OFF
 };
 
+enum mouse_action_type{
```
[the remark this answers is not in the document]
leyuskckiran1510: O yes, MAT -> matrix, mb will make those changes to, rn I am at work, will make those changes at night and update it

leyuskckiran1510:
> hello. thank you for the contribution. I like the idea however there are some things we might improve. I am busy right now so I will just leave some initial review. also please fix the errors

about the error, their seems to be memory leak issue in my code, will fix those issues to, thank you

src/settings_data.h
bynect: yes I was thinking something like that

leyuskckiran1510:
> > hello. thank you for the contribution. I like the idea however there are some things we might improve. I am busy right now so I will just leave some initial review. also please fix the errors
> 
> about the error, their seems to be memory leak issue in my code, will fix those issues to, thank you

i am working on, it will update when I am done, rn now busy with other stuffs

bynect:
dont worry. let me know when its ready for review

leyuskckiran1510:
@bynect could you rerun the workflow, I think the leak was due the mouse_script_scroll  not being  freed, I think the last commits solves the issue. Could you verify that and let me know if that fixes

leyuskckiran1510:
@bynect Do i need to make 100% code coverage too ? or the task here is done for me.
Please do let me know on changes required.

bynect:
Sorry right now I can't review

src/input.c
```diff
@@ -97,16 +105,22 @@ void input_handle_click(unsigned int button, bool button_down, int mouse_x, int
...
                         struct notification *n = get_notification_at(mouse_y);
-
                         if (n) {
                                 if (act == MOUSE_CLOSE_CURRENT) {
                                         n->marked_for_closure = REASON_USER;
                                 } else if (act == MOUSE_DO_ACTION) {
                                         notification_do_action(n);
                                 } else if (act == MOUSE_OPEN_URL) {
                                         notification_open_url(n);
+                                } else if (act == MOUSE_SCROLL_SCRIPT){
```
bynect, line 117: could you please properly indent this? we leave a space between if and (
also spaces between == and commas. but the content seems right

src/notification.h
```diff
@@ -19,6 +18,12 @@ enum icon_position {
...
+    MOUSE_ACTION_OTHERS=0,
```
bynect, line 22: =0 is unnecessary

src/notification.h
```diff
@@ -86,8 +91,17 @@ struct notification {
 
         enum markup_mode markup;
         char *format;
+
+        enum mouse_action_type script_choice;
```
bynect, line 95: rather than having this choice variable inside notification I would rather add mouse_action parameter to run_script function. this seems a bit more clean and you don't risk forgetting to set the type of action beforehand

src/settings_data.h
```diff
@@ -1321,6 +1344,26 @@ static const struct setting allowed_settings[] = {
...
+                .type = TYPE_LIST,
+                .default_value = "mouse_scroll_script",
+                .value = &settings.mouse_scroll_up,
+                .parser = NULL,
+                .parser_data = GINT_TO_POINTER(MOUSE_LIST),
+        },
+        {
+                .name = "mouse_scroll_down",
+                .section = "global",
+                .description = "Action of right click event",
```
bynect, line 1360: please update the description. also of scroll_up

bynect:
I left a review. mostly it is formatting. but there are some architecture changes I suggest. let me know what do you think

leyuskckiran1510:
> I left a review. mostly it is formatting. but there are some architecture changes I suggest. let me know what do you think

okay I will make those changes and let you know, thank you for the time to review the code

PR #1463, opened 2025-03-25
Add MOUSE_CLEAR_CURRENT action

ORDL-AMF:
**Title**: Add `MOUSE_CLEAR_CURRENT` Action to Close Notification Under Cursor

**Description**:
Introduces a new mouse action, `MOUSE_CLEAR_CURRENT`, allowing users to close the notification under the mouse cursor with a left-click or desired by editing 'dunstrc' file. This enhances user control over notifications.

### Changes
- **src/settings.h**: Added `MOUSE_CLEAR_CURRENT` to `enum mouse_action`.
- **src/settings_data.h**: Added `clear_current` | `MOUSE_CLEAR_CURRENT`.
- **src/input.c**: Implemented in `input_handle_click` to close the notification, remove it from history, and redraw.

### Details
- Closes notification with `REASON_USER`.
- Uses a static `dunst_status` to reset fullscreen/pause/idle states.
- Compatible with X11 and Wayland.
- Triggers redraw via `wake_up` even if no notification is found.

### Testing
- Verified on X11 and Wayland simulators; works as intended.
- No conflicts with existing functionality across 65 files.

Feedback welcome!

PR #1491, opened 2025-06-27
Feature: remove_current in notification queues

FelipeFTN:
# Pull Request Purpose :thinking: 
This pull request introduces a new "**remove_current**" mouse action to allow users to **remove notifications from history**; implementing its functionality in the notification handling logic, and updating tests and documentation. :rocket: 

I really love the dunst project, and I think a very basic and useful feature that it's currently missing is to click and remove the notification from history.
So I decided to implement it! :tada: 

**Here's my quick contribution to the project :heart:** 
The main ideia, is to have "do_action" and "remove_notification" to work together nicelly, and avoid stuff like this:

<div align="center">

image
> A history with a lot of already read/clicked notifications

</div>

I hope the Pull Request goes well and the code is not that mess hahaha
This PR was highly inspired by #1463 :rocket:

I hope it works and get merged soon for other users to also use this amazing feature! :pray: 

PS: I also updated the docs to include this new option/functionality, and added a tiny test suit in the valid_options, nothing big. :+1: 

Thanks!


dunstrc
```diff
@@ -320,13 +320,14 @@
...
     # * close_current: Close current notification.
+    # * remove_current: Remove current notification from history.
     # * close_all: Close all notifications.
     # * context: Open context menu for the notification.
     # * context_all: Open context menu for all notifications.
     # These values can be strung together for each mouse event, and
     # will be executed in sequence.
     mouse_left_click = close_current
-    mouse_middle_click = do_action, close_current
+    mouse_middle_click = do_action, remove_current
```
bynect, line 330: I think it's better to keep the current default just in case

src/dbus.c
```diff
@@ -1117,6 +1117,63 @@ void signal_notification_closed(struct notification *n, enum reason reason)
         }
 }
 
+void signal_notification_removed(struct notification *n, enum reason reason)
```
bynect, line 1120: i am not sure this signal is necessary. because we already have the one for closed and for history removed.

to my understanding remove_current does a close + remove_history

bynect (changes requested):
thanks for your contribution. I think that it's best to not change the default values of the config.

Also for the code, instead of making a separate queues_remove... function, why not reuse the existing functions to close and remove history?

bynect:
let me know what do you think

FelipeFTN:
**Hello, @bynect, Thanks for helping me with this contribution!** :tada: 

I just read the comments, nice catches! I'll just point some questions here for we to finish this feature as soon as possible!

So, the idea of keeping the dunstrc file as is, is great! I should not have changed the default behavior haha Just thought about it after opening the Pull Request, but really doesn't seems to be a good idea! :sweat_smile: 
I will be reverting **remove_current** as a default behavior with **do_action** in the next commit, thx!!

Also, you pointed that, we already have functions to handle close and remove signals.
And in the second commentary you suggested using the already built-in functions for **close notification** and **remove from history**. well pointed, but I was thinking about separating the responsibilities of close and history do keep them handling just what they should. With that, in case you modify one of these functions in the future, it doesn't affect the remove_current; that logic I also tried to apply to the **remove current signal**.

Yes, you're right, the remove_current does a simple **close & remove from history**. But seems cleaner to keep separated responsibilities (close functions only for close action, remove functions only for removing actions, also, I thought the history removing should be used with the close_all or something).

Please, tell me what you think about this logic I used, and if it's incorrect! :pray: 
My main idea is to keep using separated queues, signals and menu logic isolated for **remove_current** action!

Thanks again for your messages! If you believe it's strictly necessary or better to keep the code using shared functions to achieve this new action (remove_current), just tell me and I will change it right away! :heart:

bynect:
you are right that the operation in the future could differ from a simple combination of those function. as a compromise what about declaring functions remove_current but using internally those two functions? 
so that if we ever need we can change the implementation later on

FelipeFTN:
**Hello, @bynect!** The changes are done! Let me know if you have any update. :pray:

bynect:
code looks good. the ci problem is unrelated. I will try it and merge shortly 👍🏻

bynect:
thanks

PR #1505, opened 2025-09-06
Ensure PropertiesChanged signal is sent when handling SIGUSR*

### LINKED ISSUE #832: Subscription for some dbus properties doesn't work
Subscribers to the dbus properties `displayedLength`, `historyLength` and `waitingLength` are not informed about changes. In contrast this works with the property `paused`.

Start:
```
$ dbus-monitor path=/org/freedesktop/Notifications
signal time=1613901346.651660 sender=org.freedesktop.DBus -> destination=:1.134 serial=2 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameAcquired
   string ":1.134"
signal time=1613901346.651714 sender=org.freedesktop.DBus -> destination=:1.134 serial=4 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameLost
   string ":1.134"
```

After sending a notification with `$ notify-send test`:
```
method call time=1613901357.736964 sender=:1.135 -> destination=:1.116 serial=6 path=/org/freedesktop/Notifications; interface=org.freedesktop.Notifications; member=GetServerInformation
method call time=1613901357.738638 sender=:1.135 -> destination=:1.116 serial=7 path=/org/freedesktop/Notifications; interface=org.freedesktop.Notifications; member=Notify
   string "notify-send"
   uint32 0
   string ""
   string "test"
   string ""
   array [
   ]
   array [
      dict entry(
         string "urgency"
         variant             byte 1
      )
   ]
   int32 -1
signal time=1613901359.306892 sender=:1.116 -> destination=:1.135 serial=26 path=/org/freedesktop/Notifications; interface=org.freedesktop.Notifications; member=NotificationClosed
   uint32 6
   uint32 2
```

After pause dunst with `dunstctl set-paused true`:
```
method call time=1613901403.092561 sender=:1.136 -> destination=org.freedesktop.Notifications serial=2 path=/org/freedesktop/Notifications; interface=org.freedesktop.DBus.Properties; member=Set
   string "org.dunstproject.cmd0"
   string "paused"
   variant       boolean true
signal time=1613901403.093260 sender=:1.116 -> destination=(null destination) serial=27 path=/org/freedesktop/Notifications; interface=org.freedesktop.DBus.Properties; member=PropertiesChanged
   string "org.dunstproject.cmd0"
   array [
      dict entry(
         string "paused"
         variant             boolean true
      )
   ]
   array [
   ]
```

### Installation info

- Version: `Dunst - A customizable and lightweight notification-daemon v1.5.0-93-g3994895`
- Install type: `git, manual build, executed locally within the build folder`
- Distro and version: `archlinux `

CelDaemon:
This PR fixes #832 , ensuring any applications listening to the pause state properties will always receive proper updates.

The introspection XML is also modified to advertise the changed signals for the *Length properties, which was missing before, and mentioned in the same issue.

### DIFF
--- a/src/dbus.c
+++ b/src/dbus.c
@@ -1001,6 +1001,39 @@ void signal_length_propertieschanged(void)
         g_clear_pointer(&invalidated_builder, g_variant_builder_unref);
 }
 
+void signal_paused_propertieschanged(void)
+{
+        if(!dbus_conn)
+                return;
+
+        struct dunst_status status = dunst_status_get();
+
+        GVariantBuilder *builder = g_variant_builder_new(G_VARIANT_TYPE_VARDICT);
+        GVariantBuilder *invalidated_builder = g_variant_builder_new(G_VARIANT_TYPE_STRING_ARRAY);
+
+        g_variant_builder_add(builder,
+                              "{sv}",
+                              "paused", g_variant_new_boolean(status.pause_level != 0));
+        g_variant_builder_add(builder,
+                              "{sv}",
+                              "pauseLevel", g_variant_new_uint32(status.pause_level));
+
+        g_dbus_connection_emit_signal(dbus_conn,
+                                      NULL,
+                                      FDN_PATH,
+                                      PROPERTIES_IFAC,
+                                      "PropertiesChanged",
+                                      g_variant_new("(sa{sv}as)",
+                                                    DUNST_IFAC,
+                                                    builder,
+                                                    invalidated_builder),
+                                      NULL);
+
+        g_clear_pointer(&builder, g_variant_builder_unref);
+        g_clear_pointer(&invalidated_builder, g_variant_builder_unref);
+
+}
+
 static void dbus_cb_Notify(
                 GDBusConnection *connection,
                 const gchar *sender,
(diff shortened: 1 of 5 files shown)


CelDaemon:
Checks seem to have failed, I assume there's a problem in the CI?

bynect:
Yes, its unrelated to the test suite.
I'll take a look

bynect:
code looks good.
do you mind adding a test like these?
 https://github.com/dunst-project/dunst/blob/ee63a720acad34f62567a1ae6a8e7b0636e56a50/test/dbus.c#L111

CelDaemon:
@bynect I've added two tests for the pause and unpause signals, I hope these will suffice.

One thing I have noticed is that tests which subscribe to DBus signals seem to lock up when failing. This seems to be a result of the assert functions exiting the function immediately without ever getting to the cleanup code, which in turn leaks memory and never unsubscribes from the DBus signal.

I haven't solved this issue in my new test though, as doing so likely means a rework in many of the existing tests as well.

bynect:
two are fine 👍🏻 

> One thing I have noticed is that tests which subscribe to DBus signals seem to lock up when failing. This seems to be a result of the assert functions exiting the function immediately without ever getting to the cleanup code, which in turn leaks memory and never unsubscribes from the DBus signal.

you mean they will deadlock? 

i don't know if it is worth it reworking the code for all those test functions, after all if we detect a failure nothing guarantees the state of the program is correct

CelDaemon:
> you mean they will deadlock?

Yes, which may be troublesome if the intention is to report on all tests even if some of them fail.
Just thought I'd mention it in case it wasn't known yet, even if fixing it might not be worth it.

Glad to see everything is passing and working now though!

bynect:
LGTM. thanks

PR #1511, opened 2025-10-07
Update meson build with completions

fedang:
the use of file names instead of some glob seems silly, but I couldn't figure out a smarter way


CelDaemon:
> the use of file names instead of some glob seems silly, but I couldn't figure out a smarter way

A smarter way isn't needed, meson doesn't support glob for performance reasons and recommends manually specifying files.
The same thing is also done in the rest of the project, like in `src/meson.build`:
https://github.com/dunst-project/dunst/blob/fdcda3de7c3791db0a494edb1b44a76c3c5790df/src/meson.build#L1-L19

PR #1519, opened 2025-11-03
Reference override_pause_level in RULES modify section

lverns:
The RULES section says

    The following attributes can be overridden:

which I interpret as meaning that _only_ the attributes listed in this section can be modified by a rule, but override_pause_level can also be set.

This adds a reference to the documentation of `override_pause_level` that exists in the General section.


bynect:
thanks. probably override_pause_level should have been added to rules to begin with

PR #1522, opened 2025-11-12
Make all colors a gradient

fedang:
Step 1:
Make everything a gradient


fedang:
TODO: Push this in the next release after I fix the sepcolor auto

PR #1534, opened 2026-01-20
Add an option to pause timeout while the notification is being hovered

CarlosDLMC:
Add an option to pause timeout while the notification is being hovered.
When the mouse is removed, the notification will dissapear by it's timeout.

A functionality I needed so hard I spent my free day doing this.
@bynect

### DIFF
--- a/dunstrc
+++ b/dunstrc
@@ -330,11 +330,6 @@
     mouse_middle_click = do_action, close_current
     mouse_right_click = close_all
 
-    # Pause notification timeout when mouse hovers over the notification window.
-    # When enabled, notifications won't timeout while the mouse pointer is over
-    # them. The timeout resumes when the pointer leaves the window.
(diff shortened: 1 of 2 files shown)


fedang:
Overall it looks good. However since you only implemented the wayland version (also I am not sure of the naming, maybe it should be integrated with other mouse actions), would you mind putting it in the experimental section rather than global? (so we can refine it later)

CarlosDLMC:
> Overall it looks good. However since you only implemented the wayland version (also I am not sure of the naming, maybe it should be integrated with other mouse actions), would you mind putting it in the experimental section rather than global? (so we can refine it later)

Hi! Thanks for being so fast!
I just updated the PR with the changes you requested
I did it only for wayland because it already had pointer_handle_enter() and pointer_handle_leave() callbacks registered so the required changes are minimal.

PR #1536, opened 2026-01-27
Add support for PCRE

### LINKED ISSUE #1040: Feature: Add negative comparison for regex (and other things?)
Even with posix regex implemented (#1017) you cannot do negative matching (for example "appname is **not** firefox"). To achieve negative matching in an easy way, I'd suggest to allow writing `!=` instead of `=` for regex rules. In the future this could even be used for other things like matching `msg_urgency != critical`. 

This would look as follows:

```ini
[bigger-icons]
    appname != firefox # except for you, firefox
    icons_size = 64
```

fedang:
Glib exposes PCRE with GRegex. Since we already have the library lying around we might as well use it

### DIFF
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -717,6 +717,12 @@ This setting is currently implemented only for X11.
 Pause notification timeout when hovering with the mouse pointer.
 This setting is currently implemented only for Wayland.
 
+=item B<enable_pcre_regex> (default: false)
+
+Use PCRE regular expressions for filtering rules.
+For more information see https://en.wikipedia.org/wiki/Perl_Compatible_Regular_Expressions.
+This setting overrides B<enable_posix_regex>.
+
 =back
 
 =head1 DUNSTCTL
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -815,6 +815,7 @@ For filtering rules that filter based on strings you can use regular
 expressions. It's recommended to set B<enable_posix_regex> to true. You can then
 use the POSIX Extended Regular Expression syntax:
 https://en.m.wikibooks.org/wiki/Regular_Expressions/POSIX-Extended_Regular_Expressions.
+Eventually B<enable_posix_regex> will be enabled by default.
 
 A matching rule is made by assigning a filter attribute to the value you want to match.
 Shell-like globbing is supported when matching values.
--- a/dunstrc
+++ b/dunstrc
@@ -217,6 +217,13 @@
     # Display indicators for URLs (U) and actions (A).
     show_indicators = yes
 
+    # When set to true (recommended), you can use POSIX regular expressions for filtering rules.
+    # If this is set to false (not recommended), dunst will use fnmatch(3) for matching strings.
+    # Dunst doesn't pass any flags to fnmatch, so you cannot make use of extended patterns.
+    #
+    # Note that this will eventually be true by default.
+    enable_posix_regex = false
+
     ### Icons ###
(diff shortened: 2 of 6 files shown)


fedang:
@fwsmit let me know what do you think of this

fwsmit:
I think it would be better to have a regex type config variable instead if multiple booleans that have an arbitrary priority.
What is the advantage of having two regex types in dunst? Is there not one that is the best?

fedang:
> I think it would be better to have a regex type config variable instead if multiple booleans that have an arbitrary priority.
> What is the advantage of having two regex types in dunst? Is there not one that is the best?

The best is undoubtedly PCRE as it allows many things like negative matches, lookahead ecc. In an ideal world I would remove the posix regex boolean and just use this. However the config is literally called `enable_posix_regex` so it would be misleading to enable PCRE in my opinion. The idea was to add this new experimental flag and eventually remove enable_posix_regex (with proper docs).

So mainly it's backwards compatibility. 
If we are willing to break it I can just make `enable_posix_regex` issue a warning and basically be an alias to `enable_pcre_regex`.

Another option is to add an option like `match_flavour` that could be any of fnmatch, regex, pcre, ecc. So that there are no arbitrary booleans. But anyhow the same breakage as the previous example is needed.

Let me know what do you think it's best moving forward

fedang:
@dunst-project/developers thoughts?

fwsmit:
Is pcre regex backward compatible with the current posix regex?

fedang:
as  far as I know it's not 100% backwards compatible. that's why I didn't silently upgrade enable_posix_regex to use GRegex implementation.

bebehei:
Main difference between PCRE and non-PCRE is the escaping of braces. One needs to escape them to have special meaning, the other you have to escape to match a brace literal. This sucks. It's like a USB-A connector. Correct at the ~~first~~ third try.

For the rest, PCRE is mostly an upgrade.

---

BTW: you're allowed to release dunst 2.0 and drop the unwanted code. So you could drop ERE support here and switch to PCRE.

fedang:
> Main difference between PCRE and non-PCRE is the escaping of braces. One needs to escape them to have special meaning, the other you have to escape to match a brace literal. This sucks. It's like a USB-A connector. Correct at the ~~first~~ third try.
> 
> For the rest, PCRE is mostly an upgrade.
> 
> ---
> 
> BTW: you're allowed to release dunst 2.0 and drop the unwanted code. So you could drop ERE support here and switch to PCRE.

Yes the idea was to provide "experimental" PCRE support and in v2 just phase out ERE.
Also I was thinking of dropping fnmatch alongside ERE, let me know if that would be too drastic. (Maybe I am a bit biased but regex seems just better than fnmatch)

bebehei:
I thought `fnmatch` was the low hanging fruit available and enough. I expected PRCE/Regex would increase the footprint and therefore refrained from adding it. I wasn't aware it was basically free in GLib already.

Regex is for devs definitely more flexible, but more complex for entry level linux users.

You do the decision. I can go with both.

fedang:
@dunst-project/developers I was thinking of doing something like this:

1) initially, add the pcre setting in "experimental" section
2) if all goes well, move it to "global" and issue a deprecation message for `enable_posix_regex`
3) eventually remove posix regex in favour of pcre regex
4) when we reach v2 we can think if we want to keep fnmatch or not (I will postpone this decision for then)

if this sounds reasonable I will merge.
let me know your opinions

fwsmit:
Sounds fine to me

PR #1538, opened 2026-01-28
Improve doxygen docs

fedang:
I'm trying to make the documentation a bit better. I was thinking of polishing doxygen support and maybe loading it in the website for easier access

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,12 @@
+<!-- Please put the pull request description above -->
+
+---
+
+Please check all the boxes that apply:
+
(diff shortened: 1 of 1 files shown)


fedang:
@fwsmit I'm pinging you just to see what do you think of this.
 I overhauled the doxygen documentation hoping that it can help new contributors navigate the codebase.
Also I added an SPDX license line to each source file and uniformed (somewhat) the comment style.
See https://github.com/dunst-project/dunst-project.org/pull/26 for support in the website.

Let me know if I missed something.

PS: Also I assumed that files with no license info are also BSD 3-clause. For example your code in wayland/. Let me know if some files should use another license

fedang:
A live preview of the doxygen docs are live at https://dunst-project.org/doxygen/index.html.
They can be taken down at any point by building the website against the master branch rather than this one

Issue #1333, opened 2024-04-10
Proposal: More consistent filter/setter rule names


bynect:
I propose a more consistent syntax for filters and rules following the good example of `match_transient/set_transient`.
We can add for every notification property a `match_*` and a `set_*` value for filtering and modifying respectively.
This is much more consistent than the currently somewhat confusing names (`msg_urgency/urgency`, etc).

Obviously for backwards compatibility the current name will still work (as an alias for the appropriate new name).

If I get to implement this I will also probably change the rules internal representation (see #607).

If you have any ideas about this let me know 👍🏻

bynect:
For reference

`msg_urgency` & `msg_urgency` -> `set_urgency` & `match_urgency`
`override_dbus_timeout` -> `set_dbus_timeout`
`new_icon` & `icon` -> `set_icon` & `match_icon`
...etc

Ideally this new names can be parsed automatically as "set_"/"match_" + prop_name

fwsmit:
Yeah, sounds good. I think it would also be good to write some documentation about this. Rules are added one by one and it might go wrong again in the future

fwsmit:
Making it work programmatically would also be a solution to force it to be consistent

bynect:
> Making it work programmatically would also be a solution to force it to be consistent

That was the idea, like creating an enum with the rule property (urgency, color, ...) and one with actions (set, match, ...maybe we'll add more...) and programmatically parse them.

This seems like a nice qol goal for v2

bynect:
I found that an old issue proposed the same thing #349 🤦🏻

bynect:
I am undecided about using `match_property` or `has_property` like it was proposed in #349. Mostly for brevity and symmetry (3 letters like set).

@fwsmit @zappolowski what do you think about this ^^^^

If I could I would make a poll, but I don't think it is possible.

fwsmit:
I'm not sure either. Making a poll is a good idea

bynect:
> I'm not sure either. Making a poll is a good idea

I made it in discussions. I don't know how much reach it will have unfortunately.

Issue #607, opened 2019-02-23
Refactor rules


tsipinakis:
Currently we're storing the rules in a [huge struct](https://github.com/dunst-project/dunst/blob/977a7a7ea47f2d86ea3168e609576ebeb0ffa719/src/rules.h#L11-L39). This is obviously neither space nor time efficient as we store a full value for all unset rules and on match we have to compare all possible match and set rules.

My proposed implementation would be something like this: (pseudocode, don't expect it to compile! :) )
```c
enum rules_match {
    MATCH_APPNAME,
    ...
};
enum rules_set {
    SET_FORMAT,
    ...
};

struct rule_part {
    union {
        enum rules_match match;
        enum rules_set set;
    };
    void *data;
};

struct rule {
    struct rule_part *match_rules;
    struct rule_part *set_rules;
};
```

So as it's apparent this involves having each rule knop in an enum, and inside the rule struct we have 2 arrays one for the match rules and one for the set rule.

In order to apply the rule we simply iterate over the match array and check if each rule_part matches correctly. If it does we start iteration on the set rules and applying each one appropriately.

The 2 big downsides are:
* We have to go through each rule twice, one to count the number of match and set rules in order to create the array and one got actually creating it.
    This may be remedied by replacing the array with a linked list but that may be overcomplicating it.

* Given Cs quite stingy approach with types we (unfortunately) have to abuse the void pointer again which leads to a big number of tiny allocations which might not be the best thing
    An alternative here might be to replace the void pointer with a union of every possible data type.

bebehei:
The approach seems tempting. Running two times over the list might be a downside, but I'd gladly accept it if it allows abstraction of our rule.

Especially this lump of code would vanish

https://github.com/dunst-project/dunst/blob/7b3d85a6d0957fba03ed89a92b588e2349d10c75/src/rules.c#L93-L104

and would get replaced with a simple for loop:

```c
bool rule_matches_notification(struct rule *r, struct notification *n)
{
        for (struct rulepart **r->match_rules; *rp; rp++)
                if (!rule_part_matches(n, *rp)
                        return false;
        return true;
}
```

Sexy.

---

But, how do you solve `rule_part_matches`?

That has to be a super huge switch case fuckup:


```c
bool rule_part_matches(struct rule_part rp, notification *n)
{
        switch (rp->match) {
        case MATCH_APPNAME:
                return fnmatch(n->appname, rp->data...);
                break;
        case MATCH_...:
[...]

PR #1539, opened 2026-02-01
Add fullscreen suppress mode

### LINKED ISSUE #1187: Feature suggestion: skip notification when in fullscreen mode
Currently there are 3 options for fullscreen: show, delay and pushback.
It would be nice to have 1 more: skip.

If the notification is already shown => undraw it when going into fullscreen.
If the notification comes when in fullscreen => don't draw it.

Thanks in advance :)

fedang:
Closes #1187. Don't show/draw notification received when fullscreen, just directly put them in history

---

Please check all the boxes that apply:

- [ ] I have read the contribution guidelines in [CONTRIBUTING](https://github.com/dunst-project/dunst/blob/master/CONTRIBUTING.md) guide.
- [ ] I have added documentation for new features (if applicable).
- [ ] I have added tests for new features (if applicable).
- [ ] I have used the assistance of AI/LLM tools (not forbidden, but please disclose).
- [x] The new code successfully builds locally (`make all` shows no error).
- [x] The new code successfully passes the test suite (`make test` shows no error).

### DIFF
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -918,20 +918,20 @@ Equivalent to the C<format> setting.
 
 The frame color color of the notification. See COLORS for possible values.
 
-=item C<fullscreen> (values: [delay/show/pushback/drop], default: show)
+=item C<fullscreen> (values: [delay/show/pushback/suppress], default: show)
 
 This attribute specifies how notifications are handled if a fullscreen window
-is focused. By default it's set to show so notifications are being shown.
+is focused. By default it's set to I<show> so notifications are being shown.
 
-Other possible values are delay: Already shown notifications are continued to be
+Other possible values are I<delay>: already shown notifications are continued to be
 displayed until they are dismissed or time out but new notifications will be
 held back and displayed when the focus to the fullscreen window is lost.
 
-Or pushback which is equivalent to delay with the difference that already
+Or I<pushback> which is equivalent to delay with the difference that already
 existing notifications are paused and hidden until the focus to the fullscreen
 window is lost.
 
-There is additionally the drop value. Existing notifications will be closed
+There is additionally the I<suppress> value. Existing notifications will be closed
 when entering fullscreen and new notifications will go straight to history.
 This is similar to pushback without redisplaying after leaving fullscreen mode.
 
--- a/docs/dunst.5.pod
+++ b/docs/dunst.5.pod
@@ -918,7 +918,7 @@ Equivalent to the C<format> setting.
 
 The frame color color of the notification. See COLORS for possible values.
 
-=item C<fullscreen> (values: [delay/show/pushback], default: show)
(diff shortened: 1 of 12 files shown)


fedang:
Ready for me. Not sure if I missed some edge cases, but at the very least this should not break the other modes

bebehei:
The values `skip`, `drop` are IMHO quite ambiguous. I have thought about `mute`, but this is also ambiguous.

I'd name the value `history` or `skip_display` to match the rule engine.

--

I'm not sure how much AI you use, but I'd suggest making functional enhancements and unrelated formatting into separate commits. Makes it much easier for Code Reviewers.

fedang:
0% AI, 100% I can't stand that switch/case style :)
Sorry for letting the refactoring seep in

fedang:
> The values `skip`, `drop` are IMHO quite ambiguous. I have thought about `mute`, but this is also ambiguous.
> 
> I'd name the value `history` or `skip_display` to match the rule engine.


`history` seems very vague to me. `skip_display` is clear but much longer/different than the other (show/pushback/delay).
Would something like `hide` be more appropriate than `skip`/`drop` ?
Maybe `ignore` also applies?

bebehei:
I've suggested `skip_display`, because this is the value in the rules, which does the same.

fedang:
> I've suggested `skip_display`, because this is the value in the rules, which does the same.

There is a future plan to "modernize" the rules mechanism by providing a `match_property` and `set_property` value for each property instead of the fragmented thing we have now.
(See https://github.com/dunst-project/dunst/discussions/1341 https://github.com/dunst-project/dunst/issues/1333 https://github.com/dunst-project/dunst/issues/607)

This is not strictly related, but i would rather choose a short and recognizable name for new features rather than how rules are named right now.

With that said, here's a list of meaningful verbs that I can think of:
- `skip`: The one proposed by the issue, might be ambiguous (skip show? history?)
- `drop`: Currently used by the PR, the intention was to highlight the fact that the notification was removed from the queue
-  `ignore`: Clarifies the intent, the user doesn't care about the notification (but the semantics are less clear)
- `conceal`/`hide`: Makes clear that the notification is not shown, but again is not very specific
- `suppress`: This more closely aligns with how we described other way to not display notifications

Right now I am tending towards `suppress` as it is semantically clear and could be reused in the future.
Let me know if you prefer another name

bebehei:
Yep, suppress, seems to be the best.

fxzzi:
is there a possibility to extend this feature, so that if the monitor that the notification would normally be displayed on is in fullscreen, then display on another monitor instead?

fedang:
> is there a possibility to extend this feature, so that if the monitor that the notification would normally be displayed on is in fullscreen, then display on another monitor instead?

It would be possible to add such mechanism, albeit that would be more complex than this.
But how would you choose the other monitor when many are present?

PR #1550, opened 2026-05-11
[Feature] Timeout bar

Harshit-Dhanwalkar:
## [Feature] Timeout Bar

| Low urgency  | High urgency  |
| --- | --- |
|low-urgency| normal-urgency|

---
## Update: Independent of content updates
Changes:
- Added a dedicated bar_start field in struct notification. This is set only once when notification first appears, so the bar doesn't reset when text (notification context) changes.
- The `start`, `timeout`, and `bar_start` fields are now preserved when a notification is updated or stacked.
- The timeout bar now continues shrinking smoothly even when notification text updates (e.g., during a countdown).

| Before  | After  |
|---|---|
|before|after|
|Timeout_bar was reseting everytime notification context updated| Now timeout_bar is independent of updating context|

---
## Update: Made timeout_bar render behind content and more customisable by adding enable option, style options, bar_height

- The bar is now drawn behind the notification text and icons, so it never block the content.
- Timeout bar now can be enabled or disabled from `dunstrc` by setting `enable_timeout_bar = true # or false`

| `top` | `bottom` |
|---|---|
|timeout_bar_style_top|timeout_bar_bottom|
| `top_bottom` | `glow` |
|timeout_bar_style_top_bottom|timeout_bar_glow| 
| `up_span` | `down_span` |
|timeout_bar_style_up|timeout_bar_style_down|
| `floodin` | `drainout` |
|timeout_bar_floodin|timeout_bar_drainout|
| `left_span` |
|timeout_bar_style_full|

---

Please check all the boxes that apply:

- [x] I have read the contribution guidelines in [CONTRIBUTING](https://github.com/dunst-project/dunst/blob/master/CONTRIBUTING.md) guide.
- [ ] I have added documentation for new features (if applicable).
- [ ] I have added tests for new features (if applicable).
- [ ] I have used the assistance of AI/LLM tools (not forbidden, but please disclose).
- [x] The new code successfully builds locally (`make all` shows no error).
- [x] The new code successfully passes the test suite (`make test` shows no error).


fedang:
hello. there is another feature called progress bar, so maybe this time visualization should be called differently?
also, please do not format automatically the code, it makes impossible for me to review the actual changes

Harshit-Dhanwalkar:
> please do not format automatically the code, it makes impossible for me to review the actual changes

I have updated everything to `.clang-format`, i.e,
```yaml
BasedOnStyle: Google
BreakBeforeBraces: Allman
ColumnLimit: 0
IndentWidth: 8
```

I have seen different files uses different formats, which i have updated as well. I highly recommend everyone to use same format style.

Harshit-Dhanwalkar:
> hello. there is another feature called progress bar, so maybe this time visualization should be called differently?

Regarding name 'progress bar' I will be working afterwards.

EDIT: I have renamed it to `timeout_bar`

Harshit-Dhanwalkar (commented):
I have updated the code and  edited description. Please check it.

fedang:
thanks for the description, the gifs are nice. 
I can't review the code today, but I will leave you some notes.

First, make sure that the test suite passes (make test and make test-valgrind). The ci shows a failure but it seems a warning.
Then, have you considered what happens when a notification is popped from history to be rewatched.
What about when dunst is reloaded with dunstctl reload?


Could you address these various small things? If something else comes to mind I will write it.

Harshit-Dhanwalkar:
> thanks for the description, the gifs are nice.

You are welcome.

> make test

`Expected: 100001   Got: 100000  	FAIL test_timeout: expected_timeout != n->timeout (test/dbus.c:1610) (9481 ticks, 0.009 sec)`
`Expected: testing RuleList  Got: test_timeout  FAIL test_dbus_cb_dunst_RuleList: "testing RuleList" != str (test/dbus.c:844) (15921 ticks, 0.016 sec)`

This is because timeout test check for micro second precision (in `test/dbus.c` in `test_timeout(void)`)
```c
        rule = rule_new("test_timeout");
        rule->appname = "dunstteststack";
        rule->timeout = 100001;
```
, but timeout leaks somewhere. I will check into that.

D‑Bus updates in milliseconds (not in microseconds), and actually microsecond precision is not needed here. So if this rule for timeout to check for micro second precession can be changed to millisecond (`rule->timeout = 100000;`) then this test will be pass. 

Edit: After changing `100001` to `100000`, all tests are passed:
`Total: 211 tests (66778 ticks, 0.067 sec), 2823 assertions
Pass: 211, fail: 0, skip: 0.`

> make test-valgrind.

before it was showing this:
`==36183== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 6 from 6)`
`==36188== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)`

Edtr: After changing `100001` to `100000`, all tests are passed.

Harshit-Dhanwalkar (commented):
I have reviewed the change. `test_timeout` was failing because `rule->timeout = 100001` is not applied; the notification's timeout becomes `100000`  (from `dbus_timeout`). The `100001` checks for micro second precison which I think is not required, just `100000` (milil seconds) is fine for this.

Harshit-Dhanwalkar:
Build error is showing failing because in `test/setting.c`, the variable `message` array is declared locally, just by making it `static` it's most probably be fixed.

Harshit-Dhanwalkar:
> what happens when a notification is popped from history to be rewatched?

`n->timeout = settings.sticky_history ? 0 : n->timeout;`
Because sticky_history defaults to true, the timeout becomes 0 (sticky). That means the notification never times out and the timeout bar is not drawn because the condition `n->timeout > 0` fails. If anyone wants timeout_bar in history-pop then one can set `sticky_history = false`. Also it was updating content (like timer) history pop will show each frames (since updating dunst requires assigning each frame it's own ID).

Maybe for this I can implement something like `groupID` so a updating content have it's own ID as a whole and not only to it's each frames.

> What about when dunst is reloaded with dunstctl reload?

Timeout_bar become mismatched because the original `bar_start` is still the same but the total duration is different. (Bug?)

Makefile
Harshit-Dhanwalkar, line 1: `make` to spawn multiple jobs for faster compilation.

config.mk
Harshit-Dhanwalkar, line 1: Add extra warning in config.mk.

Harshit-Dhanwalkar (commented):
Add timeout_bar styling options and argument to disable timeout bar from `dunstrc`.

Makefile
fedang: please avoid doing this. jobs should be specified explicitly by the end user/packager as parameters

config.mk
fedang: why add all of those? I think most are already covered by -Wall.
Also, you left some of the warnings in the CFLAGS inconsistently.
I think this change should be done separately from the new feature

src/settings_data.h
```diff
@@ -1438,6 +1448,47 @@ static const struct setting allowed_settings[] = {
                 .parser = NULL,
                 .parser_data = NULL,
         },
+        {
```
fedang, line 1454: this section has improper indentation

src/dbus.c
```diff
@@ -494,6 +495,67 @@ static void gradient_entry(const struct gradient *grad, GVariantDict *dict, cons
         }
 }
 
+static void dbus_cb_dunst_ConfigReload(GDBusConnection *connection,
```
fedang: these changes seem to just reshuffle code around?

Harshit-Dhanwalkar: Yeah, I saw the comment note `// NOTE: Keep the names sorted alphabetically`

fedang: that refers specifically to the list since we use binary search.
for the code, you can leave it like it was so the diff is cleaner

src/draw.c
```diff
@@ -908,6 +908,19 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, int
                 cairo_fill(c);
 
                 // top layer (fill)
+                // if (COLOR_VALID(cl->n->colors.timeout_bar))
+                // {
```
fedang: why is this comment here?

src/draw.c
```diff
@@ -917,6 +930,10 @@ static void render_content(cairo_t *c, struct colored_layout *cl, int width, int
                         settings.progress_bar_corner_radius, scale, settings.progress_bar_corners);
                 cairo_fill(c);
 
+                draw_rounded_rect(c, x_bar_1, frame_y, progress_width_1, progress_height,
```
fedang, line 915: why is this duplicated?

src/settings_data.h
```diff
@@ -375,6 +376,15 @@ static const struct string_to_enum_def corners_enum_data[] = {
         ENUM_END,
 };
 
+static const struct string_to_enum_def timeout_bar_style_enum_data[] = {
```
fedang, line 378: what is top_bottom? it seems unclear what those values do (please add to the docs)
maybe reusing another enum like the corners can avoid adding a new enum (with a sanitation step in settings.c)

src/draw.c
Harshit-Dhanwalkar: It was leftover from an earlier attempt to color the progress bar (notification progress, e.g., download progress) using the timeout bar colour. But the timeout bar is drawn in layout_render. I have remove it.

src/draw.c
Harshit-Dhanwalkar: This is resolved. removed duplicate lines.

src/settings_data.h
Harshit-Dhanwalkar: > avoid adding a new enum

Current styles include span behaviors that don’t map directly to a simple corner flags (`C_TOP`, `C_BOT`, etc. ). So I think keeping existing small list of named styles is good practice here, also it keeps different features separate (corner and timeout_bar), which would be good for documenting.

src/settings_data.h
Harshit-Dhanwalkar: Indentations are fixed now.

fedang:
is this ready for review?

Harshit-Dhanwalkar:
> is this ready for review?

Yes

PR #1562, opened 2026-08-23
fix: docs: clarify pause_level vs urgency relationship

### LINKED ISSUE #1554: Unclear pause level and urgency relation in the docs
The [docs](https://dunst-project.org/documentation/) on pause level say

> Specify a pause level to be set on dunst startup. Notifications are only shown if their urgency level is higher than the pause level. See override_pause_level for more information.

While the `pause_level` takes a value from 0 to 100, the `urgency` takes low, normal or critical. It's not clear how these relate. Also, how do `urgency` and `override_pause_level` play together?

VedantMadane:
## Summary

docs: clarify pause_level vs urgency relationship

## Changes

- CHANGELOG.md

Fixes #1554

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@
 ### Changed
 - Improve Makefile and build flags (#1456, #1457)
 - Change the way icons are cached (#1473, #1471)
+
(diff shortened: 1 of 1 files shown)


fedang:
why did you put it in the changelog instead of the doc files? mistake?

PR #1563, opened 2026-09-19
wayland: drop the surface when its output goes away

dmitry-j-mikhin:
Fixes the case behind #1396: notifications silently stop being shown after an
output is destroyed and recreated (monitor unplugged, DPMS cycle that drops
HPD, suspend/resume).

## The bug

`destroy_output()` clears `ctx.surface_output` and `ctx.layer_surface_output`
but leaves `ctx.surface`, the shell surface, `ctx.dirty` and
`ctx.frame_callback` alone. sway did not send
`zwlr_layer_surface_v1.closed` in any of the output removals I looked at, so
the cleanup in `layer_surface_handle_closed()` never runs — its
`LOG_I("Destroying layer")` did not fire once across six destroy/recreate
events. A frame callback that was in flight is therefore orphaned: its surface
is not shown on any output any more, so `wl_callback.done` never arrives and
`frame_handle_done()` never runs.

That leaves `ctx.dirty == true` together with a non-NULL `ctx.frame_callback`,
which is terminal:

```c
void set_dirty(void) {
        if (ctx.dirty) return;            // stuck true -> no-op forever
        ...
}
static void schedule_frame_and_commit(void) {
        if (ctx.frame_callback) return;   // stuck non-NULL -> no-op forever
        ...
}
```

`send_frame()` is never called again. dunst keeps accepting notifications and
reports them as displayed, redraws them into a fresh shm buffer on every tick
from `wl_display_surface()`, and throws every buffer away because it never
gets as far as attaching one. Only `dunstctl reload` or a restart recovers.

State read out of a stuck 1.12.2 process with gdb:

```
ctx.dirty                = true
ctx.frame_callback       = 0x558e8ae33ed0
ctx.surface_output       = NULL
ctx.layer_surface_output = NULL
ctx.configured           = true
ctx.width, ctx.height    = 309, 38
ctx.cur_dim              = 309 x 190
```

The output list itself was intact, and a breakpoint on `send_frame()` did not
fire once in two minutes.

## The change

`wl_teardown_surface()` destroys whichever shell surface exists, clears the
two output pointers and defers the rest to the existing
`surface_handle_closed()`. `destroy_output()` calls it when the surface lives
on the output being removed.

Two details worth a look during review:

- the call is made **after** `wl_list_remove()`, so the surface is rebuilt on
  one of the remaining outputs instead of the one being destroyed;
- it is guarded on the surface pointers themselves, because `wl_deinit()`
  destroys them before walking the output list and must not resurrect a
  surface during shutdown.

I deliberately left `layer_surface_handle_closed()` and
`xdg_toplevel_handle_close()` alone to keep the diff small, though both could
be refactored onto `wl_teardown_surface()` if you would prefer that.

## Testing

`make test` on this branch: 211 passed, 0 failed. Builds clean under
`-Wall -pedantic`.

Live, on this branch under sway 1.10, with `monitor = HDMI-A-1`,
`follow = none`, `timeout = 0`, driving the output away with
`swaymsg output HDMI-A-1 disable` / `enable`:

```
baseline      dirty=0 fcb=(nil) surf_out=0x392a4f70 lso=0x392a4f70 ctx=309x54 cur_dim=309x54
after disable dirty=0 fcb=(nil) surf_out=0x392a2750 lso=0x392a2750 ctx=309x54 cur_dim=309x54
              destroy_output: Output with our surface is gone, dropping the surface
after enable  dirty=0 fcb=(nil) surf_out=0x393369f0 lso=0x393369f0 ctx=309x54 cur_dim=309x54
```

The surface follows the output across both transitions and notifications keep
working, where before the first transition would wedge it permanently.

The same change against 1.12.2 has also been running here through five
monitor power cycles and one physical cable unplug, in a single process with
no restarts, recovering every time.


dmitry-j-mikhin:
The two red checks here (`build-linux (clang, archlinux)` and `build-linux (gcc, archlinux)`) are a pre-existing CI infrastructure problem, not something introduced by this PR.

On the Arch image, `make test-valgrind` dies before it runs a single test:

```
valgrind:  Fatal error at startup: a function redirection
valgrind:  which is mandatory for this platform-tool combination
valgrind:  cannot be set up.  Details of the redirection are:
valgrind:
valgrind:  A must-be-redirected function
valgrind:  whose name matches the pattern:      memcmp
valgrind:  in an object with soname matching:   ld-linux-x86-64.so.2
valgrind:  was not found whilst processing
valgrind:  symbols from the object with soname: ld-linux-x86-64.so.2
```

Arch ships a stripped `ld.so` without debuginfo, so Valgrind 3.25.1 cannot install its mandatory `memcmp` redirection and bails out at startup.

Everything before that step is green on Arch as well:

- the build succeeds with `-Werror` under both clang and gcc;
- the plain test suite passes: `Total: 211 tests, 2775 assertions. Pass: 211, fail: 0, skip: 0`;
- `test-install.sh` passes.

Only the subsequent `make test-valgrind` invocation fails.

The same two jobs fail identically on `master`, e.g. in [run 31514717839](https://github.com/dunst-project/dunst/actions/runs/31514717839) (push to `master` on 2026-08-11, a README-only commit) — same Valgrind startup error, unrelated to any code change. So the Arch target has been broken for a while.

Fixing it would mean installing `glibc-debuginfo` in the `ghcr.io/dunst-project/docker-images/archlinux` image, or skipping the Valgrind step on Arch until the image is updated. Happy to open a separate issue or PR against the CI setup if that's useful — just let me know.

dmitry-j-mikhin:
Follow-up on the two red `archlinux` checks: I traced the whole chain, and the
fix is now up as two PRs.

**Why the Arch jobs fail.** The CI images have not been rebuilt since
**2026-05-15** — every scheduled refresh since has failed. The image therefore
still carries glibc `2.43+r22+g8362e8ce10b2-2` while Arch has moved on to
`2.44+r24+g16be1518495f-1`, and `debuginfod.archlinux.org` only serves debug
info for packages *currently* in the repos, so the image's own build-id is gone:

```
url 0 https://debuginfod.archlinux.org/buildid/8f271b1884fa6225b5ac96a6bcb7326a5f1f25ec/debuginfo
server response HTTP response code said error
url 0 The requested URL returned error: 404
```

Without those symbols valgrind cannot set up its mandatory `memcmp` redirection
against Arch's stripped `ld.so`, and aborts before running a single test.

**Why the images went stale.** `debian-bullseye` broke in June (its
`bullseye-security` pool is gone, every `.deb` 404s), and the image matrix runs
with the default `fail-fast`, so that one failure **cancels the other ten
jobs**. In the last scheduled run
([#34949890661](https://github.com/dunst-project/docker-images/actions/runs/34949890661))
ten of eleven jobs are `cancelled`. `alpine` and `alpine-edge` are broken too —
they build lcov from an unpinned clone of its master branch, which no longer
compiles without `g++` and `sphinx-build`.

**The PRs:**

* **dunst-project/docker-images#18** — `fail-fast: false` so one bad
  distribution cannot block the rest; `bullseye` switched to the
  `snapshot.debian.org` sources its base image already ships; `alpine`/
  `alpine-edge` switched to the packaged lcov; and `archlinux` now installs
  `glibc-debug` from `[core-debug]`, which pins the debug info to the exact
  glibc in the image instead of hoping debuginfod still has it. All eleven
  images build, and the full dunst CI sequence passes inside the rebuilt
  `archlinux` (clang and gcc), `debian-bullseye` and `alpine`.

* **#1565** — one extra valgrind suppression. Refreshing the images is not
  quite free: a rebuilt `alpine-edge` picks up a newer fontconfig whose leak is
  reached through `libglycin-2` rather than pango, so the existing
  `fontconfig_pango_leaks` entry does not match it. That PR adds the sibling
  entry; verified before/after on clean checkouts in seven freshly built images.

Merging #18 alone turns the two Arch jobs here green; #1565 should land with it
so `alpine-edge` does not go red in its place. Nothing in either PR touches this
branch — it needs no changes.

(Note: the CI run on #1565 is sitting in `action_required` and needs a
maintainer to approve it.)

